PR #1, opened 2017-03-24
Fix composer validate position

tractorcow:
Move to before running composer install, since it's too late after that fact.


tractorcow:
@dhensby

PR #2, opened 2017-03-28
ENHANCEMENT Add campaign description field

phalkunz:


src/ChangeSet.php
```diff
@@ -50,7 +51,8 @@ class ChangeSet extends DataObject
     private static $db = [
         'Name'  => 'Varchar',
         'State' => "Enum('open,published,reverted','open')",
-        'IsInferred' => 'Boolean(0)' // True if created automatically
+        'IsInferred' => 'Boolean(0)', // True if created automatically
+        'MetaDescription' => 'Text', // True if created automatically
```
sminnee: This is just "description", not "meta-description".

It's not designed to populate a meta-tag, it's a human description of what the campaign is for to help content authors do their job.

For example, we might have a campaign with the following details:

 * Title: Aug 2017 Promo
 * Description: E-commerce catalogue Updates of new Nike gear to be launched in our Aug 2017 promotions. Needs to match our print brochures going out the same week.

It should also be considered if/how this should be listed on the campaign list view, and on the campaign detail view. Talk to Paul.

sminnee:
@phalkunz moved back to "in progress" and gave some direction on requirements.

src/ChangeSet.php
phalkunz: @sminnee Thanks for the clarification.

phalkunz: @sminnee I talk to Paul about having this field on the list view and he said no because he wanted to have space for other details like last publish date, due date etc.

tractorcow: Could we have a tooltip and / or toggle button that we could click to get a "hover" description?

src/ChangeSet.php
```diff
@@ -437,11 +439,13 @@ public function getCMSFields()
...
         if ($this->IsInferred) {
             $fields->addFieldToTab('Root.Main', ReadonlyField::create('Name', $this->fieldLabel('Name')));
+            $fields->addFieldToTab('Root.Main', ReadonlyField::create('Description', $this->fieldLabel('Description')));
         } else {
             $fields->addFieldToTab('Root.Main', TextField::create('Name', $this->fieldLabel('Name')));
+            $fields->addFieldToTab('Root.Main', TextareaField::create('Description', $this->fieldLabel('Description')));
         }
         if ($this->isInDB()) {
-            $fields->addFieldToTab('Root.Main', ReadonlyField::create('State', $this->fieldLabel('State')));
+            $fields->addFieldToTab('Root.Main', ReadonlyField::create('State', $this->fieldLabel('State')), 'Description');
```
phalkunz: Just making sure that `State` field comes before `Description` field.

flamerohr: oh you're right, I guess I overlooked a bracket, lol

PR #3, opened 2017-03-28
Fix min-stability and README

chillu:


tractorcow:
Improves tests, even though it doesn't fix them. Merging.

PR #4, opened 2017-03-28
BUG Add admin module to build configuration

tractorcow:


sminnee:
I don't quite see why this is necessary, or why the framework tests are part of the versioned test run.

tractorcow:
👍

chillu:
Four +1s on Sam's alternative suggestion (incl. me), so closing this

tractorcow:
Admin module is still needed; We should have just removed the tests, not the actual .travis.yml fix.

PR #7, opened 2017-03-30
FIX augmentDatabase should be called via extend(), not dirrectly

mikenz:
augmentDatabase is added via a DataExtension. If there's more than one extension defining it, this test was only running one of them.


dhensby:
Hmm - extend isn't typically an external API to be used in this way... though, this is a test (and a rather gross one at that)...

sminnee:
Build failure unrelated to this PR. If it passes I'm okay with this being merged, as it's replacing one internal call with another internal call.

sminnee:
@mikenz could you rebase this so as to hopefully fix the tests?

mikenz:
Rebased

PR #8, opened 2017-04-03
FIX: Remove unnecessary test runs of other modules.

sminnee:


sminnee:
...let's see what travis makes of this.

sminnee:
OK I believe that the codecov failure is because this'll be the first successful coverage run.

sminnee:
@silverstripe/core-team this was approved, can we merge? :)

PR #9, opened 2017-04-04
ENHANCEMENT Remove item from change sets when it's archived

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-framework/issues/6258

### DIFF
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -1612,11 +1612,29 @@ class Versioned extends DataExtension implements TemplateGlobalProvider, Resetta
         }
 
         $owner->invokeWithExtensions('onBeforeArchive', $this);
+        $owner->deleteFromChangeSets();
         $owner->doUnpublish();
         $owner->deleteFromStage(static::DRAFT);
         $owner->invokeWithExtensions('onAfterArchive', $this);
 
         return true;
+    }
+
+    public function deleteFromChangeSets()
+    {
+        $owner = $this->owner;
+        if (!$owner->canArchive()) {
+            return false;
+        }
+
+        $ids = [$owner->ID];
+        if ($owner->hasMethod('getDescendantIDList')) {
+            $ids = array_merge($ids, $owner->getDescendantIDList());
+        }
+
+        ChangeSetItem::get()
+            ->filter(['ObjectID' => $ids])
+            ->removeAll();
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/Versioned.php
```diff
@@ -1612,13 +1612,31 @@ public function doArchive()
...
         $owner->invokeWithExtensions('onBeforeArchive', $this);
+        $owner->removeFromChangeSets();
         $owner->doUnpublish();
         $owner->deleteFromStage(static::DRAFT);
         $owner->invokeWithExtensions('onAfterArchive', $this);
 
         return true;
     }
 
+    public function removeFromChangeSets()
```
flamerohr: maybe rename to `deleteFromChangeSets()` to be more aligned with another function `deleteFromStage()`

probably just me being a little bit pedantic :D

src/Versioned.php
```diff
@@ -1612,13 +1612,31 @@ public function doArchive()
...
+    {
+        $owner = $this->owner;
+        if (!$owner->canArchive()) {
+            return false;
+        }
+
+        $changeSetItems = ChangeSetItem::get()->filter(['ObjectID' => $owner->ID]);
```
flamerohr: I wonder if we can have a `getDescendantIDs()` recursive function,
 to get a list of all "Children" IDs to do the filter->delete with instead...(similar to `AssetAdminFile::nestedFolderIDs()`)
This would help reduce the number of queries required a fair amount.

src/Versioned.php
phalkunz: Nope, not pedantic at all 👍

src/Versioned.php
phalkunz: Good point! I need to remind myself to think about the cost of db queries :)

src/Versioned.php
```diff
@@ -1612,13 +1612,31 @@ public function doArchive()
...
 
+    public function deleteFromChangeSets()
+    {
+        $owner = $this->owner;
+        if (!$owner->canArchive()) {
+            return false;
+        }
+
+        $ids = [$owner->ID];
+        if ($owner->hasMethod('getDescendantIDList')) {
```
tractorcow: descendants isn't relevant in terms of versioned; This should be versioned ownership only.

src/Versioned.php
```diff
@@ -1612,13 +1612,31 @@ public function doArchive()
...
+            $ids = array_merge($ids, $owner->getDescendantIDList());
+        }
+
+        ChangeSetItem::get()
+            ->filter(['ObjectID' => $ids])
```
tractorcow: Ignores object class; This will match all objects of different classes that happen to overlap by ID.

tractorcow:
Somehow deleteFromChangesets() got deleted from 1.1 branch.... doh.

PR #14, opened 2017-05-22
API Remove object class references

tractorcow:
See https://github.com/silverstripe/silverstripe-framework/issues/6085


flamerohr:
Merging this first, framework tests cannot pass without these, and these cannot pass without framework...

PR #17, opened 2017-06-25
API Update for HTTPMiddleware

tractorcow:
Merge with https://github.com/silverstripe/silverstripe-framework/pull/7052


_config/versionedrequestprocessors.yml
```diff
@@ -2,11 +2,11 @@
 Name: versionedrequestprocessors
 After:
   - 'requestprocessors'
+  - 'coresecurity'
 ---
 SilverStripe\Core\Injector\Injector:
-  VersionedRequestFilter:
-    class: SilverStripe\Versioned\VersionedRequestFilter
-  SilverStripe\Control\RequestProcessor:
+  SilverStripe\Control\Middleware\HTTPMiddleware.director:
```
sminnee: This is now out of date syntax.

src/VersionedHTTPMiddleware.php
```diff
@@ -0,0 +1,66 @@
...
+use SilverStripe\Security\Security;
+
+/**
+ * Initialises the versioned stage when a request is made.
+ */
+class VersionedHTTPMiddleware implements HTTPMiddleware
+{
+    public function process(HTTPRequest $request, callable $next)
+    {
+        // Ensure Controller::curr() is available
```
sminnee: Do we still need Controller::curr() to be available thanks to other refactors?

tractorcow: For the time being, yes, but at least it becomes less necessary.

I would remove this block at the same time as we remove controller::curr() altogether. :)

_config/versionedrequestprocessors.yml
tractorcow: Doh, I thought I updated this.

tractorcow: Ok, pushed up fix. :)

src/VersionedHTTPMiddleware.php
tractorcow: I think it's still necessary for security class.

PR #32, opened 2017-08-07
API Don’t unpublish owners when unpublishing children

### LINKED ISSUE #25: Don't unpublish owner when unpublishing owned object
## Overview

At the moment, `Versioned->onAfterUnpublish()` will trigger a cascading unpublish on direct and indirect owners. That's consistent behaviour from a technical perspective, since the owner record now contains unpublished records and can't display the same way any more. In practice, that's not really reflecting the authors intentions. Most templates can deal with missing records:

 * If you list a has_many relationship, it'll simply omit the now unpublished owned object
 * If you show a has_one relationship, it'll skip this part of the template
 * If you have a required has_one relationship (template fails without it), you should code your template defensively (don't show that portion of the template). If it's absolutely required for the template to function, then you should opt in to cascading unpublish

An example: `Page has_many PromoItem` and `PromoItem has_many Image`, with ownership. If you unpublish that image, it'll unpublish the `PromoItem` *and* the `Page` containing this `PromoItem`.

## Acceptance Criteria

 * A developer understands the implications of allowing versioned controls on owned records (e.g. through ModelAdmin, or opting into VersionedGridFieldEditForm elsewhere)
 * `Versioned->onAfterUnpublish` does not cascade by default
 * A developer can opt in to cascading unpublish on a per-relationship level without writing custom `onAfterUnpublish` hooks

## Tasks

- [ ] Disable recursive unpublish
- [ ] Update tests

## Pull requests
- [ ] https://github.com/silverstripe/silverstripe-versioned/pull/32

## Notes

 * I'm proposing to [make VersionedGridFieldEditForm opt-in](https://github.com/silverstripe/silverstripe-versioned/issues/21)

/cc @pitchandtone @stevie-mayhew

tractorcow:
API Introduce new hasPublishedOwners() method to toggle warnings prior to unpublish
Fixes #25

### DIFF
--- a/tests/php/VersionedOwnershipTest.php
+++ b/tests/php/VersionedOwnershipTest.php
@@ -354,7 +354,7 @@ class VersionedOwnershipTest extends SapphireTest
     }
 
     /**
-     * Test that owning objects get unpublished as needed
+     * Test that owning objects don't get unpublished when object is unpublished
      */
     public function testRecursiveUnpublish()
     {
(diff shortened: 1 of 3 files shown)


tests/php/VersionedOwnershipTest.php
```diff
@@ -354,7 +354,7 @@ public function testRecursivePublish()
     }
 
     /**
-     * Test that owning objects get unpublished as needed
+     * Test that owning objects don't unpublished owners
```
flamerohr: description reads a bit weird, maybe:
> Test that owning objects don't get unpublished when object is unpublished

PR #35, opened 2017-08-22
Change GridfieldDetailForm to be opt-in

### LINKED ISSUE #21: Change VersionedGridFieldDetailForm to opt-in
## Overview

The new `VersionedGridFieldDetailForm` will add the `VersionedGridFieldItemRequest` behaviour to any object with `has_extension(Versioned::class)` by default. This encourages individual publication of objects, even if they're related to a parent through the new ownership model. In most cases, a cascading publish without individual publications is far more user friendly (easier to form a mental model). Currently, a developers only choice is to remove this `Extension` from all GridFields. 

At the same time, even for a cascading publish, individual actions on related objects such as "unpublish" or "roll back" are desireable. They don't work the same way as cascading publish - it's common for an author to take an object related to a page off the published view back into draft mode, without unpublishing the entire page. @clarkepaul Feedback welcome :)

## Acceptance Criteria

 * As a developer I can understand the tradeoffs between cascading publishing (through ownership) and individual publication (through docs)
 * `VersionedGridFieldDetailForm` is not applied by default. I can opt in to `VersionedGridFieldDetailForm` on a per-GridField basis (regardless of ownership status)
 * Records managed in ModelAdmin get `VersionedGridFieldDetailForm` by default (not edited in the context of an ownership record)
 * I can remove/disable individual buttons in `VersionedGridFieldDetailForm` easily to enforce more granular behaviour (e.g. allowing publish, but enforcing a custom cascading unpublish)

## PRs

- [ ] https://github.com/silverstripe/silverstripe-versioned/pull/35
- [ ] https://github.com/silverstripe/silverstripe-framework/pull/7295

## Related

 * [GridField support for Versioned records](https://github.com/silverstripe/silverstripe-framework/issues/7032)
 * [RFC: Neo Versioning](https://github.com/silverstripe/silverstripe-framework/issues/4932)

tractorcow:
Fixes #21

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -25,7 +25,7 @@ before_script:
   - phpenv config-rm xdebug.ini
   - composer validate
   - composer install --prefer-dist
-  - composer require symfony/config:^3.2 silverstripe/framework:4.0.x-dev embed/embed:^3.0@stable silverstripe/assets:1.0.x-dev silverstripe/config:1.0.x-dev --prefer-dist --no-update
+  - composer require symfony/config:^3.2 silverstripe/framework:4.0.x-dev embed/embed:^3.0@stable silverstripe/assets:1.0.x-dev silverstripe/admin:1.0.x-dev silverstripe/config:1.0.x-dev --prefer-dist --no-update
   - composer update
   - "if [ \"$DB\" = \"PGSQL\" ]; then composer require silverstripe/postgresql:2.0.x-dev --prefer-dist; fi"
 
--- a/composer.json
+++ b/composer.json
@@ -22,7 +22,8 @@
         "silverstripe/framework": "^4@dev"
     },
     "require-dev": {
-        "phpunit/phpunit": "^5.7"
+        "phpunit/phpunit": "^5.7",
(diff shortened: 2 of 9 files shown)


tractorcow:
See https://github.com/silverstripe/silverstripe-framework/pull/7295 for docs.

tractorcow:
> We all know DataObject is pretty monolithic and hosts a wide array of concerns, but it seems to me this is the first member of its API to speak directly to another class. Things like summary_fields and searchable_fields primarily used for GridField, but there's a very reasonable case for using them elsewhere. So I guess without getting into a huge philosophical discussion on one static, I'm wondering two things:

It's not an API of DataObject; I didn't add any `private static` to dataobject.

> Is there better semantics we can assign this, so that it could conceivably be consumed by other APIs?

It doesn't need to be consumed by other APIs. This is a setting for a specific extension.

> Is this the best place to store this setting? Just thinking out loud here, but what if we made it GridFieldItemRequest's concern, and registered versioned DataObjects with it, rather than the other way around?

It would be a whole lot more complicated to do it that way so I chose the simplest way.

Issue #34, opened 2017-08-15
Cascade deletes should work on subsets of deleted items


tractorcow:
If I have a class A has_many B, and A cascade_deletes to B, then if I delete a subset of B records on draft, I would expect a publish of A to trigger those records to be deleted on Live also.

At the moment, B records are only deleted if A is deleted.

Specific code is in ChangeSetItem:

```php
public function findReferenced()
    {
        if ($this->getChangeType() === ChangeSetItem::CHANGE_DELETED) {
            // If we have deleted this record, recursively delete live objects on publish
            /** @var DataObject $record */
            $record = $this->getObjectInStage(Versioned::LIVE);
            if ($record) {
                // Get all cascade deletions (recursively)
                return $record->findCascadeDeletes(true);
            }
        } else {
            // If changed on stage, look at owned objects there
            $record = $this->getObjectInStage(Versioned::DRAFT);
            if ($record) {
                return $record->findOwned();
            }
        }

        // Empty set
        return new ArrayList();
    }
```

In the `else` we should also include `// Get all cascade deletes on live but not stage`. We probably will only be able to do this on getChangeType() !== CREATED (since otherwise it won't be on live).

robbieaverill:
Here's some tests for this behaviour (currently failing):

```patch
diff --git a/tests/php/VersionedCascadeDeletesTest.php b/tests/php/VersionedCascadeDeletesTest.php
index 7af8376..58e0d5a 100644
--- a/tests/php/VersionedCascadeDeletesTest.php
+++ b/tests/php/VersionedCascadeDeletesTest.php
@@ -162,4 +162,40 @@ class VersionedCascadeDeletesTest extends SapphireTest
             Versioned::get_by_stage(VersionedCascadeDeletesTest\GrandChildObject::class, Versioned::LIVE)
         );
     }
+
+    /**
+     * Test that deleting from draft then publishing removes the record from live
+     */
+    public function testDeleteDraftThenPublishRemovesFromLive()
+    {
+        /** @var VersionedCascadeDeletesTest\ParentObject $parent */
+        $parent = $this->objFromFixture(VersionedCascadeDeletesTest\ParentObject::class, 'parent1');
+        $parent->publishRecursive();
+
+        $parent->Children()->first()->delete();
+
+        $this->assertEmpty(
+            Versioned::get_by_stage(
+                VersionedCascadeDeletesTest\ChildObject::class,
+                Versioned::DRAFT,
[...]

PR #41, opened 2017-08-29
ENHANCEMENT Ensure that partial deletes of cascade_deletes get unpublished

tractorcow:
Practical use case of this example: In userforms, if you delete a field and publish the parent, now the deleted field is unpublished via changeset.


tractorcow:
Oops, forgot that this closes #34

PR #42, opened 2017-09-04
FIX Check for Versioned extension before looking for isOnLiveOnly

robbieaverill:


tractorcow:
I've written some tests for this at https://github.com/silverstripe/silverstripe-versioned/pull/43

PR #43, opened 2017-09-05
BUG Fix issue with cascading deletes to unversioned objects

tractorcow:
Replaces https://github.com/silverstripe/silverstripe-versioned/pull/42 with some tests


tractorcow:
Pushed up a fix to CI.

PR #45, opened 2017-09-25
Select stage with explicit get param

dhensby:
Depends on https://github.com/silverstripe/silverstripe-framework/pull/7399

This moves current stage setting to a GET param instead of relying on session state.

This means users can open multiple tabs and have more control over whether they are seeing the draft site or the live site and reduces the confusion for users when they unexpectedly see stage sites in other tabs when the CMS is open.


src/Versioned.php
```diff
@@ -1291,7 +1291,7 @@ public function canViewVersioned($member = null)
         }
 
         // Bypass if site is unsecured
-        if (Controller::has_curr() && Controller::curr()->getRequest()->getSession()->get('unsecuredDraftSite')) {
+        if (Versioned::get_stage() === $stage) {
```
dhensby: This is a bit of a weird change, but basically if you're trying to look at the current stage, then it's assumed that's ok (ie: you're not trying to switch stages)...

tractorcow:
Instead of url rewriting you could pull the stage from the referer header, and do a 302 redirect.

e.g.
land: `mypage/?stage=Stage`
click link: `anotherpage`
page sees `stage` in referer, 301 to `anotherpage/?stage=Stage`.

And so on :)

dhensby:
That feels much more fragile because if a browser fails to send the referrer through then you get a surprise change of stage.

I could also see this having implications when users aren't authorised and getting stuck in some kind of loop

tractorcow:
I think a link missing the querystring arg is going to be a lot more fragile; You are relying on a feature that rewrites links outside of silverstripe, and is going to act a lot less predictably and controllably.

tractorcow:
I need to spend some time looking into url_rewriter to gather my feelings on whether or not I feel it's appropriate. It feels like something that should be a middleware (rewrites HTTPResponse::body()).

chillu:
Good stuff! Didn't even know PHP had a built-in link rewriter! I'm looking to get this over the line fairly quickly, see https://github.com/silverstripe/silverstripe-cms/issues/1578 for context.

Todo from my perspective:
- Make opt-in to avoid breaking APIs in 4.x and give people a chance to fix any custom logic
- Implement `stage` param in `SiteTree->Link()` via `Versioned->updateLink()`. It's not as reactive as url rewriting before output, and hence will catch more use cases (API responses, `data-*` HTML attributes, etc).
- Document implementation (incl. surprising things like adding `<input>` tags to forms)
- Document upgrading considerations: 
  - Describe opt-in considerations: Generate links on the server, use `$this->extend('updateLink')` in your own `Link()` implementations on versioned objects
  - Usage of `Controller::join_links()` to gracefully concat URLs that might already have GET params. 
- Backport to 3.x (with opt-in)
- Remove opt-in for 5.x, make default (and adjust documentation accordingly)
- Sanity check performance of PHP's url_rewriter feature
- Check that the shareddraftcontent module still works

chillu:
I'm running PHP 7.1.12 and can't even get their example on https://secure.php.net/manual/en/outcontrol.examples.rewrite.php to work. My output:

```
<a href="//php.net/index.php?bug=1234">bug1234</a>
<form action="https://php.net/?bug=1234&edit=1" action="post"><input type="hidden" name="test" value="value" />
 <input type="text" name="title" />
</form>
```

That page also states:

>  Since PHP 7.1.0, output rewrite functions have it's own INI settings, url_rewriter.tags and url_rewriter.hosts. 

We're relying on both of these settings, specifically defining a host name that matches to the `Director` context. And we still support PHP 5 in SS4 until it's officially going out of support. Hence I don't think we can rely on this feature. I guess by making it opt in for SS4, we could make this opt-in dependant on you running PHP 7.1?

tractorcow:
Ok I had a bit of an offline chat with @ingo and we feel that maybe this "global rewrite" solution may not be the safest.

@chillu has discussed trying out an alternative solution:
 - An extra extension with `updateLink` method that decorates querystring args for links.
 - Apply to RequestHandler, ensure that RequestHandler::Link() calls this extension.
 - Apply to DataObject, ensuring that any subclass with a Link method that invokes `updateLink` will have the appropriate link.
 - Document that objects / controllers that generate links should either call a decorated method, or should have the `updateLink` extension added after the fact.

chillu:
@dhensby I've updated your PRs for framework and versioned, and rolled back that url rewriting change - hope that's OK.

> Apply to DataObject, ensuring that any subclass with a Link method that invokes updateLink will have the appropriate link.

I don't think we should add a default `DataObject->Link()`, since that would imply that all objects are linkable from an API perspective. I've explained in Versioned markdown docs how I see this working instead. It's not ideal to force users to apply `VersionedRequestHandlerExtension` to any controllers that deal with versioned objects - maybe we should just apply that to all request handlers?

src/VersionedRequestHandlerExtension.php
```diff
@@ -0,0 +1,31 @@
...
+     * Does not concern itself with verifying permissions for performance reasons.
+     *
+     * This should also pull through to form actions.
+     *
+     * @param String $link
+     * @param String $action
+     */
+    public function updateLink(&$link)
+    {
+        if(Versioned::get_stage() === Versioned::DRAFT) {
```
chillu: This will always add the `stage` param, even when `use_session=true`. That's getting the best of both worlds: Unique URLs per content variation (draft and live), and graceful legacy behaviour where URLs are constructed outside of this logic.

tests/php/VersionedRequestHandlerExtensionTest.php
```diff
@@ -0,0 +1,43 @@
...
+    public function testUpdateLinkAddsStageParamsInDraftMode()
+    {
+        Versioned::set_stage(Versioned::DRAFT);
+        $ext = new VersionedRequestHandlerExtension();
+        $link = 'my-relative-link/?some=var';
+        $ext->updateLink($link);
+        $this->assertEquals('my-relative-link/?some=var&stage=' . Versioned::DRAFT, $link);
+    }
+
+    public function testUpdateLinkAddsStageParamsOnlyOnceInDraftMode()
```
chillu: Maybe this is overkill and should be only tested in Controller::join_links()?

dhensby:
> I think a link missing the querystring arg is going to be a lot more fragile; You are relying on a feature that rewrites links outside of silverstripe, and is going to act a lot less predictably and controllably.

If we don't use PHP's build in link re-writer we are going to have a much more fragile situation because we can't be sure all links will be generated with the appended query string. If we are worried that PHP's built in rewriter will be too fragile (which seems like a strange stance to me, because how many PHP features do we not use because we think they are fragile despite them being fully documented and supported?),then we do have a link re-writer in core, so we might be able to use it?

If we are worried about fragilility of relying on a url parameter / the possibility of some links missing the query string, then I think using a decorator to update all `Link` methods seems like a worse solution than rewriting links as we start pushing the responsibility to devs to make sure all their `Link` methods call data extension hooks and correctly append versioned params.

src/Versioned.php
```diff
@@ -390,6 +401,27 @@ public function updateInheritableQueryParams(&$params)
...
+     * record in draft stage, they can also view other draft records.
+     * Does not concern itself with verifying permissions for performance reasons.
+     *
+     * @param String $link
+     * @param String $action
+     * @param String $relativeLink
+     */
+    public function updateLink(&$link)
+    {
+        if(self::get_stage() === self::DRAFT) {
```
dhensby: why are we using `self::DRAFT` here and `Versioned::DRAFT` below?

src/Versioned.php
```diff
@@ -390,6 +401,27 @@ public function updateInheritableQueryParams(&$params)
...
+     * Auto-append current stage if we're in draft,
+     * to avoid relying on session state for this,
+     * and the related potential of showing draft content
+     * without varying the URL itself.
+     *
+     * Assumes that if the user has access to view the current
+     * record in draft stage, they can also view other draft records.
+     * Does not concern itself with verifying permissions for performance reasons.
+     *
+     * @param String $link
```
dhensby: `string` needs to be lowercase

src/VersionedRequestHandlerExtension.php
```diff
@@ -0,0 +1,31 @@
...
+     * and the related potential of showing draft content
+     * without varying the URL itself.
+     *
+     * Assumes that if the user has access to view the current
+     * record in draft stage, they can also view other draft records.
+     * Does not concern itself with verifying permissions for performance reasons.
+     *
+     * This should also pull through to form actions.
+     *
+     * @param String $link
```
dhensby: Lower case `s` in `string`

chillu:
> If we don't use PHP's build in link re-writer we are going to have a much more fragile situation because we can't be sure all links will be generated with the appended query string. 

The highest impact of this is that authors won't see their draft changes in an app that doesn't follow best practice around using/implementing `Link()`. And that will be mostly limited to apps with custom versioned objects that have their own views, or muck around with `SiteTree.URLSegment` directly. So I'd say that's 1 in 50 apps maybe?

> If we are worried that PHP's built in rewriter will be too fragile (which seems like a strange stance to me, because how many PHP features do we not use because we think they are fragile despite them being fully documented and supported?),then we do have a link re-writer in core, so we might be able to use it?

I generally have pretty low expectations on PHP's lesser used APIs ;) In this case, it's simply not working for my installation. Does that example work in yours? And even if it would, we need the `url_rewriter.hosts` setting in all SilverStripe projects, but it's only available in PHP 7.1+. We could default to `use_session=true` and make this opt-in with the provision of "PHP 7.1+ only", but that'll get a lot less coverage.

> If we are worried about fragilility of relying on a url parameter / the possibility of some links missing the query string, then I think using a decorator to update all Link methods seems like a worse solution than rewriting links as we start pushing the responsibility to devs to make sure all their Link methods call data extension hooks and correctly append versioned params.

I disagree. It centralises the effort where it's obvious, at the origin (link generation). We should strive to make this as easy as possible. I've suggested just applying the `VersionedRequestHandlerExtension` to all `RequestHandler` to do this. If you do some HTML body rewriting magic with PHP, you'll still miss MIME types other than `text/html`, which confuses devs more ("it worked with rendered pages but doesn't if I get the page data as JSON")

tractorcow:
Also, using the middleware approach means it's impossible to create non-draft links on the draft site. E.g. `<a href="/">back to live site</a>`

src/Versioned.php
tractorcow: We haven't considered CMS links which don't need `?stage=` querystring args.

tractorcow: Also ignoring non-stage modes (e.g. archive, which is equally important for history viewer)

src/Versioned.php
tractorcow: Not all those with CMS access (which is in draft stage) are allowed to view draft records, so this isn't correct.

Remember that choose_site_stage() is not the ONLY way that current stage can be set.

tractorcow: Shifting into get_draft_site_secured() method, bootstrapped by that session var, but NOT if use_session is off (new default).

Another possible security flaw covered. :)

tractorcow:
Ok, I've done a bit of edge-case analysis and made some suggested API tweaks in a separate commit.

See https://github.com/silverstripe/silverstripe-admin/pull/467 for new admin PR to go with this.

tests/php/VersionedRequestHandlerExtensionTest.php
tractorcow: I've shifted this logic to the extension; It's now idempotent, and won't attempt to decorate links that already exist.

tractorcow:
Ok, I've got versioned tests passing, but there are regressions in other modules, mostly due to versioned test state.

Apparently VersionedTestState has been doing nothing due to a typo in the yml file. Fixing this modified persistence of stage between tests (has been persisting, but shouldn't).

Looking into CMS tests.

tractorcow:
Getting closer to green. Started fixing CMS module regressions at https://github.com/silverstripe/silverstripe-cms/pull/2129

dhensby:
> The highest impact of this is that authors won't see their draft changes in an app that doesn't follow best practice around using/implementing Link(). And that will be mostly limited to apps with custom versioned objects that have their own views, or muck around with SiteTree.URLSegment directly. So I'd say that's 1 in 50 apps maybe?

Almost every project I work on has a DO with a `Link` method and I've never added a `updateLink` hook to those; I don't really expect modules to need to modify my links... I suppose that's going to have to be a change in expectations going forward, now.

> I generally have pretty low expectations on PHP's lesser used APIs ;) In this case, it's simply not working for my installation. Does that example work in yours? And even if it would, we need the url_rewriter.hosts setting in all SilverStripe projects, but it's only available in PHP 7.1+. We could default to use_session=true and make this opt-in with the provision of "PHP 7.1+ only", but that'll get a lot less coverage.

I haven't implemented that exact example, however I did test the PRs I made and they worked as expected.

dhensby:
What happens if we add the get param to the base tag for pages too? Will that cover the majority of links on a page?

src/Versioned.php
```diff
@@ -83,6 +83,22 @@ class Versioned extends DataExtension implements TemplateGlobalProvider, Resetta
...
+    /**
+     * Set if draft site is secured or not. Fails over to
+     * $draft_site_secured if unset
+     *
+     * @var bool|null
+     */
+    protected static $is_draft_site_secured = null;
+
+    /**
+     * Default config for $is_draft_site_secured
```
chillu: Maybe define what "secured" means in this context? I assume it's something like:

> Enforces canView() checks, which are extended through this class to check for permission codes defined through `non_live_permissions` on the owner class.

src/Versioned.php
```diff
@@ -83,6 +83,22 @@ class Versioned extends DataExtension implements TemplateGlobalProvider, Resetta
      */
     protected static $cache_versionnumber;
 
+    /**
+     * Set if draft site is secured or not. Fails over to
+     * $draft_site_secured if unset
+     *
+     * @var bool|null
+     */
+    protected static $is_draft_site_secured = null;
```
chillu: `unsecuredDraftSite` vs. `is_draft_site_secured` - using negation in half of the API only, which gets a bit confusing. I'd say let's switch from `unsecuredDraftSite` to `isDraftSiteSecured` in the session var, it's not exactly a public API, and quite likely only used in the shareddraftcontent module.

src/Versioned.php
```diff
@@ -1480,16 +1483,16 @@ public function canViewVersioned($member = null)
      * @param Member $member
      * @return bool
      */
-    public function canViewStage($stage = 'Live', $member = null)
+    public function canViewStage($stage = self::LIVE, $member = null)
     {
-        $oldMode = Versioned::get_reading_mode();
-        Versioned::set_stage($stage);
+        return static::withVersionedMode(function () use ($stage, $member) {
```
chillu: Very elegant! :)

src/VersionedStateExtension.php
```diff
@@ -0,0 +1,102 @@
...
+     * Does not concern itself with verifying permissions for performance reasons.
+     *
+     * This should also pull through to form actions.
+     *
+     * @param string $link
+     */
+    public function updateLink(&$link)
+    {
+        // Skip if link already contains reading mode
+        if ($this->hasVersionedQuery($link)) {
```
chillu: This prevents us from forcing a different stage in the link, or changing the archivedate as part of the request processing. Given that `Controller::join_links()` already merges duplicate params on key, I don't see why this is necessary? I've pushed a change to fix this, and clarify how to force a live link in the docs of the framework PR

tractorcow:
Ok, after some intensive testing I've got this more or less working with most of the modules green (all of the ones I've tested at least).

Tomorrow I'll finish testing framework and run behat and we can merge this in.

src/Versioned.php
tractorcow: secured as in "non-authenticated users are rejected".

src/Versioned.php
chillu: Fair enough!

src/VersionedStateExtension.php
chillu: Discussed separately with Damian, we've agreed that `join_links()` should be used *outside* of this behaviour for link overrides, and keeping `updateLink()` with side effect free and predictable behaviour. I've documented this accordingly

tests/php/ReadingModeTest.php
```diff
@@ -0,0 +1,145 @@
...
+            "Convert {$readingMode} from querystring encoded string"
+        );
+    }
+
+    /**
+     * Return list of reading modes in order:
+     *  - reading mode string
+     *  - dataquery params array
+     *  - query string array
+     *  - query strig (string)
```
dhensby: `strig`?

tests/php/ReadingModeTest.php
```diff
@@ -0,0 +1,145 @@
...
+     * @return array
+     */
+    public function providerReadingModes()
```
dhensby: can we name it *provideReadingModes*? 👼

tractorcow: ok done.

tests/php/ReadingModeTest.php
tractorcow: Fixed.

tractorcow:
I'm happy with this PR now; I've added some more unit tests to cover more edge cases (e.g. ensure that non-versioned dataobjects ALSO get ?stage querystring args).

tractorcow:
I can't figure out why this fails on travis but passes locally.

tractorcow:
I think we may need to merge https://github.com/silverstripe/silverstripe-framework/pull/7399 first and then this. That will be why.

chillu:
Ran PHPUnit tests locally with merged framework PR, passing here.

Issue #22, opened 2017-07-05
Indicate draft/publish state in GridField list and detail view


chillu:
## Overview

Versioned records can now be managed by a GridField in a default configuration (via `VersionedGridFieldDetailForm`). SS4 encourages devs to version all their records. But there's no indication to authors if records are versioned in the first place, and in which state they're at the moment.

## Acceptance Criteria

 * A GridField has draft/live indicators in the list view
 * The GridField column can be removed like any other column (e.g. if a more detailed status is required)
 * The same applies to ModelAdmin search results

## Notes

 * GridField column space is precious, so we need to find a space saving way to implement this. We could use the orange circle from the files area, and add a text representation on hover/touch?
 * Some GridFields will have added their own status columns already. Duplication is an inevitable side effect of creating better defaults here

## Related

 * [GridField support for versioned records](https://github.com/silverstripe/silverstripe-framework/issues/7032)

## Pull requests 

~https://github.com/silverstripe/silverstripe-versioned/pull/49~
- [ ] https://github.com/silverstripe/silverstripe-framework/pull/7422
- [x] https://github.com/silverstripe/silverstripe-admin/pull/282

chillu:
@clarkepaul This needs UX feedback and design :)

clarkepaul:
Okay added to my todo list

chillu:
@newleeland @clarkepaul This is in beta4, so will be actioned next week - can you please work on this with high priority?

newleeland:
Potential Prototype: https://invis.io/P3DGEM4TC#/252875113_Gridfield_Listing_Holder_Draft

clarkepaul:
@chillu, @jared has just installed the Framework Test Module as that was a better representation of what a normal grid field looks like, so he can work from that and update the designs to reflect.

newleeland:
@chillu this is the updated design: https://invis.io/P3DGEM4TC#/252875113_Gridfield_Listing_Holder_Draft

@phalkunz mentioned that it could be complicated with getting a status tag on the page toolbar to be influenced by gridfield items.

chillu:
Thanks for considering the inline filtering. That's out of scope for this card, but an interesting direction. I think we should generally consider decoupling GridField filtering from the available columns, it looks nice, but is way too restrictive. E.g. you can't cram a date range into a single input field (easily). Discussion for another day.

[...]

PR #49, opened 2017-10-03
Add get status flags method

phalkunz:
Fixes https://github.com/silverstripe/silverstripe-versioned/issues/22

### DIFF
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -2561,4 +2561,13 @@ class Versioned extends DataExtension implements TemplateGlobalProvider, Resetta
     {
         return $this->mode === static::STAGEDVERSIONED;
     }
+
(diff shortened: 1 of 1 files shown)


src/Versioned.php
```diff
@@ -2561,4 +2561,13 @@ public function hasStages()
...
         return $this->mode === static::STAGEDVERSIONED;
     }
+
+    public function getStatusFlags() {
+        if($this->isOnDraftOnly()) {
+            return [ 'draft' ];
+        }
+        else if($this->isModifiedOnDraft()) {
+            return [ 'modified' ];
+        }
```
flamerohr: I don't think this method should be added to the Versioned public API...
this could easily be a protected method in the `GridFieldVersionedState` class you've created.

PR #52, opened 2017-10-05
API Support stateless versioned extensions

tractorcow:
Parent story: https://github.com/silverstripe/silverstripe-framework/issues/7382

Requires https://github.com/silverstripe/silverstripe-framework/pull/7437

### DIFF
--- a/src/VersionableExtension.php
+++ b/src/VersionableExtension.php
@@ -24,4 +24,13 @@ interface VersionableExtension
      * @param array $indexes List of indexes in this model
      */
     public function updateVersionableFields($suffix, &$fields, &$indexes);
+
+    /**
+     * Modify table name with suffix.
+     * Should return $table if not modified.
+     *
+     * @param string $table
+     * @return string
+     */
+    public function extendWithSuffix($table);
 }
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -74,19 +74,19 @@ class Versioned extends DataExtension implements TemplateGlobalProvider, Resetta
     const DRAFT = 'Stage';
 
     /**
-     * A version that a DataObject should be when it is 'migrating',
-     * that is, when it is in the process of moving from one stage to another.
-     * @var string
-     */
-    public $migratingVersion;
-
-    /**
      * A cache used by get_versionnumber_by_stage().
      * Clear through {@link flushCache()}.
      *
      * @var array
      */
     protected static $cache_versionnumber;
+
+    /**
+     * Cache of version to modified dates for this object
+     *
+     * @var array
+     */
+    protected $versionModifiedCache = [];
 
     /**
      * Current reading mode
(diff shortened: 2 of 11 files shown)


src/Versioned.php
```diff
@@ -926,8 +915,8 @@ public function augmentWrite(&$manipulation)
         }
 
         // Clear the migration flag
-        if ($this->migratingVersion) {
-            $this->migrateVersion(null);
+        if ($migratingVersion) {
+            $this->setMigratingVersion(null);
```
dhensby: Here you're using `$this` but in the deprecated methof `migrateVersion` you rely on `$owner`

src/Versioned.php
```diff
@@ -964,7 +971,7 @@ public function onAfterWrite()
      */
     public function onAfterSkippedWrite()
     {
-        $this->migrateVersion(null);
+        $this->setMigratingVersion(null);
```
dhensby: Again, should this use owner?

tractorcow: Owner is already set. It doesn't need to be re-set again.

src/Versioned.php
tractorcow: Yes that's right. Can you please explain the problem with this?

src/Versioned.php
tractorcow: setMigratingVersion internally uses owner. I could call it on owner here but it's probably unnecessary stack depth.

tractorcow: I think I see your point.

In versioned where I want a method to be overridable in the extended class I use `$this->owner->method` instead of `$this->method`, although both are substitutable.

However we don't need to allow these to be overridden so I've kept it simple.

tractorcow:
Ok fixed up with your feedback. :)

src/Versioned.php
flamerohr: I think here is fine, since the new `setMigratingVersion()` is relying on `$this->owner`

src/Versioned.php
```diff
@@ -1762,13 +1772,34 @@ public function copyVersionToStage($fromStage, $toStage, $createNewVersion = fal
...
+
+    /**
+     * @deprecated 4.0...5.0
      * @param string $version The version.
      */
     public function migrateVersion($version)
     {
-        $this->migratingVersion = $version;
+        Deprecation::notice('5.0', 'use setMigratingVersion instead');
+        $this->owner->setMigratingVersion($version);
```
flamerohr: I think this should not rely on `$this->owner`

flamerohr:
I should've refreshed more often :D posted comments after the changes were made

src/VersionableExtension.php
```diff
@@ -24,4 +24,13 @@ public function isVersionedTable($table);
...
     public function updateVersionableFields($suffix, &$fields, &$indexes);
+
+    /**
+     * Modify table name with suffix.
+     * Should return $table if not modified.
+     *
+     * @param string $table
+     * @return string
+     */
+    public function extendWithSuffix($table);
```
flamerohr: this could also be a prefix right?
I mean, unless it was explicitly looking for a suffix e.g. return a blank string if no suffix, instead of returning $table
This seems more suitably named as `augmentTableName()`

tractorcow: I'm not honestly sure why this method exists; All I did was add the missing declaration to the interface.

Issue #53, opened 2017-10-06
UploadField with many_many on Image causes too many queries


DorsetDigital:
On a fresh install of Silverstripe 4 Beta 3:

Create a simple page class, extending Page with a many_many relationship to the Image class:

```
namespace DorsetDigital\SS4\VersionTest;

use SilverStripe\AssetAdmin\Forms\UploadField;
use SilverStripe\Assets\Image;

class ManyManyPage extends \Page
{

 private static $many_many = [
     'ManyManyPhotos' => Image::class
 ];
 private static $owns = ['ManyManyPhotos'];

 public function getCMSFields()
 {
  $folderName = 'TestPhotos';
  $galleryUploads = new UploadField('ManyManyPhotos');
  $galleryUploads->setFolderName($folderName);
  $galleryUploads->setAllowedFileCategories('image/supported');

  $fields = parent::getCMSFields();
  $fields->addFieldToTab('Root.Images', $galleryUploads);
  return $fields;
 }
}
```

As you add images to the page, the number of SQL queries rises dramatically (hundreds of queries, possibly even thousands), getting to the point where the web server times out during the save process and the page remains unpublished.   A snippet of the query log is attached.

[query_log_snippet.txt](https://github.com/silverstripe/silverstripe-versioned/files/1362478/query_log_snippet.txt)

# PRs

- [x] https://github.com/silverstripe/silverstripe-versioned/pull/59
- [x] https://github.com/silverstripe/silverstripe-framework/pull/7489

DorsetDigital:
I'm not sure if it's helpful, but as part of the testing, I tried the same with a has_one relationship.   Attached is a log dump of the queries involved in adding an image to the has_one on a different page.
[ho_query.log](https://github.com/silverstripe/silverstripe-versioned/files/1362541/ho_query.log)

chillu:
How are you adding images to the page? Are you looping over them in templates? Or is this just rendering the UploadField in the CMS? Having 15 duplicates of the same query is indeed very concerning. It doesn't look like that's a permission check issue either (discussed in https://github.com/silverstripe/silverstripe-framework/issues/7442)

DorsetDigital:
The images are just being dragged-n-dropped onto the uploadField in CMS.     There are no templates in this test site, this is purely CMS behaviour.   A lot of queries fire as the files are uploading, but these don't seem to cause any major issues in terms of the responsiveness of the UI.   The big issue is when the save / publish button is clicked.

[...]

PR #59, opened 2017-10-18
ENHANCEMENT Reduce number of queries used outside of caches

tractorcow:
Parent issue https://github.com/silverstripe/silverstripe-versioned/issues/53

Pretty much every one of the stage / live checks can be built on top of get_versionnumber_by_stage() method, which has its own internal cache.


src/Versioned.php
```diff
@@ -2203,10 +2191,11 @@ public static function get_versionnumber_by_stage($class, $stage, $id, $cache =
                 self::$cache_versionnumber[$baseClass][$stage] = [];
             }
 
-            self::$cache_versionnumber[$baseClass][$stage][$id] = $version;
+            // Internally store nulls as 0
+            self::$cache_versionnumber[$baseClass][$stage][$id] = $version ?: 0;
```
flamerohr: maybe document that `0` means `does not exist` against this variable at the top of the class?

it's a neat approach - but certainly not obvious at first glance.

tractorcow:
Done.

src/Versioned.php
dhensby: Does a save and/or publish event clear the cache for the record?

tractorcow: Yes it does. All writes call flushCache().

PR #60, opened 2017-10-23
FIX Don’t override ItemRequest if already versioned

wilr:
If a developer customises the ItemRequest don’t override it as long as it is already a versioned item hander.

### DIFF
--- a/src/VersionedGridFieldDetailForm.php
+++ b/src/VersionedGridFieldDetailForm.php
@@ -24,17 +24,9 @@ class VersionedGridFieldDetailForm extends Extension
         if ($record
             && $record->has_extension(Versioned::class)
             && $record->config()->get('versioned_gridfield_extensions')
+            && (!$class || !is_subclass_of($class, VersionedGridFieldItemRequest::class))
         ) {
-            // don't override custom classes if they already subclass this
-            if (!$class) {
-                $class = VersionedGridFieldItemRequest::class;
-
-                return;
-            }
-
-            if (!is_subclass_of($class, VersionedGridFieldItemRequest::class)) {
-                $class = VersionedGridFieldItemRequest::class;
-            }
+            $class = VersionedGridFieldItemRequest::class;
         }
     }
 }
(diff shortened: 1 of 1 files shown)


src/VersionedGridFieldDetailForm.php
```diff
@@ -20,12 +21,27 @@ class VersionedGridFieldDetailForm extends Extension
...
-            $class = VersionedGridFieldItemRequest::class;
+            // don't override custom classes if they already subclass this
+            if (!$class) {
+                $class = VersionedGridFieldItemRequest::class;
+
+                return;
+            }
+
+            try {
+                $obj = new ReflectionClass($class);
```
dhensby: You don't need to use `Reflection` to find out if a class is a subclass of another. `is_subclass_of` or `is_a` will do.

wilr: Ah night! I had thought those only worked on instances of `$class` rather than just names. Nice change in 5.3.9

src/VersionedGridFieldDetailForm.php
```diff
@@ -20,12 +20,21 @@ class VersionedGridFieldDetailForm extends Extension
...
     {
-        // Conditionally use a versioned item handler
+        // Conditionally use a versioned item handler if it doesn't already have one.
         if ($record
             && $record->has_extension(Versioned::class)
             && $record->config()->get('versioned_gridfield_extensions')
         ) {
-            $class = VersionedGridFieldItemRequest::class;
+            // don't override custom classes if they already subclass this
+            if (!$class) {
```
dhensby: These conditions can be combined

wilr: How'd you mean?

robbieaverill: @dhensby Will's on leave at the moment - I assume you mean that the if statements in this can be consolidated into one, so I've pushed a fix for that. Would you mind re-checking when you have time please?

If this is a bug fix, do you think we could sneak it into 1.0 before a stable release?

robbieaverill:
@dhensby Will's on leave at the moment - I assume you mean that the if statements in this can be consolidated into one, so I've pushed a fix for that. Would you mind re-checking when you have time please?

If this is a bug fix, do you think we could sneak it into 1.0 before a stable release?

I hope we can, so have switched the branch to 1.0.

dhensby:
much better, thanks

PR #74, opened 2017-12-12
API Split ownership and recursive publishing into RecursivePublishable

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/7359

Framework docs update at https://github.com/silverstripe/silverstripe-framework/pull/7695


src/RecursivePublishable.php
```diff
@@ -0,0 +1,310 @@
...
+        $schema = DataObject::getSchema();
+        $ownedHasMany = array_intersect($owns, array_keys($hasMany));
+        foreach ($ownedHasMany as $relationship) {
+            // Check the owned object is actually versioned and staged
+            $joinClass = $schema->hasManyComponent(get_class($owner), $relationship);
+            $joinInstance = DataObject::singleton($joinClass);
+
+            /** @var Versioned $versioned */
+            $versioned = $joinInstance->getExtensionInstance(Versioned::class);
+            if (!$versioned || !$versioned->hasStages()) {
```
flamerohr: is it not enough to check `Extensible::has_extension($joinClass, Versioned::class)`?

seems a lot more verbose but I don't see the gain in this approach

tests/php/VersionedOwnershipTest.php
```diff
@@ -347,12 +374,93 @@ public function testRecursivePublish()
...
+        /** @var VersionedOwnershipTest\OwnedByUnversioned $book1 */
+        $book1 = $this->objFromFixture(VersionedOwnershipTest\OwnedByUnversioned::class, 'book1');
+        /** @var VersionedOwnershipTest\OwnedByUnversioned $book2 */
+        $book2 = $this->objFromFixture(VersionedOwnershipTest\OwnedByUnversioned::class, 'book2');
+        /** @var VersionedOwnershipTest\OwnedByUnversioned $book3 */
+        $book3 = $this->objFromFixture(VersionedOwnershipTest\OwnedByUnversioned::class, 'book3_published');
+
+        // Remove book3 from draft
+        $book3ID = $book3->ID;
+        $book3->delete();
```
flamerohr: add an assert that `book3` is published

src/RecursivePublishable.php
tractorcow: I also realised that the object could be versioned, but without staging enabled (hence the hasStages() check).

tests/php/VersionedOwnershipTest.php
tractorcow: Done.

src/RecursivePublishable.php
flamerohr: I didn't even remember that was a thing :D

filiplikavcan:
I created simple extension which publishes all assets has_one components that are marked as owned by the non-versioned data object: https://github.com/bratiask/own-assets

Issue #64, opened 2017-11-02
GridField extensions for versioned DataObjects in ModelAdmin should be opt-out, not opt-in


colintucker:
After a healthy discussion over on Slack with @tractorcow and others, there is some confusion and frustration in the community about why the automatic detection of `Versioned` objects in `ModelAdmin` was removed, and made opt-in with the `$versioned_gridfield_extensions` private static.

This seems counter-intuitive to me.  The previous behavior in SS4 was that versioned data objects would be automatically detected and the appropriate CMS actions added to the `ModelAdmin` for said object.  It's a case of just adding the `Versioned` extension to your model, and boom, you're off and racing.

Having to tell developers "oh by the way, you'll also need to opt-in to having the admin handle your versioning" just seems unnecessary.  The bar to entry in SS4 has already increased, and it's clear that the versioning of assets in SS4 has created some confusion among devs.  Changes like this only make it more confusing, in my humble opinion.  I believe it should actually be the opposite, that is, opt-out, instead of opt-in.  Thoughts?

## Pull request
 - [x] https://github.com/silverstripe/silverstripe-versioned/pull/80

tractorcow:
Note the issue discussed here is https://github.com/silverstripe/silverstripe-versioned/issues/21

My suggestion is that we should probably turn this on by default for ModelAdmin by default, since it's the "top level" section, similar to how cms pages section has save / publish at the top level.

adrexia:
The acceptance criteria of #21 says `Records managed in ModelAdmin get VersionedGridFieldDetailForm by default (not edited in the context of an ownership record)`. With that change this would feel less like a bug.

tractorcow:
Ah that's right, so we can make this a bug after all. :) Thanks for spotting that @adrexia

PR #80, opened 2017-12-22
Pulls/1/gridfield versioned default

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-versioned/issues/64

### DIFF
--- a/_config/versionedgridfield.yml
+++ b/_config/versionedgridfield.yml
@@ -4,3 +4,5 @@ Name: versionedgridfield
 SilverStripe\Forms\GridField\GridFieldDetailForm:
   extensions:
     - SilverStripe\Versioned\VersionedGridFieldDetailForm
+SilverStripe\ORM\DataObject:
+  versioned_gridfield_extensions: true


tractorcow:
No, this is NOT meant to be opt-out. See https://github.com/silverstripe/silverstripe-versioned/issues/21 for the discussion.

This should only be turned on by default for the modeladmin section, not all gridfields everywhere.

dhensby:
hmm, it's causing a lot of discord in the community that versioned objects don't automatically get versioned actions in the gridfields that display them.. why do we want to make this opt-in when it effectively breaks the gridfield actions on versioned objects?

main complaint I'm aware of is using a gridfield to delete a versioned record results in a deletion from draft site and it disappears from the gridfield, then there's no way to delete from live and so we see an inconsistent state in the CMS vs the live site.

tractorcow:
I'm just going to leave it... I had originally made it on-by-default, and was requested to turn it off by default, and have it be opt-in. It probably should just be on by default.

@phalkunz for you to look at once you get back.

dhensby:
@tractorcow you're first instinct is usually the right one ;)

PR #81, opened 2018-01-07
API Remove Versioned_Version class and Versioned::allVersions() method

tractorcow:
Related: https://github.com/silverstripe/silverstripe-framework/issues/7663 (implements 5.0 changes only)

- [x] Changelog PR: https://github.com/silverstripe/silverstripe-framework/pull/7737
- [ ] CMS PR: https://github.com/silverstripe/silverstripe-cms/pull/2065


robbieaverill:
Needs a rebase, and also could we add `@deprecated` PHPDocs in 4.1 for the methods that are removed?

tractorcow:
I don't really want to back-port the functionality to 4.1; If we deprecated it, we would need to remove usages at the same time.

robbieaverill:
Yeah fair enough

tractorcow:
Rebased.

PR #82, opened 2018-01-09
FIX Provide expected argument to onBefore/AfterPublish hooks

dhensby:
According to `SiteTreeExtension` the `onBeforePublish` and `onAfterPublish` extension hooks expect the `$original` published object to be passed as an argument:

https://github.com/silverstripe/silverstripe-cms/blob/c7214b4dcde6ac2e14ca5d451e17e96dfc739651/code/Model/SiteTreeExtension.php#L14-L30

This patch restores this functionality

### DIFF
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -1495,11 +1495,21 @@ class Versioned extends DataExtension implements TemplateGlobalProvider, Resetta
         if (!$owner->canPublish()) {
             return false;
         }
+        if ($this->isPublished()) {
+            // get the last published version
+            $baseClass = $owner->baseClass();
+            $baseTable = $owner->baseTable();
 
-        $owner->invokeWithExtensions('onBeforePublish');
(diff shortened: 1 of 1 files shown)


src/Versioned.php
```diff
@@ -1495,11 +1495,21 @@ public function publishSingle()
...
             return false;
         }
+        if ($this->isPublished()) {
+            // get the last published version
+            $baseClass = $owner->baseClass();
+            $baseTable = $owner->baseTable();
 
-        $owner->invokeWithExtensions('onBeforePublish');
+            $original = self::get_one_by_stage($baseClass, self::LIVE, [
+                "\"$baseTable\".\"ID\" = ?" => $owner->ID,
```
mikenz: We have this patched in our branch but used:
$original = Versioned::get_by_stage(get_class($this->owner), self::LIVE)->byID($owner->ID);

dhensby: `$baseClass` is needed, but that's an alternative approach, yes..

Issue #51, opened 2017-10-03
Indicate draft/publish state in GridField detail view


chillu:
# Overview

Authors need context of the current publication status when editing a record. For example, is pressing "publish" just updating an existing record, or publishing a draft for the first time? Split out from https://github.com/silverstripe/silverstripe-versioned/issues/22

# Acceptance Criteria

 * The GridField edit form indicates if the edited objects is in draft mode (meaning the "publish" action will publish this record for the first time)
 * State updates when value is saved
 * Applies to all versioned GridFields by default (incl. ModelAdmin)
 * Applies to pages editing view

# Notes

 * Excludes Files section
 * Label wraps correctly with long titles and breadcrumb chains. 
 * Excludes "last saved" info in button toolbar (unless where it already exists, like in the CMS area)
 * Excludes "propagating" draft state to owners

# Accepted Design

![page_tinymce_states left top](https://user-images.githubusercontent.com/8085295/32310538-e8b96c8c-bff7-11e7-8157-7c09c11db4ec.png)

## Pull requests

- [x] https://github.com/silverstripe/silverstripe-admin/pull/388
- ~https://github.com/silverstripe/silverstripe-framework/pull/7727~
- [x] https://github.com/silverstripe/silverstripe-versioned/pull/86

chillu:
@newleeland Can you please figure out which design we'll implement - breadcrumbs or actions?

newleeland:
![page_tinymce_states action bar](https://user-images.githubusercontent.com/8085295/32310045-7941b4ca-bff4-11e7-85f8-8dd619731f9d.png)
The issue with placing at the bottom is that, tends to last things people see, generally we want to know earlier on, furthermore theres quite a lot things on Toolbar south is quite populated already, especially on mobile view

![page_tinymce_states right top](https://user-images.githubusercontent.com/8085295/32310337-5e6d10ac-bff6-11e7-93bd-82761b81b66a.png)
It could be on the top right however, it is a little out of place in with tabs
![page_tinymce_states right top 2](https://user-images.githubusercontent.com/8085295/32310509-b2160fdc-bff7-11e7-8544-e54a02ff4b56.png)
This is another option, however, it not a consisent pattern we currently have.

clarkepaul:
[...]

PR #86, opened 2018-01-17
ENHANCEMENT Add version status badge in gridfield detail view

phalkunz:
Issue: https://github.com/silverstripe/silverstripe-versioned/issues/51

### DIFF
--- a/src/VersionedGridFieldItemRequest.php
+++ b/src/VersionedGridFieldItemRequest.php
@@ -19,6 +19,23 @@ use SilverStripe\ORM\ValidationResult;
 class VersionedGridFieldItemRequest extends GridFieldDetailForm_ItemRequest
 {
 
+    public function Breadcrumbs($unlinked = false)
+    {
+        $items = parent::Breadcrumbs($unlinked);
+        if ($this->record->hasExtension(Versioned::class) && $this->record->config()->get('versioned_gridfield_extensions')) {
+            $status = $this->getRecordStatus();
+            $lastItem = $items->last();
+            if ($status) {
+                $badge = sprintf(
+                    '<span class="badge version-status version-status--%s">%s</span>',
+                    $status['class'],
+                    $status['title']
+                );
+                $lastItem->setField('Extra', $badge);
+            }
+        }
+        return $items;
+    }
     protected function getFormActions()
     {
         // Check if record is versionable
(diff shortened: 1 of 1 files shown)


src/VersionedGridFieldItemRequest.php
```diff
@@ -19,6 +19,23 @@
 class VersionedGridFieldItemRequest extends GridFieldDetailForm_ItemRequest
 {
 
+    public function Breadcrumbs($unlinked = false)
+    {
+        $items = parent::Breadcrumbs($unlinked);
+        if ($this->record->hasExtension(Versioned::class) && $this->record->config()->get('versioned_gridfield_extensions')) {
```
tractorcow: I think we can remove the `versioned_gridfield_extensions` check; If you have this itemrequest class, you are already using the versioning extension for gridfield.

src/VersionedGridFieldItemRequest.php
```diff
@@ -19,6 +19,23 @@
...
+            $status = $this->getRecordStatus();
+            $lastItem = $items->last();
+            if ($status) {
+                $badge = sprintf(
+                    '<span class="badge version-status version-status--%s">%s</span>',
```
tractorcow: This is coming up as unreadable white text.

tractorcow:
@phalkunz I've updated the code with the requested changes.

Also please see https://github.com/silverstripe/silverstripe-framework/pull/7773; In 4.1 I'm going to move this to versioned.

src/VersionedGridFieldItemRequest.php
```diff
@@ -223,19 +226,32 @@ protected function setFormMessage($form, $message)
...
+     *
+     * @return array|null
+     */
     protected function getRecordStatus()
     {
+        /** @var DataObject|Versioned $record */
         $record = $this->record;
 
+        // No status if un-versioned
+        if (!$this->record->hasExtension(Versioned::class)) {
```
phalkunz: This needs to check `versioned_gridfield_extensions` flag on the model because user can opt out.

tractorcow: The opt out is done before VersionedGridFieldItemRequest is constructed.

```php
// Conditionally use a versioned item handler if it doesn't already have one.
        if ($record
            && $record->has_extension(Versioned::class)
            && $record->config()->get('versioned_gridfield_extensions')
            && (!$class || !is_subclass_of($class, VersionedGridFieldItemRequest::class))
        ) {
            $class = VersionedGridFieldItemRequest::class;
        }
```

tractorcow: That was why my earlier request to move it here; We only had to check this once, and the condition would apply globally. :)

phalkunz: Ah, right! 👍

src/VersionedGridFieldItemRequest.php
```diff
@@ -19,6 +21,24 @@
...
+        if ($status) {
+            // Generate badge
+            $badge = DBField::create_field('HTMLFragment', sprintf(
+                '<span class="badge version-status version-status--%s">%s</span>',
+                $status['class'],
+                $status['title']
+            ));
+            /** @var ArrayData $lastItem */
+            $lastItem = $items->last();
+            $lastItem->setField('Extra', $badge);
```
flamerohr: what happens if `Extra` was already defined by another class? Shouldn't this append rather than overwrite?

Issue #87, opened 2018-01-18
GridFieldVersionedState should be in Versioned module not framework


tractorcow:
As a follow up to https://github.com/silverstripe/silverstripe-versioned/issues/22, this component was added to framework, resulting in tight coupling between it and versioned. We should refactor this into the versioned module.

Also, none of the badges are styled properly; We can fix this at the same time.

Note: this story must be rebased / merged AFTER https://github.com/silverstripe/silverstripe-versioned/issues/51 is done as it re-uses those CSS classes.

# PRs:

 - [x] https://github.com/silverstripe/silverstripe-framework/pull/7774 (removes deprecated usage)
 - [x] https://github.com/silverstripe/silverstripe-versioned/pull/88 (main pr, adds new component)

tractorcow:
Updated based on feedback and ready to peer review.
 - Removed `published` state
 - Remove border style from inline badge

PR #88, opened 2018-01-18
API Refactor GridFieldVersionedState into versioned module

tractorcow:
Main story https://github.com/silverstripe/silverstripe-versioned/issues/87

Also adds in status for published records.

image

@clarkepaul can you please eyeball this and tell me if this is the correct gridfield style?


tractorcow:
I've just been asked to adjust this to remove the border and the "published" state. Will do this next week.

tractorcow:
Updated based on feedback and ready to peer review.
 - Removed `published` state
 - Remove border style from inline badge

tractorcow:
There was a regression in framework tests I've fixed at https://github.com/silverstripe/silverstripe-framework/pull/7782

Issue #28, opened 2017-07-13
Utilise $owns for recursive duplication


pitchandtone:
Since $owns handles versioning cascade, and should handle unpublish, we should also be able to imply duplication (similar to publish). The logic should be similar to the recursive publish method.

# Acceptance Criterea:

  - Any duplicated object will include all owned objects (or duplicates as necessary) from the original
  - Developers can customise the list of objects that are duplicated automatically
  - Documentation describes how this works
  - Tests cover any new features
  
# PRs

  - [x] https://github.com/silverstripe/silverstripe-framework/pull/7819
  - [x] https://github.com/silverstripe/silverstripe-versioned/pull/106

pitchandtone:
I have supplied an example of how we've handled this on Elemental here: https://github.com/silverstripe/silverstripe-versioned/issues/25#issuecomment-314701347

pitchandtone:
@sminnee, @chillu, @tractorcow have also made similar comments here: 
https://github.com/silverstripe/silverstripe-versioned/issues/23#issuecomment-314252517
https://github.com/silverstripe/silverstripe-versioned/issues/23#issuecomment-314253516
https://github.com/silverstripe/silverstripe-versioned/issues/23#issuecomment-314257726

tractorcow:
At the moment we have the following method signature for duplicate:

```
public function duplicate($doWrite = true, $manyMany = 'many_many')
```

What we could do is change it to

```
public function duplicate($doWrite = true, $withOwned = true)
```

So that it allows you to duplicate owned objects, rather than trying to infer it from the relationship type.

Note that many_many currently only duplicates the mapping table.

We could use the following behaviour:

 - $withOwned = true, duplicates owned many_many mapping table, duplicate actual object for has_one and has_many.
 - $withOwned = false, duplicate only this object and not related.
 
@pitchandtone would love to hear your feelings. :)

pitchandtone:
@tractorcow wouldn't matching the publish api signature be better? duplicateRecursive and duplicateSingle? and then deprecating duplicate for v5, but calling Recursive if $withOwned = true

sminnee:
+1 for API consistency.

chillu:
The ticket title reads "Utilise $owns for recursive duplication **and deletion**", but the comments only discuss duplication. On https://github.com/silverstripe/silverstripe-versioned/issues/23 the consensus seems to be that we can't infer delete/unpublish from ownership. Should we limit this ticket to duplication?

[...]

PR #106, opened 2018-01-31
API Ensure that duplications fail over to owned objects

tractorcow:
Parent story https://github.com/silverstripe/silverstripe-versioned/issues/28

Ensures that, in the absence of cascade_duplications() versioned will use `owns` to infer duplications.

### DIFF
--- a/src/RecursivePublishable.php
+++ b/src/RecursivePublishable.php
@@ -307,4 +307,32 @@ class RecursivePublishable extends DataExtension
             $disowned->execute();
         }
     }
+
+    /**
+     * If `cascade_duplications` is empty, default to `owns` config
+     *
+     * @param DataObject $original
+     * @param bool $doWrite
+     * @param array|null|false $relations
+     */
+    public function onBeforeDuplicate($original, &$doWrite, &$relations)
+    {
+        // If relations to duplicate are declared (or forced off) don't rewrite
+        if ($relations || $relations === false) {
+            return;
+        }
+
+        // Only duplicate owned relationships that are either exclusively owned,
+        // or require additional writes. Also exclude any custom non-relation ownerships.
+        $allowed = array_merge(
+            array_keys($this->owner->manyMany()), // Require mapping table duplications
+            array_keys($this->owner->belongsTo()), // Exclusive record must be duplicated
+            array_keys($this->owner->hasMany()) // Exclusive records should be duplicated
+        );
+        // Note: don't assume that owned has_one needs duplication, as these can be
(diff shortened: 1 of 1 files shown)


src/RecursivePublishable.php
```diff
@@ -307,4 +307,32 @@ public function unlinkDisownedObjects($sourceStage, $targetStage)
...
+        if ($relations || $relations === false) {
+            return;
+        }
+
+        // Only duplicate owned relationships that are either exclusively owned,
+        // or require additional writes. Also exclude any custom non-relation ownerships.
+        $allowed = array_merge(
+            array_keys($this->owner->manyMany()), // Require mapping table duplications
+            array_keys($this->owner->belongsTo()), // Exclusive record must be duplicated
+            array_keys($this->owner->hasMany()) // Exclusive records should be duplicated
```
dhensby: I'm not sure the `has_many`s should be duplicated. it's a relationship to an external record.

It's difficult because there's no one-size-fits-all approach here. Sometimes the `has_many` should be duplicated and other times it shouldn't and it really depends on the case...

With `WidgetArea`s for example, we leave it to the `WidgetArea` to duplicate the linked `Widget`s because in that instance it's what's needed (partly out of necessity because there's no interface for detaching a widget from one `WidgetArea` so changes and deletions propagate unexpectedly when using the interface.

But for ContentBlocks where they should be shared among pages, this behaviour is going to be undesired.

I think out of the box, we shouldn't be duplicating `has_many` relations...

tractorcow: This is only duplicated if you explicitly name that has_many in the cascade_duplications. It isn't duplicating all of them.

tractorcow: If it's owned (and thus necessary) and a has_many, then it's necessary to duplicate it.

You can work-around this by declaring a cascade_duplicates for that object, and then `owns` will no longer be used as the default.

tractorcow: > But for ContentBlocks where they should be shared among pages, this behaviour is going to be undesired.


ContentBlocks don't own pages so it won't be a problem there.

robbieaverill: FYI at this stage we don't support many_many for content blocks

flamerohr: Is many_many on the roadmap?
It always made sense to me to keep it unsupported for Elemental

robbieaverill: Support for it is in dnadesign/silverstripe-elemental-virtual at the moment, but I don't know about commercial support for it longer term =)

PR #109, opened 2018-02-04
BUG Prevent versioned records self-publishing

tractorcow:
I've noticed that this component was treating "save" as "save and publish"

### DIFF
--- a/src/VersionedGridFieldItemRequest.php
+++ b/src/VersionedGridFieldItemRequest.php
@@ -22,7 +22,6 @@ use SilverStripe\View\ArrayData;
  */
 class VersionedGridFieldItemRequest extends GridFieldDetailForm_ItemRequest
 {
-
     /**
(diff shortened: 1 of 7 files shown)


src/VersionedGridFieldItemRequest.php
```diff
@@ -144,6 +150,7 @@ public function doPublish($data, $form)
 
         // Initial save and reload
         $record = $this->saveFormIntoRecord($data, $form);
+        $record->publishRecursive();
```
flamerohr: hello `publishRecursive()` my old friend

tractorcow: No matter where you put it, it's never in the right place. :)

tractorcow:
Tests fixed with https://github.com/silverstripe/silverstripe-framework/pull/7832

PR #113, opened 2018-02-12
BUG Prevent nested permissions from breaking recursive publishing

### LINKED ISSUE #112: BUG Publishing recursive records sometimes fails if canPublish() = false for a child record
To replicate:

 - Page has_one / owns a File
 - Log in as a content author with only edit pages permissions
 - Edit a page, upload a file
 - Publish the page
 - Notice "server error" message (because File::canPublish() is false without 
image

 
Solution needs discussion.

fixed with https://github.com/silverstripe/silverstripe-versioned/pull/113

Also see https://github.com/silverstripe/silverstripe-framework/issues/7837 for related discussion

tractorcow:
ENHANCEMENT Modify permission checks to require outside invocation instead of verifying internally (ref: https://docs.silverstripe.org/en/4/developer_guides/model/permissions/)

Fixes #112 
Fixes https://github.com/silverstripe/silverstripe-framework/issues/7837

Note: It would be good for a @silverstripe/core-team to confirm my thinking regarding the permissions (see the link above). I feel that we should stick with the convention of requiring user code to invoke those permission checks, rather than internally failing if any are denied, causing exceptions.

### DIFF
--- a/src/ChangeSet.php
+++ b/src/ChangeSet.php
@@ -114,6 +114,8 @@ class ChangeSet extends DataObject
     /**
      * Publish this changeset, then closes it.
      *
+     * User code should call {@see canPublish()} prior to invoking this method.
+     *
      * @throws Exception
      * @return bool True if successful
      */
@@ -129,9 +131,6 @@ class ChangeSet extends DataObject
             throw new ValidationException(
                 "ChangeSet does not include all necessary changes and cannot be published."
             );
-        }
-        if (!$this->canPublish()) {
-            throw new LogicException("The current member does not have permission to publish this ChangeSet.");
         }
 
         DB::get_conn()->withTransaction(function () {
@@ -394,7 +393,8 @@ class ChangeSet extends DataObject
         if (!$member) {
             $member = Security::getCurrentUser();
         }
-        foreach ($this->Changes() as $change) {
+        // Check all explicitly added items
+        foreach ($this->Changes()->filter(['Added' => ChangeSetItem::EXPLICITLY]) as $change) {
             /** @var ChangeSetItem $change */
             if (!$change->canPublish($member)) {
                 return false;
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -1487,14 +1487,13 @@ class Versioned extends DataExtension implements TemplateGlobalProvider, Resetta
     /**
      * Publishes this object to Live, but doesn't publish owned objects.
      *
+     * User code should call {@see canPublish()} prior to invoking this method.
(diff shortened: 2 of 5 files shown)


src/ChangeSet.php
```diff
@@ -114,6 +114,8 @@ class ChangeSet extends DataObject
     /**
      * Publish this changeset, then closes it.
      *
+     * User code must call {@see canPublish()} prior to invoking this method.
```
flamerohr: "should call" rather than "must call" right?
since CLI invocations could just forego permission checks

src/Versioned.php
```diff
@@ -1487,14 +1487,13 @@ public function publishRecursive()
     /**
      * Publishes this object to Live, but doesn't publish owned objects.
      *
+     * User code must call {@see canPublish()} prior to invoking this method.
```
flamerohr: Same comment as above

src/Versioned.php
```diff
@@ -1587,36 +1586,29 @@ public function unlinkDisownedObjects($sourceStage, $targetStage)
     /**
      * Removes the record from both live and stage
      *
+     * User code must call {@see canArchive()} prior to invoking this method.
```
flamerohr: Same comment above

src/Versioned.php
```diff
@@ -1631,15 +1623,13 @@ public function deleteFromChangeSets()
     /**
      * Removes this record from the live site
      *
+     * User code must call {@see canUnpublish()} prior to invoking this method.
```
flamerohr: Same comment as above

src/Versioned.php
```diff
@@ -1688,15 +1678,13 @@ public function hasPublishedOwners()
     /**
      * Revert the draft changes: replace the draft content with the content on live
      *
+     * User code must call {@see canRevertToLive()} prior to invoking this method.
```
flamerohr: Same comment as above

flamerohr:
Overall makes sense, I needed to refer to the big red box at the bottom of the link to understand why the `canPublish` checks were removed completely - as adding the explicit filter seemed like enough at first glance.

tractorcow:
Done @flamerohr

flamerohr:
will this require any admin/cms/asset pull requests to re-add the `canPublish` or `canArchive` calls/checks back in?

tractorcow:
Those should already be checking those permissions.

tractorcow:
@kinglozzer do you mind giving this a review? I'd appreciate a third opinion please. :)

PR #114, opened 2018-02-14
Update reference to TestAssetStore

tractorcow:
So we can use dist version of assets module for testing


robbieaverill:
The only annoyance is for modules that have been using this from 4.0 compatibility onwards, will now need to do a class_exists check for both of the FQCNs (comment related to the overall change, not this particular PR)

tractorcow:
Those modules can just use the old class. They just have to use --prefer-source when installing tests in travis

PR #118, opened 2018-02-19
BUG Fix doRollbackTo() writing old / unsaved version over restored version

tractorcow:
See warning on PHPDoc

Fixes https://github.com/tractorcow/silverstripe-fluent/issues/313

### DIFF
--- a/tests/php/VersionedTest.php
+++ b/tests/php/VersionedTest.php
@@ -307,6 +307,7 @@ class VersionedTest extends SapphireTest
 
     public function testRollbackTo()
     {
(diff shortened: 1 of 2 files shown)


tests/php/VersionedTest.php
```diff
@@ -318,6 +319,7 @@ public function testRollbackTo()
         $page1->copyVersionToStage(Versioned::DRAFT, Versioned::LIVE);
         $changedVersion = $page1->Version;
 
+        $page1->Content = 'should be discarded';
```
tractorcow: Without the fix, the page would be overwritten with this content and the restored version lost.

Issue #26, opened 2017-07-06
Show warning when unpublishing an object with owners


chillu:
## Overview

By default, objects will not cascade an unpublish to their owners (see https://github.com/silverstripe/silverstripe-versioned/issues/25). This is based on the assumption that most owner templates gracefully handle the disappearance of a previously published owned object. Sometimes the presence of this object is required for the template to function correctly though. 

Example: `NewCustomerPage has_one SignupFormContentBlock`. The same content block might be used in different scenarios (e.g. a footer), and an author might not be aware that it's also the main entry point for new customers. Without this form, the whole page is considered functionally broken.

## Acceptance Criteria

 * Show a warning to authors before unpublishing objects with owner objects. 
 * Show the warning both for the default behaviour (unpublish object only) and when a developer has opted into cascading unpublish to owner objects
 * The warning allows authors to clearly identify affected records (e.g. a natural language object type plus the title)
 * The warning allows authors to click through to the CMS view of the object if `CMSEditLink()` is implemented
 * When no owner objects are detected, the CMS author isn't bothered with an additional step during unpublish
 * The warning shows in the Pages UI, Assets UI and GridField detail forms (incl. ModelAdmin)

## Tasks

- [x] Implement unpublish warning in asset-admin for files
- [x] Implement unpublish warning with versioned gridfield
- [x] Update tests

## Out of Scope

 * Show warning before unpublishing objects with *owned* objects (see [allow recursive unpublish](https://github.com/silverstripe/silverstripe-versioned/issues/23))

## Pull Requests
- [x] https://github.com/silverstripe/silverstripe-admin/pull/444
- [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/739
- [x] https://github.com/silverstripe/silverstripe-versioned/pull/122

/cc @clarkepaul @pitchandtone @stevie-mayhew

pitchandtone:
Warning might be a bit harsh, is notification better?

sminnee:
If they proceed, their site will end up with broken links / image references — this seems warning-worthy, doesn't it?

However, not all unpublishes will have that effect — if it's a has_many relationship it would just end up with 1 fewer items in the list. So we might need to distinguish the different situations.

Does it need to be a special case for the assets that are linked by inclusion in a WYSIWYG editor?

pitchandtone:
Files are going to be the edge case that require a warning, mainly because of the wysiwyg.

[...]

PR #122, opened 2018-02-20
Add warning when unpublishing owned records

unclecheese:
Resolves https://github.com/silverstripe/silverstripe-versioned/issues/26

### DIFF
--- a/src/VersionedGridFieldItemRequest.php
+++ b/src/VersionedGridFieldItemRequest.php
@@ -259,7 +259,7 @@ class VersionedGridFieldItemRequest extends GridFieldDetailForm_ItemRequest
     /**
      * Getting buttons that are for versioned objects
      *
-     * @param DataObject|Versioned $record
+     * @param DataObject|Versioned|RecursivePublishable $record
      * @param FieldList $actions
      */
     protected function addVersionedButtons(DataObject $record, FieldList $actions)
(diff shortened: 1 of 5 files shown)


src/VersionedGridFieldItemRequest.php
```diff
@@ -277,6 +277,7 @@ protected function addVersionedButtons(DataObject $record, FieldList $actions)
         // Unpublish action
         $isPublished = $record->isPublished();
         if ($isPublished && $record->canUnpublish()) {
+            $owners = ($record->hasExtension(RecursivePublishable::class)) ? $record->findOwners()->count() : 0;
```
tractorcow: Checking the extension is redundant at this point. Anything we call addVersionedButtons() on will be either versioned, or at the very least have RecursivePublishable on it.

tractorcow: I suggest to pass in false to findOwners() to just get the immediate level. A recursive search is slower and unnecessary.

src/VersionedGridFieldItemRequest.php
```diff
@@ -296,6 +296,7 @@ protected function addVersionedButtons(DataObject $record, FieldList $actions)
                         'Remove this record from the published site'
                     ))
                     ->addExtraClass('btn-secondary')
+                    ->setAttribute('data-owners', $record->findOwners(false)->count())
```
tractorcow: Need to change this to live owners, not draft owners (I'll fix).

PR #129, opened 2018-03-01
FIX Duplicate only what we want

### LINKED ISSUE #127: Duplication check is too permissive and duplicates when it shouldn't.
https://github.com/silverstripe/silverstripe-versioned/blob/eba04d8b243afda7e060c7b91f9cee005e6b5c79/src/RecursivePublishable.php#L321

This line checks that we both want to duplicate relations, and that we haven't already set what to duplicate.
In legacy; if we _do_ want to duplicate a _specific type_ of relation, then ideally `$relations` will be an array with field names/keys.
But if e.g. we want to copy _only_ the `many_many` fields, and this particular class does not have any many many relationships defined, then this will be an [empty array](https://github.com/silverstripe/silverstripe-framework/blob/4.1/src/ORM/DataObject.php#L425).

An empty array is 'falseish', but is neither `false`, nor `trueish` as is checked for here.

**TL:DR;**
So we ask to copy only `many_many` fields, but end up having _every single relationship_ duplicated instead.

This is rather not ideal for backwards compatibility in that a class that specifies this recently deprecated functionality will also likely [manage it's own duplication methods for such relationships](https://github.com/silverstripe/silverstripe-userforms/blob/5.0.0/code/Model/EditableFormField/EditableMultipleOptionField.php#L123).

NightJar:
In SilverStripe core 4.1 line, when we duplicate a record we have the
option of trying to duplicate only many_many, or not. 4.1 extends this
logic to specify any of the relationship types (in a legacy support mode
that is marked deprecated to be fair). This way any old module
that sepcifies 'many_many' for the relations parameter (as was the old
default) in theory keeps working.

The issue is that not all models define a `many_many` relationship - and
when this happens the resulting list of reltationship names to duplicate
is an empty array. This would be fine, but then Versioned here goes and
writes into that reference passed array, every single relationship defined
on the model. So the end result is actually that any old module that
manages it's own relations on duplication ends up with two of everything
in said defined relationship.

This is easily recitfied by making the comparison more strict, rather than
either 'trueish' or specificlly `false`, we'll short-circuit on anything
that isn't `null`.

Resolves #127

### DIFF
--- a/src/RecursivePublishable.php
+++ b/src/RecursivePublishable.php
@@ -318,7 +318,7 @@ class RecursivePublishable extends DataExtension
     public function onBeforeDuplicate($original, &$doWrite, &$relations)
     {
         // If relations to duplicate are declared (or forced off) don't rewrite
-        if ($relations || $relations === false) {
+        if ($relations !== null) {
             return;
         }
 


robbieaverill:
Grave digging a little, I'm curious about whether [this patch](https://github.com/silverstripe/silverstripe-framework/pull/8260) is fixing the same problem you're talking about?

NightJar:
No I believe they're different. The patch at https://github.com/silverstripe/silverstripe-framework/pull/8260 will prevent duplicate keys in the array - i.e. every relationship is duplicated only once.

This patch is a little different. It will prevent every relationship from being duplicated.

The intention is that only relationships defined on this particular model are duplicated (not including any defined on super classes, for example).

The [framework patch](https://github.com/silverstripe/silverstripe-framework/pull/8260) means that the array can still contain super class relationship definitions - it will just never contain them _twice_.

NightJar:
Versioning is a complex topic with a lot of brainpower going into it at the moment, how does everyone feel about this patch?

maxime-rainville:
Could you add some unit tests that would have broken before and now work with the fix. It would help figure the expected behaviour here.

maxime-rainville:
Any luck with this?

NightJar:
Nope not yet :>

Issue #93, opened 2018-01-30
Archive mode should include all versioned records


chillu:
# Overview

Content authors, owners and auditors want to have certainty that changes to the website are recorded correctly, both for auditing purposes and to restore content to an earlier state. The main way this detail is exposed to users is through the "History" view. Users can select an older version and preview a page at this past state. In SilverStripe 4, we're relying on versioned objects and nested relationships more strongly. These should be consistently included in previews of older versions.

# Acceptance Criteria

 * The archive mode includes nested relationships for the website preview panel
 * The archive mode includes nested relationships for readonly CMS form showing the record (incl. related records shown there)
 * Works with nested versioned relationships (e.g. Page has_many GalleryContentBlock has_many GalleryItem has_one Image)
 * Works regardless of ownership on relationship
 * Works with has_one and belongs_to
 * Works with has_many
 * Works with [many_many through](https://docs.silverstripe.org/en/4/developer_guides/model/relations/#many-many-through-relationship-joined-on-a-separate-dataobject) where all three models are versioned
 * Respects the given stage of the versioned record. I.e. viewing an unpublished record will show the stage version of all nested records, but viewing the published record will show the live website at that point in time.
 * Respects records which have been unpublished (but not deleted), either by showing them on draft, or not showing them on live.
 * Respects sort order on has_many and many_many
 * Hides records which have been created after the previewed version
 * Hides records which have been created before the previewed version, but have been associated through a versioned relationship after the previewed version
 * Shows records which have been deleted after the previewed version
 * Does not rely on specific version associations in versioned relationships (uses archive date)

# Scenarios

## Delete owned with republishing owner

 * Create Page1
 * Publish Page1v1
 * Create GalleryBlock1, add to Page1
 * Publish Page1v2 (publishes GalleryBlock1v1)
 * Delete GalleryBlock1 ("archive")
 * Publish Page1v3

Expected: Viewing Page1v3 should not show GalleryBlock1

## Delete owned without republishing owner

 * Create Page1
 * Publish Page1v1
 * Create GalleryBlock1, add to Page1
 * Publish Page1v2 (publishes GalleryBlock1v1)
 * Delete GalleryBlock1 ("archive")

Expected: Viewing Page1v2 should show GalleryBlock1

## Delete nested owned without republishing owner

 * Create Page1
 * Publish Page1v1
 * Create GalleryBlock1, add to Page1
 * Create GalleryBlockItem1, add to GalleryBlock1
[...]

PR #131, opened 2018-03-06
WIP: Include all versions in archive mode

unclecheese:
## WIP -- DON'T MERGE ##

Resolves https://github.com/silverstripe/silverstripe-versioned/issues/93

### DIFF
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -1564,8 +1564,7 @@ SQL
             $original = null;
         }
         $owner->invokeWithExtensions('onBeforePublish', $original);
-        $owner->write();
-        $owner->copyVersionToStage(static::DRAFT, static::LIVE);
+        $owner->writeToStage(static::LIVE);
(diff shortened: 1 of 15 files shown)


tractorcow:
Looking at this.

src/Versioned.php
```diff
@@ -299,16 +302,22 @@ protected function getLastEditedForVersion($version)
...
+            $list = [
+                $result['LastEdited'],
+                $result['WasPublished'] ? static::LIVE : static::DRAFT,
+            ];
+            $this->versionModifiedCache[$key] = $list;
+
+            return $list;
         }
-        return $date;
+        return $result;
```
tractorcow: I wonder if this is an error not simply a return $result.

tractorcow:
Overall looking good. :)

tractorcow:
Pushed a fix that might fix postgres.

tractorcow:
Behat tests are all green ,but cms phpunit tests are broken. I'll fix soon.

PR #132, opened 2018-03-08
NEW Add isLiveVersion and isLatestDraftVersion to Versioned and GraphQL DataObject scaffolding

robbieaverill:
This surfaces whether each version is the current live version and the current/latest draft or modified version.

Required for #37

### DIFF
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -2445,6 +2445,38 @@ SQL
     }
 
     /**
+     * Returns whether the current record's version is the current live/published version
+     *
+     * @return bool
+     */
+    public function isLiveVersion()
+    {
+        $id = $this->owner->ID ?: $this->owner->OldID;
+        if (!$id || !$this->isPublished()) {
+            return false;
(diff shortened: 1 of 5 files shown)


src/Versioned.php
```diff
@@ -2379,7 +2379,7 @@ public function isLiveVersion()
         }
 
         $liveVersionNumber = static::get_versionnumber_by_stage($this->owner, Versioned::LIVE, $id);
-        return $liveVersionNumber === $this->owner->Version;
+        return (int) $liveVersionNumber === (int) $this->owner->Version;
```
dhensby: I'm not sure of the benefit of strict type checking with casting? we might as well do a loose check?

robbieaverill: Updated, cheers

robbieaverill:
Rebased to get tests passing

robbieaverill:
Self merging with three approvals plus @dhensby's comments addressed

Issue #75, opened 2017-12-14
Inconsistency: Publishing files linked in HTMLEditorField 


bummzack:
# Acceptance Criteria

 * Files and images are tracked in each `HTMLText` field regardless of record type
 * As a dev upgrading from SS3 or starting over with SS4, I don't have to do do anything special to get this behaviour
 * As a dev ugprading to SS4 (from 3.x or <=4.1), I retain the existing tracking data
 * As a dev, I can opt into running a migration task that backfills this data for non-page records
 * As an author, this change does not impact my saving and publishing behaviour (performance)
 * As an author, I can see new link tracking info in the "file usage" tab on AssetAdmin

# Notes

 * Excludes "HasBrokenLink" / "HasBrokenFile" on global dataobjects (retained only for SiteTree)
 * Both the dev/build and migration tasks work on large databases (50k records)
 * Excludes versioned tracking
 * Not doing ORM refactor for polymorphic many_many (using existing API)
 * Track under ORB-175

# Original Description

Issue found with silverstripe-versioned `1.0.1`

When a file is inserted/linked inside the `HTMLEditorField` it's not getting published if the parent is a `DataObject`.

Consider these three cases:

 -  `SiteTree`: upon publish, files linked in `HTMLEditorField` get published as well
 - `DataObject` (non-versioned): upon saving, files linked in `HTMLEditorField` are still unpublished
 - `DataObject` (versioned): upon publishing, files linked in `HTMLEditorField` are still unpublished

I guess not publishing files in unversioned `DataObjects` is reasonable, still, it would be nice to have a config flag for this (or a way to publish the file from the "insert file" dialogue).

Publishing a file that is in an editor-field of a _versioned_ `DataObject` should behave the same as `SiteTree` though.

# PRs:

 - [x] https://github.com/silverstripe/silverstripe-framework/pull/7988
 - [x] https://github.com/silverstripe/silverstripe-framework/pull/7928
 - [x] https://github.com/silverstripe/silverstripe-cms/pull/2123
 - [x] https://github.com/silverstripe/silverstripe-assets/pull/122
 - [x] https://github.com/silverstripe/silverstripe-versioned/pull/137
 - [x] https://github.com/silverstripe/silverstripe-framework/pull/7960
 - [x] https://github.com/silverstripe/silverstripe-asset-admin/pull/754

Dependent issue:

- [x] https://github.com/silverstripe/silverstripe-framework/issues/7962
[...]

PR #137, opened 2018-03-27
BUG Don't crash on polymorphic ownership

tractorcow:
Parent story https://github.com/silverstripe/silverstripe-versioned/issues/75


tractorcow:
Damn there was a bug in this. Should have used array_unshift not pop...;_;

tractorcow:
See https://github.com/silverstripe/silverstripe-framework/issues/7962 for follow up patch

PR #138, opened 2018-03-28
BUG Ensure polymorphic ownership works

tractorcow:
Fix and tests for https://github.com/silverstripe/silverstripe-framework/issues/7962.

Merge framework first.


tractorcow:
I got it confused when I wrote it lol.

dhensby:
maybe `subclassesFor` should just return subclasses :P

PR #140, opened 2018-04-04
ENHANCEMENT / BUG Improve cleanupVersionedOrphans

### LINKED ISSUE #139: BUG cleanupVersionedOrphans fails if table case needs fixing
## Affected Version

4.x (all)

## Description

When migrating from a database with a `SiteTree_versions` table, the ORM will rename this to `SiteTree_Versions`. however, Versioned::cleanupVersionedOrphans() will attempt to run queries on this table, prior to the rename, but it does this using the incorrect case.

On case-sensitive db / filesystems, this will result in a database exception prior to the table getting renamed.

## Steps to Reproduce

Install 3.x and dev/build
Update to 4.x and dev/build

## Possible solutions

 - Run cleanupVersionedOrphans() after table rename
 - Make cleanupVersionedOrphans act in a case-sensitive fashion (raw queries use real case of table, not desired case).
 
# PRs:

 - [x] https://github.com/silverstripe/silverstripe-versioned/pull/140

tractorcow:
ENHANCEMENT Allow cleanupVersionedOrphans to be disabled
BUG Fix cleanupVersionedOrphans breaking on table case change
BUG Use single query for deleting orphans for better performance on huge datasets

Fixes #139
Fixes https://github.com/silverstripe/silverstripe-framework/issues/7944

For peer review by @mateusz please who raised the original issues.

With a single SQL DELETE we should now be able to cleanup much larger datasets without error, although the config is now available if you wish to disable it anyway.

### DIFF
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -21,6 +21,7 @@ use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\DataQuery;
 use SilverStripe\ORM\DB;
 use SilverStripe\ORM\FieldType\DBDatetime;
+use SilverStripe\ORM\Queries\SQLDelete;
 use SilverStripe\ORM\Queries\SQLSelect;
 use SilverStripe\Security\Member;
 use SilverStripe\Security\Permission;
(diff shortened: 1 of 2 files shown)


unclecheese:
@tractorcow can you check the Postgres tests?

tractorcow:
Postgres needs a fix for `delete bob from bob, othertable`.

See https://stackoverflow.com/questions/11753904/postgresql-delete-with-inner-join

tractorcow:
Ok, I had to revert a bunch of the optimisations because they didn't work in postgres. Can't do delete with joins in postgres, and can't do delete with subquery in mysql... ugh.

Issue #124, opened 2018-02-27
SS5: WasPublished seems broken


dhensby:
silverstripe/cms:dev-master tests are failing because of an assumption about CSS classes that will be present in the history controller (specifically the `VersionsForm`).

It appears that all records, whether saved as draft or published come back with `WasPublished` being truthy and this means the wrong CSS classes are being applied and also that the `Show unpublished versions` checkbox does nothing.

# PRs:

 - [x] https://github.com/silverstripe/silverstripe-versioned/pull/141

tractorcow:
It's not just master.

tractorcow:
Ah no it is... separate issue .;)

tractorcow:
Just needs a merge up.

Also, graphql 2.x has been retconned for framework 4.2 and above. framework 5 should now be using graphql 3.x.

dhensby:
versioned seems to be all merged up...

dhensby:
One of the CMS tests is passing that was failing, but still 2 failing: https://travis-ci.org/silverstripe/silverstripe-cms/builds/356924038

tractorcow:
I haven't finished merging up. Only 1/2 way through the modules.

tractorcow:
All merged up. :D

tractorcow:
waiting for https://travis-ci.org/silverstripe/silverstripe-cms/builds/362334146 to build.

tractorcow:
Well it's closer! Only one bug now, not a huge list of them.

tractorcow:
```
1) SilverStripe\CMS\Tests\Controllers\CMSPageHistoryControllerTest::testVersionsFormTableContainsInformation
Failed asserting that 'even published' contains "internal".
```

Probably a mix of Versioned_Version being deprecated in 5.x and the WasDraft / WasDeleted change in 4.x not working together. Bugfix in 5.x IMO as a separate PR.

tractorcow:
Additionally, fails with MySQLPDODatabase, but passes with MySQLDatabase

tractorcow:
because DBString::exists() treats "0" string for "WasPublished" as true, when it's a boolean.

Fix: Cast WasPublished as boolean

tractorcow:
```


    private static $casting = [
        "RecordID" => "Int",
        "Version" => "Int",
        "WasPublished" => "Boolean",
        "WasDeleted" => "Boolean",
        "WasDraft" => "Boolean",
        "AuthorID" => "Int",
[...]

PR #141, opened 2018-04-04
BUG Ensure extra fields have correct casting

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-versioned/issues/124

Once merged, merge up to master to fix cms master tests.

The problem is that the `Was*` fields were cast to default_cast (varchar), which treats "0"(string) as non-falsey. PDO connections use strings for int results, hence the tests would pass on non-pdo connectors.

Casting to boolean correctly ensures the correct cast.

### DIFF
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -168,6 +168,21 @@ class Versioned extends DataExtension implements TemplateGlobalProvider, Resetta
     ];
 
     /**
+     * Ensure versioned records cast extra fields properly
+     *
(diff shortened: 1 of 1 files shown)


dhensby:
Interesting, so why isn't `$db_for_versions_table` enough?

tractorcow:
> $db_for_versions_table

Because `ViewableData::castingHelper` ignores `db_for_versions_table`

PR #142, opened 2018-04-10
ENHANCEMENT Ensure invalid stage values are thrown as exceptions

tractorcow:
Reference; https://github.com/dnadesign/silverstripe-elemental/pull/215

This change will prevent invalid stages creeping into the current mode, causing hard-to-track down errors.


tractorcow:
We did throw exceptions, we just threw them later on, rather than earlier. All of the exceptional states should be caught up front.

PR #148, opened 2018-04-23
[WIP] ENHANCEMENT Ensure that default caches are segmented based on versioned state

tractorcow:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/8020

This is in a bit of a rough state; No tests, and there ARE some core caches that should NOT have this mode enabled (e.g. VersionProvider_composerlock, RateLimiter, InheritedPermissions, essentially all default caches).

I've roughed up an opt-out arg that can be set in yml, but it's not tested well with these core caches.

For review @sminnee  / @chillu.


src/Caching/ProxyCacheAdapter.php
```diff
@@ -0,0 +1,195 @@
...
+        }
+    }
+
+    /**
+     * Map user cache to internal cache
+     *
+     * @param string $key
+     * @return string
+     */
+    abstract protected function getKeyID($key);
```
tractorcow: So the idea here is that you can map a key to an internal ID that's actually used to store the value. You can segment based on stage, subsite, etc... but this implementation is hard-fixed to map based on versioned reading mode.

src/Caching/ProxyCacheAdapter.php
```diff
@@ -0,0 +1,195 @@
...
+ * psr-6 cache proxy for an internal cache, which provides segmentation of
+ * cache keys based on current versioned mode. This ensures that cross-stage
+ * content cannot cross-pollenate each other.
+ *
+ * Note: segmentation can be disabled via 'versionedstate = false' being supplied as a
+ * constructor arg.
+ *
+ * Based on Symfony\Component\Cache\Simple\TraceableCache
+ */
+abstract class ProxyCacheAdapter implements CacheInterface, ResettableInterface, PruneableInterface
```
chillu: Right, so the "adapter" naming is consistent with other symfony cache implementations (https://sourcemaking.com/design_patterns/adapter). I think the specific implementation here is closer to a "decorator" (https://sourcemaking.com/design_patterns/decorator) than a "proxy" (https://sourcemaking.com/design_patterns/proxy), since the class here isn't taking responsibility for constructing the inner object - it gets it via `__construct()`.

tests/php/Caching/VersionedCacheTest.php
```diff
@@ -0,0 +1,45 @@
...
+                'constructor' => [
+                    'namespace' => 'myapp',
+                ],
+            ],
+        ]);
+        /* @var CacheInterface $cacheInstance */
+        $cacheInstance = Injector::inst()->get($cacheService);
+        $cacheInstance->clear();
+
+        Versioned::set_stage(Versioned::LIVE);
```
chillu: @unclecheese Please update those tests to be consistent with the changes in the 3.x tests I requested

unclecheese: Sorry, I had done, but forgot to push up.

tractorcow:
Looks good so far @unclecheese

PR #149, opened 2018-04-26
Added missing HTML casting on the "permission failure" error message.

mfendeksilverstripe:
"permission failure" message contains HTML which currently displays as a plaintext. This changeset adds casting to the message so it can display properly.


dhensby:
does this affect 1.0?

mfendeksilverstripe:
@dhensby I think it effects 1.0 as well as the code in the related file is lacking the casting.

dhensby:
OK, I've rebased this onto 1.0

PR #151, opened 2018-05-07
Use record ID to detect new item for button label

lukereative:
Fixes https://github.com/silverstripe/silverstripe-framework/issues/8029

### DIFF
--- a/src/VersionedGridFieldItemRequest.php
+++ b/src/VersionedGridFieldItemRequest.php
@@ -343,8 +343,7 @@ class VersionedGridFieldItemRequest extends GridFieldDetailForm_ItemRequest
         if (!$saveAction) {
             return;
         }
-        $ownsObjects = !empty($record->config()->get('owns'));
-        if (!$ownsObjects) {
+        if (!$this->record->ID) {
             return;
         }
 
--- a/tests/php/VersionedGridFieldItemRequestTest.php
+++ b/tests/php/VersionedGridFieldItemRequestTest.php
@@ -47,10 +47,8 @@ class VersionedGridFieldItemRequestTest extends SapphireTest
         $this->assertNull($actions->fieldByName('action_doPublish'));
         $this->assertInstanceOf(FormAction::class, $actions->fieldByName('action_doSave'));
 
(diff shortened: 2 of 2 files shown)


robbieaverill:
I'm unsure why this has to be changed in PHP - the timeline as I see it was:

* Introduce a new UI feature
* Forget to update the behat tests

Wouldn't this change undo the new UI feature of "Apply changes" rather than "Save"? Or is there a bug in that that this is fixing?

lukereative:
It's actually a bug where the button is erroneously is "Apply Changes" rather than "Create" when creating a new DataObject. So not sure why this wasn't picked up originally since the Behat tests were correctly reporting an issue.

robbieaverill:
Ok great! Thanks :-)

src/VersionedGridFieldItemRequest.php
```diff
@@ -343,8 +343,7 @@ protected function addUnversionedButtons(DataObject $record, FieldList $actions)
         if (!$saveAction) {
             return;
         }
-        $ownsObjects = !empty($record->config()->get('owns'));
-        if (!$ownsObjects) {
+        if (!$this->record->ID) {
```
robbieaverill: Judging by @tractorcow's comment at https://github.com/silverstripe/silverstripe-framework/issues/8029#issuecomment-384473794, and the test failures at https://travis-ci.org/silverstripe/silverstripe-versioned/jobs/376140206#L923 I suspect that we might need to have this condition _as well_ as what was there originally, maybe?

unclecheese: I don't think so. If the record is new, the button should always read "create," regardless of ownership structure.

tractorcow: Yes but if the record is NOT new but doesn't own anything you also want to return.

unclecheese: Yup, that's been addressed in a separate PR. https://github.com/silverstripe/silverstripe-versioned/pull/154

For now, @chillu has advised we simply patch the behat tests and get this running, because it's holding up installer.

PR #160, opened 2018-06-14
Update recipe branching

tractorcow:
Parent issue: https://github.com/silverstripe/recipe-core/issues/24

From 4.2 onwards all recipes will use the same numbering as the core framework version.


chillu:
Base branch fails the same way in tests

PR #165, opened 2018-06-27
Install 5.x-dev recipe when testing master.

maxime-rainville:
Parent issue silverstripeltd/open-sourcerers#31


robbieaverill:
The linting error has been fixed in the 1 branch already which I've just merged up to master, so this would be green now if restarted. Merging anyway for expediency.

PR #167, opened 2018-07-10
Add versioned graphql types via extension instead of direct YAML

unclecheese:
In adherence to new multi-schema work.

Resolves: https://github.com/silverstripe/silverstripe-graphql/issues/158

### DIFF
--- a/src/GraphQL/Extensions/ManagerExtension.php
+++ b/src/GraphQL/Extensions/ManagerExtension.php
@@ -1,0 +1,31 @@
+<?php
+
+namespace SilverStripe\Versioned\GraphQL\Extensions;
+
+use SilverStripe\Core\Extension;
+use SilverStripe\Versioned\GraphQL\Types\CopyToStageInputType;
+use SilverStripe\Versioned\GraphQL\Types\VersionedInputType;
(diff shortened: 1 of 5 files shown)


src/GraphQL/Extensions/ManagerExtension.php
```diff
@@ -0,0 +1,31 @@
...
+use SilverStripe\Versioned\GraphQL\Types\VersionedStatus;
+
+class ManagerExtension extends Extension
+{
+    /**
+     * Adds the versioned types to all schemas
+     *
+     * @param $config
+     */
+    public function updateConfig(&$config)
```
maxime-rainville: Any way we could had a bit of logic to test those extra types are added correctly?

maxime-rainville: Good enough. I withdraw my pedantic objection.

maxime-rainville:
We need silverstripe/silverstripe-graphql#173 and silverstripe/recipe-cms#18 to be merged first.

Then we can re-trigger the build.

robbieaverill:
Dependencies merged and builds restarted

PR #168, opened 2018-07-11
NEW Archive admin for managing archived records

lukereative:
**Implements:** https://github.com/silverstripe/silverstripe-framework/issues/6202

**Sibling PRs:**
- https://github.com/silverstripe/silverstripe-framework/pull/8240
- https://github.com/silverstripe/silverstripe-versioned-admin/pull/29
- https://github.com/silverstripe/silverstripe-admin/pull/559


src/RestoreAction.php
```diff
@@ -0,0 +1,162 @@
...
+     * @param Object $item
+     * @return array $message
+     */
+    public static function restore($item)
+    {
+        $isArchived = $item->isArchived();
+        $canRestoreToDraft = $item->canRestoreToDraft();
+
+        if (!$canRestoreToDraft) {
+            throw new ValidationException(
```
maxime-rainville: Should have a 403 code.

src/RestoreAction.php
```diff
@@ -0,0 +1,162 @@
...
+                _t(
+                    __CLASS__ . '.RESTORE_FALIURE_PERMISSION',
+                    'Insufficient permission to restore item'
+                )
+            );
+        }
+
+        if (!$isArchived) {
+            throw new ValidationException(
```
maxime-rainville: Should return a 409 HTTP code.

src/Versioned.php
```diff
@@ -1438,6 +1438,44 @@ public function canRevertToLive($member = null)
...
+    {
+        $owner = $this->owner;
+
+        // Skip if invoked by extendedCan()
+        if (func_num_args() > 4) {
+            return null;
+        }
+
+        // Can't revert if not archived
+        if (!$owner->isArchived()) {
```
maxime-rainville: I don't think we should have this check here as it's not based on the Member's permission.

Basically, `canRestoreToDraft()` should return true if the provided Member has the right permission to restore the record, regardless of whatever the record is in a *restorable* state.

src/Versioned.php
```diff
@@ -1438,6 +1438,44 @@ public function canRevertToLive($member = null)
         return $owner->canEdit($member);
     }
 
+    /**
+     * Check if the user can restore this record to draft
+     *
+     * @param Member $member
+     * @return bool
+     */
+    public function canRestoreToDraft($member = null)
```
maxime-rainville: Probably wouldn't hurt to have a dedicated unit test for this one.

lukereative: Yep, done

src/Versioned.php
lukereative: Makes sense, done

src/RestoreAction.php
lukereative: Done

src/RestoreAction.php
lukereative: Done

src/Versioned.php
maxime-rainville: Did you forget to push a commit? The archive check is still there.

PR #169, opened 2018-07-13
Fix escaping for back link

wilr:
Currently the back link is not correctly escaped as HTML Text.

Example: 
screen shot 2018-07-13 at 2 17 04 pm


wilr:
@robbieaverill changed base to 1.1, updated php lint check

robbieaverill:
Oh, I just merged #149 which does the same thing. It'll get merged up to this line too. Thanks for the PR though @wilr =)

PR #173, opened 2018-08-27
FIX Do not update LeftAndMain link with Stage param

maxime-rainville:
Add a condition so we don't try to add Stage URL param to LeftAndMain links.

Fixes https://github.com/silverstripe/silverstripe-admin/issues/607

### DIFF
--- a/src/VersionedStateExtension.php
+++ b/src/VersionedStateExtension.php
@@ -2,6 +2,7 @@
 
 namespace SilverStripe\Versioned;
 
+use SilverStripe\Admin\LeftAndMain;
 use SilverStripe\Control\Controller;
 use SilverStripe\Control\RequestHandler;
 use SilverStripe\Core\Extension;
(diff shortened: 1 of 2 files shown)


dhensby:
This is a bit gross as we're mixing knowledge about modules in a way we shouldn't.

It would be much nicer to have a way for `LeftAndMain` to disable this link re-writing; can't we disable/remove the extension for the LeftAndMain controller?

lukereative:
Any further suggestions @robbieaverill on this regarding @dhensby's comment and my reply?

robbieaverill:
@lukereative yeah I've been thinking about this. I thought about maybe moving the extension point from RequestHandler to something higher in the chain, but then we risk breaking people's code that extends low level classes like this and expects versioned capability.

I personally don't mind the coupling, particularly when there's no great alternative. @dhensby do you have any ideas?

lukereative:
@chillu has asked this be merged as it's `impact/critical` and since the only pushback is semantic (which doesn't seem to be that avoidable either) rather than anything regarding the function of the fix – we can fix this up later if a better way presents itself.

ScopeyNZ:
Can we get this merged up to `1` please?

PR #174, opened 2018-08-28
FIX Detection of extendedCan now works preventing infinite loops in unit tests

robbieaverill:
GridFieldRestoreActionTest::testDontShowRestoreButtons via TestObject::canRestoreToDraft was causing infinite loops and exceeding memory limits. The Versioned::canRestoreToDraft method calls itself via $owner->extendedCan(canRestoreToDraft, $member), but it never passes more than two arguments. This patch fixes the loose recursion check.

This should get the SS5 tests green.

### DIFF
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -1442,7 +1442,7 @@ SQL
         $owner = $this->owner;
 
         // Skip if invoked by extendedCan()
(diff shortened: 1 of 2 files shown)


robbieaverill:
Self merging with Maxime's approval

PR #176, opened 2018-09-04
Compliance with new graphql resolvedTypeName() method

unclecheese:
Partial resolution of https://github.com/silverstripe/silverstripe-graphql/pull/176
Depends on https://github.com/silverstripe/silverstripe-graphql/pull/182

### DIFF
--- a/src/GraphQL/Operations/CopyToStage.php
+++ b/src/GraphQL/Operations/CopyToStage.php
@@ -40,7 +40,7 @@ class CopyToStage extends MutationScaffolder implements OperationResolver
             return $name;
         }
 
-        $typeName = $this->getTypeName();
+        $typeName = $this->getResolvedTypeName();
 
         return 'copy'.ucfirst($typeName).'ToStage';
     }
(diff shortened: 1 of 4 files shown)


unclecheese:
Yeah good point. Not the first time this has come up, either. We have so many modules that aren't necessarily direct dependencies of each other, but they are by proxy. In this case `silverstripe/admin` is the glue between the two, which is likely installed, but not required. When it is, it would enforce that constraint.

I'm not sure about doing it procedurally. Seems like any solution would be pretty ugly. The `Module` class doesn't offer a `getVersion` API or anything. 

Any thoughts on how we might do it? I'm open to ideas.

maxime-rainville:
What about adding a `conflict` in on the packages? https://getcomposer.org/doc/04-schema.md#conflict

maxime-rainville:
Ping @unclecheese :cat:

emteknetnz:
@unclecheese  I'm currently going through some old pull-requests to see what is still active and what makes sense to be closed - is this one something you're still planning on keeping active or would you prefer to now close this one?

Cheddam:
@unclecheese Closing this out for now - if it makes sense to reapproach later feel free to reopen.

PR #181, opened 2018-09-17
FIX: Performance optimisation for draft pages in treeview

sminnee:
When draft pages are rendered in the treeview, an attempt to look up
their live Version # is made. This leads to a cache miss and a
subsequent query for each record.

This change marks a cache as “complete” when all records of a given
stage are pre-cached. This will suppress the record-specific query
in the case of a cache miss.

# Parent issue
* silverstripe/silverstripe-cms#2250

### DIFF
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -2282,6 +2282,13 @@ SQL
             return self::$cache_versionnumber[$baseClass][$stage][$id] ?: null;
         }
 
+        // if the cache was marked as "complete" then we know the record is missing, just return null
+        // this is used for treeview optimisation to avoid unnecessary re-requests for draft pages
+        if (!empty(self::$cache_versionnumber[$baseClass][$stage]['_complete'])) {
+            return null;
+        }
+
+
         // get version as performance-optimized SQL query (gets called for each record in the sitetree)
         $version = DB::prepared_query(
             "SELECT \"Version\" FROM \"$stageTable\" WHERE \"ID\" = ?",
@@ -2320,6 +2327,13 @@ SQL
         if (!Config::inst()->get(static::class, 'prepopulate_versionnumber_cache')) {
             return;
         }
+
+        /** @var Versioned|DataObject $singleton */
+        $singleton = DataObject::singleton($class);
+        $baseClass = $singleton->baseClass();
+        $baseTable = $singleton->baseTable();
+        $stageTable = $singleton->stageTable($baseTable, $stage);
+
         $filter = "";
         $parameters = [];
         if ($idList) {
@@ -2334,13 +2348,12 @@ SQL
             }
             $filter = 'WHERE "ID" IN (' . DB::placeholders($idList) . ')';
             $parameters = $idList;
-        }
 
-        /** @var Versioned|DataObject $singleton */
-        $singleton = DataObject::singleton($class);
-        $baseClass = $singleton->baseClass();
-        $baseTable = $singleton->baseTable();
-        $stageTable = $singleton->stageTable($baseTable, $stage);
+        // If we are caching IDs for _all_ records then we can mark this cache as "complete" and in the case of a cache-miss
+        // no subsequent call is necessary
+        } else {
+            self::$cache_versionnumber[$baseClass][$stage] = [ '_complete' => true ];
+        }
 
         $versions = DB::prepared_query("SELECT \"ID\", \"Version\" FROM \"$stageTable\" $filter", $parameters)->map();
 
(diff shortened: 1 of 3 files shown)


chillu:
Note that unless this fix is considered "critical" and rebased against the `1.2` branch, it won't be available for the targeted project (since they're trying to avoid a 4.3 recipe upgrade). I'd say it's non-critical, and has a moderate chance of functional regressions, so it's targeted correctly with `1`. But that won't help the targeted project which it has been written for.

maxime-rainville:
I tested on our key project. It doesn't look like it makes much of a difference. I've compared building the treeview for the `North Island` page in black fire.

This is comparing the original version to the patched version. 

image

We save 245 queries ( out of 2 381 ) and improve performance time by about 2.2%.

@chillu I can review this in greater details if you want, but it's not going to resolve our treeview issues for our key project.

chillu:
Assuming this is the PR which reduces number of queries, it'll have a far bigger impact in SSP with a remote database than on your local. No need to set up Blackfire on SSP for this, but I think it's still worthwhile to get peer reviewed and merged this week.

sminnee:
Note that this requires unpublished draft page to have much of an effect.

tests/php/VersionedTest.php
```diff
@@ -1491,4 +1491,45 @@ public function testStageStates()
...
+        $this->assertNull(
+            Versioned::get_versionnumber_by_stage(VersionedTest\TestObject::class, Versioned::DRAFT, 999, true),
+            'get_versionnumber_by_stage should return null when fetching a version number for cached unexisting object'
+        );
+
+        Versioned::prepopulate_versionnumber_cache(VersionedTest\TestObject::class, Versioned::DRAFT);
+        $this->assertNull(
+            Versioned::get_versionnumber_by_stage(VersionedTest\TestObject::class, Versioned::DRAFT, 888, true),
+            'get_versionnumber_by_stage should return null when fetching a version number for unexisting object with prepopulate_versionnumber_caching'
+        );
```
robbieaverill: This test is actually six or seven different tests, or different scenarios for one test. Can you please split it up or use a data provider?

chillu: Given Robbie is only available in EU timezone, I'll say: Yep, that sounds good ;)

maxime-rainville: I've switched to using a data provider and use a more thorough parameter coverage. I've moved the logic to a separate file because I need to have some `setUp` and `setUpBeforeClass` customisation.

The new tests actually found a bug, so I fix that as well.

src/Versioned.php
```diff
@@ -2278,17 +2278,16 @@ public static function get_versionnumber_by_stage($class, $stage, $id, $cache =
...
-        }
-
-        // if the cache was marked as "complete" then we know the record is missing, just return null
-        // this is used for treeview optimisation to avoid unnecessary re-requests for draft pages
-        if (!empty(self::$cache_versionnumber[$baseClass][$stage]['_complete'])) {
-            return null;
+        if ($cache) {
+            if (isset(self::$cache_versionnumber[$baseClass][$stage][$id])) {
+                return self::$cache_versionnumber[$baseClass][$stage][$id] ?: null;
+            } elseif (isset(self::$cache_versionnumber[$baseClass][$stage]['_complete'])) {
```
ScopeyNZ: This can technically be just an `if` as you're returning in the preceding `if` statement. Not a blocker though - super super minor.

maxime-rainville: I think, I'll keep it as-is just to make it clear the second bit will only run if the first if statement is false.

sminnee:
OK this looks like we've addressed Robbie's feedback. Since I wrote this does @chillu want to merge?

tests/php/VersionedNumberCacheTest.php
```diff
@@ -0,0 +1,107 @@
...
+
+    public function setUp()
+    {
+        parent::setUp();
+        TestObject::singleton()->flushCache();
+    }
+
+    public function cacheDataProvider()
+    {
+        return [
```
chillu: As a sidenote, I find that data providers should only be used if the data within them is pretty self explanatory. If you find yourself going back and forth between the provider and the implementation (like I'm doing in this case), it's not helping readability. And in unit tests, I think readability overrules DRY principles.

tests/php/VersionedTest.php
chillu: Heh its funny, I've just commented on how unreadable the data provider refactor made this code ;) But I'd say it's good enough to merge

PR #183, opened 2018-09-21
FIX: Use Hierarchy::prepopulateTreeDataCache() in CMS.

sminnee:
Doesn't do anything without https://github.com/silverstripe/silverstripe-framework/pull/8395

### DIFF
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -2312,6 +2312,20 @@ SQL
     }
 
     /**
+     * Hook into {@link Hierarchy::prepopulateTreeDataCache}.
+     *
+     * @param DataList $recordList The list of records to prepopulate caches for. Null for all records.
+     * @param array $options A map of hints about what should be cached. "numChildrenMethod" and
+     *                       "childrenMethod" are allowed keys.
+     */
+    public function onPrepopulateTreeDataCache(DataList $recordList = null, array $options = [])
+    {
+        $idList = $recordList ? $recordList->column('ID') : null;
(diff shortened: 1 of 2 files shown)


src/Versioned.php
```diff
@@ -2311,6 +2311,20 @@ public static function get_versionnumber_by_stage($class, $stage, $id, $cache =
...
     }
 
+    /**
+     * Hook into {@link Hierarchy::prepopulateTreeDataCache}.
+     *
+     * @param DataList $recordList The list of records to prepopulate caches for. Null for all records.
+     * @param array $options A map of hints about what should be cached. "numChildrenMethod" and
+     *                       "childrenMethod" are allowed keys.
+     */
+    public function onPrepopulateTreeDataCache(DataList $recordList = null, array $options = [])
```
maxime-rainville: I updated `Hierachy::prepopulateTreeDataCache` to accept an array of IDs as well as DataList. My rational at the time was that if I'm getting a list of IDs as parameters on a request it sounds counter-intuitive that I should fetch them so I can cache something about them.

Ideally onPrepopulateTreeDataCache should mimic this behavior as well.

sminnee: For simplicity, we can probably refactor the whole thing to _only_ receive an ID list.

sminnee: actually, scratch that. DataList would be the only way that you could supply a lazily-evaluated list, since we can't use "iterable" as a typehint in PHP 5.6/7.0.

PR #184, opened 2018-09-24
fix inferReciprocalComponent called on unsaved

lekoala:
DataObject::inferReciprocalComponent  cannot be called on object without IDs. There is currently no checks if owner has a proper id, which could result on errors.
Also use $owner instead of $this->owner for consistency in the function.

to be honest, I'm not sure how I managed to do that, but this extra check cannot hurt in my opinion.

### DIFF
--- a/src/RecursivePublishable.php
+++ b/src/RecursivePublishable.php
@@ -199,18 +199,20 @@ class RecursivePublishable extends DataExtension
         $owner = $this->owner;
         $owners = $owner->findRelatedObjects('owned_by', false);
 
-        // Second pass: Find owners via reverse lookup list
-        foreach ($lookup as $ownedClass => $classLookups) {
-            // Skip owners of other objects
-            if (!is_a($this->owner, $ownedClass)) {
-                continue;
-            }
-            foreach ($classLookups as $classLookup) {
-                // Merge new owners into this object's owners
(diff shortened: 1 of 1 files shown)


src/RecursivePublishable.php
```diff
@@ -199,18 +199,20 @@ public function findOwnersRecursive($recursive, $list, $lookup)
...
-                continue;
-            }
-            foreach ($classLookups as $classLookup) {
-                // Merge new owners into this object's owners
-                $ownerClass = $classLookup['class'];
-                $ownerRelation = $classLookup['relation'];
-                $result = $this->owner->inferReciprocalComponent($ownerClass, $ownerRelation);
-                $owner->mergeRelatedObjects($owners, $result);
+        // Second pass: Find owners via reverse lookup list if possible
+        if ($owner->ID) {
```
kinglozzer: I’m nitpicking, but could you change this to `if ($owner->isInDB())`?

lekoala:
actually this is somehow related to this

https://github.com/silverstripe/silverstripe-asset-admin/pull/842

I'm not quite sure how I managed to have my assets messed up, but this is happening when I upload files attached to the SiteConfig that "owns" these assets.

anyway, defensive never hurts so... :-)

PR #187, opened 2018-10-14
BUG Fix between staged/unstaged objects written in different reading modes

### LINKED ISSUE #186: Relations break for unstaged objects written in different reading modes
Environment - PHP 7.1, SilverStripe 4.2.x, Silverstripe Versioned 1.2.x

Steps:
1. Create a many-many relation between unstaged and staged versioned object with an unstaged versioned through object.
2. Create many-many relation objects in both staged and live reading modes (e.g. live controller, CMS admin panel, etc.).
3. Loop through the relation in a template in live mode. Under certain circumstances, this fails to behave in the expected manner.

This is because the select query looks for `WasPublished=(0|1)` or `WasDraft=(0|1)` on unstaged objects. This is set based on the reading mode at the time of writing the row, even though for unstaged objects there is no live stage, so `WasPublished/WasDraft` are not applicable.

hchokshi:
Resolves #186

### DIFF
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -625,9 +625,14 @@ class Versioned extends DataExtension implements TemplateGlobalProvider, Resetta
         ReadingMode::validateStage($stage);
 
         // Filter on appropriate stage column in addition to date
-        $stageColumn = $stage === static::LIVE
-            ? 'WasPublished'
-            : 'WasDraft';
+        if ($this->hasStages()) {
+            $stageColumn = $stage === static::LIVE
+                ? 'WasPublished'
+                : 'WasDraft';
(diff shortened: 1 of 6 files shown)


src/Versioned.php
```diff
@@ -1165,7 +1170,7 @@ public function augmentWrite(&$manipulation)
                 unset($manipulation[$table]['fields']['Version']);
             } else {
                 // All writes are to draft, only live affect both
-                $stages = static::get_stage() === static::LIVE
+                $stages = $this->hasStages() && static::get_stage() === static::LIVE
```
tractorcow: Actually the opposite is true. If hasStages() = false, all writes affect both live and stage. You need to use `!$this->hasStages() || static::get_stage() === Static::LIVE` here.

Issue #189, opened 2018-11-13
RecursivePublishing/RecursiveRollback really slow when dealing with lots of DataObjects


maxime-rainville:
If you're dealing with lots of Pages, Files and related DataObjects, recursive publishing can be very slow and even be unmanageable.

This issue is related to an internal project that is running into this limit.

# Pull Requests
* [x] #191 
* [x] https://github.com/silverstripe/silverstripe-cms/pull/2330

maxime-rainville:
#191 speeds up recursive publishing a bit, but it's not going to resolve the issue by itself.

maxime-rainville:
silverstripe/silverstripe-cms#2330 will make a big change for pages with lots of dependent pages like our key project.

maxime-rainville:
Those two PRs should have fixed the issue.

PR #191, opened 2018-11-13
MINOR Bypass sync validation on ChangeSet when it just got built

maxime-rainville:
Provides tiny incy wincy performance boost when doing a publish recursive. Basically, if you just built a change set, you can safely assume it's in sync. So I added a flag to the publish method on the ChangeSet that allow you to bypass the sync validation.

# Parent Issue
* #189

### DIFF
--- a/src/ChangeSet.php
+++ b/src/ChangeSet.php
@@ -115,10 +115,13 @@ class ChangeSet extends DataObject
      *
      * User code should call {@see canPublish()} prior to invoking this method.
      *
-     * @throws Exception
+     * @throws BadMethodCallException ChangeSet has already been published or reverted.
+     * @throws ValidationException ChangeSet is not synced an can not be published.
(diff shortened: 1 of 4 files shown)


src/ChangeSet.php
```diff
@@ -115,18 +115,21 @@ class ChangeSet extends DataObject
...
+    public function publish($isSynced = false)
     {
         // Logical checks prior to publish
         if ($this->State !== static::STATE_OPEN) {
             throw new BadMethodCallException(
                 "ChangeSet can't be published if it has been already published or reverted."
             );
         }
-        if (!$this->isSynced()) {
+        if (!($isSynced || $this->isSynced())) {
```
ScopeyNZ: I think `!$isSynced && !$this->isSynced()` is a little more understandable.

maxime-rainville: Done.

PR #192, opened 2018-11-14
NEW Adding GraphQL operation scaffolder for rollbackRecursive on versioned objects

### LINKED ISSUE #190: Add a Rollback[TypeName] auto-scaffolded GraphQL operation
In #94 versioned was updated to consider "owned" records when reverting. Currently the operation that we expose that's intended to be (and is) used is `copyToStage`. For reasons outlined in silverstripe/silverstripe-cms#2326 this isn't really fit for purpose.

We should expose an auto-scaffolded `rollback[TypeName]` GraphQL operation that can be used instead.

ScopeyNZ:
Fixes #190 . This is a NEW targetted at 1.3 because it's fixing silverstripe/silverstripe-cms#2326

I've also marked the "CopyToStage" mutation as internal because it's quite low level and probably won't do all that you would want.

### DIFF
--- a/_config/graphql.yml
+++ b/_config/graphql.yml
@@ -22,6 +22,7 @@ SilverStripe\GraphQL\Scaffolding\Scaffolders\OperationScaffolder:
     copyToStage: SilverStripe\Versioned\GraphQL\Operations\CopyToStage
     publish: SilverStripe\Versioned\GraphQL\Operations\Publish
     unpublish: SilverStripe\Versioned\GraphQL\Operations\Unpublish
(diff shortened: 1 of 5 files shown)


maxime-rainville:
Can we get some unit test please? Maybe some usage doc?

ScopeyNZ:
I've added some tests - thanks for keeping me in check. I had a quick look for somewhere to add docs. There's no documentation in this repo and the framework repo doesn't seem to have much documentation either. @robbieaverill added some docs on using the history viewer so I'll raise a PR to use `rollback` instead of `copyToStage`.

PR #197, opened 2018-12-03
WIP: FIX Filtering the version history by a specific date is now more performant in large datasets

### LINKED ISSUE #194: Versions list query performance with large datasets
**Affected version: SS 4.3.0-rc1**

## Description

When a versioned DataObject has a large number of versions (e.g. 1,000,000), the query to retrieve the list of versions resorts to ignoring DB indexes and using filesort, which is considerably slower (1.5 seconds or so when compared to 0.0x seconds).

Example query generated by the ORM - the issue can be reproduced in an environment which has a large number of rows in the Element_Versions table:

```mysql
SELECT DISTINCT `Element_Versions`.`ClassName`,
                `Element_Versions`.`LastEdited`,
                `Element_Versions`.`Created`,
                `Element_Versions`.`Visibility`,
                `Element_Versions`.`Version`,
                `Element_Versions`.`Title`,
                `Element_Versions`.`ShowTitle`,
                `Element_Versions`.`Sort`,
                `Element_Versions`.`ExtraClass`,
                `Element_Versions`.`Style`,
                `Element_Versions`.`ParentID`,
                `Element_Versions`.`RecordID`                          AS `ID`,
                CASE
                  WHEN `Element_Versions`.`ClassName` IS NOT NULL THEN `Element_Versions`.`ClassName`
                  ELSE "DNADesign\\Elemental\\Models\\BaseElement" END AS `RecordClassName`,
                `Element_Versions`.`RecordID`,
                `Element_Versions`.`WasPublished`,
                `Element_Versions`.`WasDeleted`,
                `Element_Versions`.`WasDraft`,
                `Element_Versions`.`AuthorID`,
                `Element_Versions`.`PublisherID`
FROM `Element_Versions`
       INNER JOIN (SELECT `Element_Versions`.`RecordID`, MAX(`Element_Versions`.`Version`) AS `LatestVersion`
                   FROM `Element_Versions`
                   WHERE `Element_Versions`.`LastEdited` <= "2018-11-08 10:29:30"
                     AND `Element_Versions`.`WasDraft` = 1
                   GROUP BY `Element_Versions`.`RecordID`) AS `Element_Versions_Latest`
         ON `Element_Versions_Latest`.`RecordID` = `Element_Versions`.`RecordID` AND
            `Element_Versions_Latest`.`LatestVersion` = `Element_Versions`.`Version`
WHERE (`Element_Versions`.`ParentID` = "307341")
  AND (`Element_Versions`.`WasDeleted` = "0")
ORDER BY `Element_Versions`.`Sort` ASC
```

The analysis of this query:

```
mysql> explain SELECT DISTINCT `Element_Versions`.`ClassName`, `Element_Versions`.`LastEdited`, `Element_Versions`.`Created`, `Element_Versions`.`Visibility`, `Element_Versions`.`Version`, `Element_Versions`.`Title`, `Element_Versions`.`ShowTitle`, `Element_Versions`.`Sort`, `Element_Versions`.`ExtraClass`, `Element_Versions`.`Style`, `Element_Versions`.`ParentID`, `Element_Versions`.`RecordID` AS `ID`, CASE WHEN `Element_Versions`.`ClassName` IS NOT NULL THEN `Element_Versions`.`ClassName` ELSE "DNADesign\\Elemental\\Models\\BaseElement" END AS `RecordClassName`, `Element_Versions`.`RecordID`, `Element_Versions`.`WasPublished`, `Element_Versions`.`WasDeleted`, `Element_Versions`.`WasDraft`, `Element_Versions`.`AuthorID`, `Element_Versions`.`PublisherID` FROM `Element_Versions` INNER JOIN ( SELECT `Element_Versions`.`RecordID`, MAX(`Element_Versions`.`Version`) AS `LatestVersion` FROM `Element_Versions` WHERE `Element_Versions`.`LastEdited` <= "2018-11-08 10:29:30" AND `Element_Versions`.`WasDraft` = 1 GROUP BY `Element_Versions`.`RecordID` ) AS `Element_Versions_Latest` ON `Element_Versions_Latest`.`RecordID` = `Element_Versions`.`RecordID` AND `Element_Versions_Latest`.`LatestVersion` = `Element_Versions`.`Version` WHERE (`Element_Versions`.`ParentID` = "307341") AND (`Element_Versions`.`WasDeleted` = "0") ORDER BY `Element_Versions`.`Sort` ASC;
+----+-------------+------------------+-------+--------------------------------------------+-------------+---------+--------------------------------------------------------------------------+---------+----------------------------------------------+
| id | select_type | table            | type  | possible_keys                              | key         | key_len | ref                                                                      | rows    | Extra                                        |
+----+-------------+------------------+-------+--------------------------------------------+-------------+---------+--------------------------------------------------------------------------+---------+----------------------------------------------+
|  1 | PRIMARY     | Element_Versions | ref   | RecordID_Version,RecordID,Version,ParentID | ParentID    | 4       | const                                                                    |       2 | Using where; Using temporary; Using filesort |
|  1 | PRIMARY     | <derived2>       | ref   | <auto_key0>                                | <auto_key0> | 9       | SS_mysite.Element_Versions.RecordID,SS_mysite.Element_Versions.Version   |      10 | Using index; Distinct                        |
|  2 | DERIVED     | Element_Versions | index | RecordID_Version,RecordID                  | RecordID    | 4       | NULL                                                                     | 1023443 | Using where                                  |
+----+-------------+------------------+-------+--------------------------------------------+-------------+---------+--------------------------------------------------------------------------+---------+----------------------------------------------+
3 rows in set (0.00 sec)
```

## Notes

* The important part of this is the first row, `Using where; Using temporary; Using filesort` when compared to the second row which is `Using index`. This query between 1.5 and 3.5 seconds on average for me (key project database, 1 million records in the table).
* The query can be "fixed" by adding the record ID version into the inner join's conditions as well, but @blueo reported that this also changes the result, so it may not be the right change to make.
* If you remove the inner join then the query runs efficiently again.
* Note that 1 million records is large (for SilverStripe) but entirely realistic for a project that exists for some time. In database terms, it's not a huge number of rows.

### Purpose of the join

The join appears to exist purely to return only the "latest draft version" for the object. When removing the join, you get a few rows.

The joined result set isn't used _directly_ in any of the conditions or select fields, so it is used purely to limit the main query's results down.

If this is the purpose, we may be able to replace the join with `ORDER BY Element_Versions.Version DESC LIMIT 1` which appears to return the correct result in "0.00 seconds".

## Pull requests

- ~https://github.com/silverstripe/silverstripe-versioned/pull/197~
- [x] https://github.com/silverstripe/silverstripe-versioned/pull/213
- [x] https://github.com/silverstripe/silverstripe-cms/pull/2387
- [x] #225

robbieaverill:
Fixes #194

cc @silverstripe/core-team

### DIFF
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -641,7 +641,7 @@ class Versioned extends DataExtension implements TemplateGlobalProvider, Resetta
         (
             SELECT MAX("{$baseTable}_Versions"."Version")
             FROM "{$baseTable}_Versions" AS "{$baseTable}_Versions_Latest"
-            WHERE "{$baseTable}_Versions"."LastEdited" <= {$date}
+            WHERE "{$baseTable}_Versions_Latest"."LastEdited" <= {$date}
                 AND "{$baseTable}_Versions_Latest"."RecordID" = "{$baseTable}_Versions"."RecordID"
                 {$stageCondition}
         )
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -630,7 +630,7 @@ class Versioned extends DataExtension implements TemplateGlobalProvider, Resetta
             $stageColumn = $stage === static::LIVE
                 ? 'WasPublished'
                 : 'WasDraft';
-            $stageCondition = "AND \"{$baseTable}_Versions\".\"{$stageColumn}\" = 1";
+            $stageCondition = "AND \"{$baseTable}_Versions_Latest\".\"{$stageColumn}\" = 1";
         } else {
             $stageCondition = '';
         }
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -634,6 +634,21 @@ class Versioned extends DataExtension implements TemplateGlobalProvider, Resetta
             $stageCondition = '';
         }
 
+        // Get base query conditions and add them to the inner join as well, this helps with performance
+        // in larger data sets
+        $baseQueryConditions = '';
+        $baseQueryParams = [];
+        foreach ($query->getWhere() as $condition) {
+            foreach ($condition as $key => $value) {
+                $baseQueryConditions .= ' AND ' . $key;
+                // Normalise array values
+                if (is_array($value) && count($value) === 1) {
+                    $value = reset($value);
+                }
+                $baseQueryParams[] = $value;
+            }
+        }
+
         // Join on latest version filtered by date
         $query->addInnerJoin(
             <<<SQL
@@ -643,6 +658,7 @@ class Versioned extends DataExtension implements TemplateGlobalProvider, Resetta
             FROM "{$baseTable}_Versions"
             WHERE "{$baseTable}_Versions"."LastEdited" <= ?
                 {$stageCondition}
+                {$baseQueryConditions}
             GROUP BY "{$baseTable}_Versions"."RecordID"
             )
 SQL
@@ -654,7 +670,7 @@ SQL
             ,
             "{$baseTable}_Versions_Latest",
             20,
-            [$date]
+            array_merge([$date], $baseQueryParams)
         );
     }
 
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -10,6 +10,7 @@ use SilverStripe\Control\Director;
 use SilverStripe\Control\HTTPRequest;
 use SilverStripe\Core\ClassInfo;
 use SilverStripe\Core\Config\Config;
+use SilverStripe\Core\Convert;
 use SilverStripe\Core\Extension;
 use SilverStripe\Core\Resettable;
 use SilverStripe\Dev\Deprecation;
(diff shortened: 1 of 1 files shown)


robbieaverill:
Ok this clearly isn't going to work, will need to come up with something else

tractorcow:
Good try, but I don't think the logic would be correct trying to duplicate the conditions.

robbieaverill:
The tests that are failing are where the base query conditions include fields from an inner join, so the inner join for latest archived record can't use them. I also tried converting it to a subselect in a "where" condition, but that didn't have the same performance benefits. My next step will be to try and rewrite the query so it doesn't use joins at all, but I'm also happy to put this on the backburner for a while and wait for more input =)

robbieaverill:
> I also tried converting it to a subselect in a "where" condition, but that didn't have the same performance benefits

I think I did it wrong, which was why. I've updated the PR and pushed a new commit which correctly moved the join logic into a where condition.

robbieaverill:
Still some bugs with this - will park it for now

ScopeyNZ:
Looks like you missed one other spot in the query. I've pushed an update 🤞

ScopeyNZ:
Still not working 😞

robbieaverill:
Closing, not ready for peer review

PR #201, opened 2019-01-09
Update phpcs ruleset to PSR-12 (with some exclusions), automated and manual linting

robbieaverill:
There are still two exclusions which will require core API changes. This PR is all syntax related.

Issue: https://github.com/silverstripe/silverstripe-framework/issues/7991


composer.json
```diff
@@ -25,7 +25,7 @@
     "require-dev": {
         "phpunit/phpunit": "^7@dev",
         "silverstripe/admin": "^2",
-        "squizlabs/php_codesniffer": "^3.0"
+        "squizlabs/php_codesniffer": "3.x-dev"
```
maxime-rainville: If squizlabs pushes some unstable changes to `codesniffer`'s dev branch, won't this break our build?

I'm OK with us relying on our own dev branches. I'm less keen on relying on other people dev branches.

robbieaverill: Yeah it's a valid concern. We won't get PSR-12 until they've released a tag containing it though. I guess we could lint locally against PSR-12 and keep PSR-2 in the CI chain?

maxime-rainville: So it's just a question of waiting for `squizlabs/php_codesniffer` 3.5?

From their release logs it sounds like they introduced at least partial support for [PSR-12 in 3.3](https://github.com/squizlabs/PHP_CodeSniffer/releases/tag/3.3.0).

># PSR-12 Standard In-Progress
>
>This release contains an incomplete version of the PSR-12 coding standard. Errors found using this standard should be valid, but it will miss a lot of violations until it is complete. If you'd like to test and help, you can use the standard by running PHPCS with --standard=PSR12

robbieaverill: Ok, let's use that instead (should be installed already in Travis)

maxime-rainville:
I think codecov/path is confused. All the change is doing is correcting linting errors, so code coverage will be the same. Merging.

PR #212, opened 2019-02-07
NEW Apply versioned filters to reading state

chillu:
This is useful for augmenting methods which return a DataObject
rather than a DataList, with their own internal querying logic.
The most common example for this is SiteTree::get_by_link().

Note that I've split out the argument validation rather than duplicating it,
and throw exceptions for the bits which aren't supported in "reading mode"
(namely "state" and "latest"versions")


chillu:
Just validated that this is still required in the `1.x` branch, and that the build is still passing (gave it a kick). @unclecheese Since you wrote `SilverStripe\GraphQL\Resolvers\ApplyVersionFilters`, maybe you want to review this?

PR #213, opened 2019-02-11
FIX Filtering the version history by a specific date is now more performant in large datasets

### LINKED ISSUE #194: Versions list query performance with large datasets

robbieaverill:
This is a replacement of https://github.com/silverstripe/silverstripe-versioned/pull/197.

A high level summary of the problem:

To get the latest version, or read from a specific time in the archive, Versioned would add an inner join to the base query which would apply a combination of MAX and GROUP to find the latest available version. This means that the database performs a second query to find this information, and is ineffective at using database indexes to do so. The result is an exponential query time with larger table sizes.

This pull request replaces the inner joins used in "latest" and "archive" mode with a WHERE subselect condition (which is executed at the same time as the base query). The "_Version IS NOT NULL" condition replicates the behaviour of an inner join. This results in [substantial performance gains](https://github.com/silverstripe/silverstripe-versioned/issues/194#issuecomment-462263124) when tested with large tables).

Fixes #194 

---

Along with https://github.com/silverstripe/silverstripe-cms/pull/2387 (fixing a dodgy unit test in the CMS) the CWP kitchen sink builds are green with this patch in place: https://travis-ci.org/creative-commoners/cwp-recipe-kitchen-sink/builds/491580597

### DIFF
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -10,6 +10,7 @@ use SilverStripe\Control\Director;
 use SilverStripe\Control\HTTPRequest;
 use SilverStripe\Core\ClassInfo;
 use SilverStripe\Core\Config\Config;
+use SilverStripe\Core\Convert;
 use SilverStripe\Core\Extension;
 use SilverStripe\Core\Resettable;
 use SilverStripe\Dev\Deprecation;
@@ -629,38 +630,31 @@ class Versioned extends DataExtension implements TemplateGlobalProvider, Resetta
             $stageColumn = $stage === static::LIVE
                 ? 'WasPublished'
                 : 'WasDraft';
-            $stageCondition = "AND \"{$baseTable}_Versions\".\"{$stageColumn}\" = 1";
+            $stageCondition = "AND \"{$baseTable}_Versions_Latest\".\"{$stageColumn}\" = 1";
         } else {
             $stageCondition = '';
         }
 
-        // Join on latest version filtered by date
-        $query->addInnerJoin(
-            <<<SQL
-            (
-            SELECT "{$baseTable}_Versions"."RecordID",
-                MAX("{$baseTable}_Versions"."Version") AS "LatestVersion"
-            FROM "{$baseTable}_Versions"
-            WHERE "{$baseTable}_Versions"."LastEdited" <= ?
+        // Filter by latest version (based on date)
+        $date = DB::get_conn()->quoteString(Convert::raw2sql($date));
+        $latestVersionSql = <<<SQL
+        (
+            SELECT MAX("{$baseTable}_Versions_Latest"."Version")
+            FROM "{$baseTable}_Versions" AS "{$baseTable}_Versions_Latest"
+            WHERE "{$baseTable}_Versions_Latest"."LastEdited" <= {$date}
+                AND "{$baseTable}_Versions_Latest"."RecordID" = "{$baseTable}_Versions"."RecordID"
                 {$stageCondition}
-            GROUP BY "{$baseTable}_Versions"."RecordID"
-            )
-SQL
-            ,
-            <<<SQL
-            "{$baseTable}_Versions_Latest"."RecordID" = "{$baseTable}_Versions"."RecordID"
-            AND "{$baseTable}_Versions_Latest"."LatestVersion" = "{$baseTable}_Versions"."Version"
-SQL
-            ,
-            "{$baseTable}_Versions_Latest",
-            20,
-            [$date]
-        );
+        )
+SQL;
+        $query->addWhere([
+            "\"{$baseTable}_Versions\".\"Version\" = {$latestVersionSql}",
+            "\"{$baseTable}_Versions\".\"Version\" IS NOT NULL",
+        ]);
     }
 
     /**
      * Return latest version instances, regardless of whether they are on a particular stage.
-     * This provides "show all, including deleted" functonality.
+     * This provides "show all, including deleted" functionality.
      *
      * Note: latest_version ignores deleted versions, and will select the latest non-deleted
      * version.
@@ -674,24 +668,20 @@ SQL
 
         // Join and select only latest version
         $baseTable = $this->baseTable();
-        $query->addInnerJoin(
-            <<<SQL
-            (
-            SELECT "{$baseTable}_Versions"."RecordID",
-                MAX("{$baseTable}_Versions"."Version") AS "LatestVersion"
-            FROM "{$baseTable}_Versions"
-            WHERE "{$baseTable}_Versions"."WasDeleted" = 0
-            GROUP BY "{$baseTable}_Versions"."RecordID"
-            )
-SQL
-            ,
-            <<<SQL
-            "{$baseTable}_Versions_Latest"."RecordID" = "{$baseTable}_Versions"."RecordID"
-            AND "{$baseTable}_Versions_Latest"."LatestVersion" = "{$baseTable}_Versions"."Version"
-SQL
-            ,
-            "{$baseTable}_Versions_Latest"
-        );
+
+        // Filter by latest version
+        $latestVersionSql = <<<SQL
+        (
+            SELECT MAX("{$baseTable}_Versions_Latest"."Version")
+            FROM "{$baseTable}_Versions" AS "{$baseTable}_Versions_Latest"
+            WHERE "{$baseTable}_Versions_Latest"."WasDeleted" = 0
+              AND "{$baseTable}_Versions_Latest"."RecordID" = "{$baseTable}_Versions"."RecordID"
+        )
+SQL;
+        $query->addWhere([
+            "\"{$baseTable}_Versions\".\"Version\" = {$latestVersionSql}",
+            "\"{$baseTable}_Versions\".\"Version\" IS NOT NULL",
+        ]);
     }
 
     /**


robbieaverill:
@tractorcow had some feedback on [the previous PR](https://github.com/silverstripe/silverstripe-versioned/pull/197), but I think it was related to my initial attempt to fix the problem which is not relevant any more.

@ScopeyNZ and I have both tested this on a large project, I've added some profiles query results to the issue in a comment and the patch passes in the kitchen sink - I think that's a pretty good coverage! 😆

robbieaverill:
cc @silverstripe/core-team for a review

robbieaverill:
Right, so the inner join is inefficient for single record lookups and the subquery is inefficient for multiple record lookups. Sounds like a bit of a tough situation to be in.

ScopeyNZ:
> I think the reason why queries returning multiple results are slower is that they each record will have an individual "where" clause subquery ... as opposed to one subquery join for the entire query without the PR.

Yep I'm pretty sure this is right. However the "join" method will be less performant based on the total number of versions in your SiteTree table.

maxime-rainville:
Just thinking ... if you keep the join and take whatever where clause you have on your main query and apply it to your subquery as well, wouldn't that speed it up? e.g.:
```diff
SELECT DISTINCT `SiteTree_Versions`.`ClassName`, `SiteTree_Versions`.`LastEdited`, `SiteTree_Versions`.`Created`, `SiteTree_Versions`.`CanViewType`, `SiteTree_Versions`.`CanEditType`, `SiteTree_Versions`.`Version`, `SiteTree_Versions`.`URLSegment`, `SiteTree_Versions`.`Title`, `SiteTree_Versions`.`MenuTitle`, `SiteTree_Versions`.`Content`, `SiteTree_Versions`.`MetaDescription`, `SiteTree_Versions`.`ExtraMeta`, `SiteTree_Versions`.`ShowInMenus`, `SiteTree_Versions`.`ShowInSearch`, `SiteTree_Versions`.`Sort`, `SiteTree_Versions`.`HasBrokenFile`, `SiteTree_Versions`.`HasBrokenLink`, `SiteTree_Versions`.`ReportClass`, `SiteTree_Versions`.`RelationFieldsTestPageID`, `SiteTree_Versions`.`SubsiteID`, `SiteTree_Versions`.`ParentID`, `SiteTree_Versions`.`RecordID` AS `ID`,
      CASE WHEN `SiteTree_Versions`.`ClassName` IS NOT NULL THEN `SiteTree_Versions`.`ClassName`
      ELSE 'SilverStripe\\CMS\\Model\\SiteTree' END AS `RecordClassName`, `SiteTree_Versions`.`RecordID`, `SiteTree_Versions`.`WasPublished`, `SiteTree_Versions`.`WasDeleted`, `SiteTree_Versions`.`WasDraft`, `SiteTree_Versions`.`AuthorID`, `SiteTree_Versions`.`PublisherID`
 FROM `SiteTree_Versions` INNER JOIN             (
            SELECT `SiteTree_Versions`.`RecordID`,
                MAX(`SiteTree_Versions`.`Version`) AS `LatestVersion`
            FROM `SiteTree_Versions`
            WHERE `SiteTree_Versions`.`LastEdited` <= '2020-01-01'
                AND `SiteTree_Versions`.`WasDraft` = 1
+                AND (`SiteTree_Versions`.`RecordID` = 100)
+                AND (`SiteTree_Versions`.`WasDeleted` = 0)
            GROUP BY `SiteTree_Versions`.`RecordID`
            ) AS `SiteTree_Versions_Latest` ON             `SiteTree_Versions_Latest`.`RecordID` = `SiteTree_Versions`.`RecordID`
            AND `SiteTree_Versions_Latest`.`LatestVersion` = `SiteTree_Versions`.`Version`
WHERE (`SiteTree_Versions`.`RecordID` = 100)
 AND (`SiteTree_Versions`.`WasDeleted` = 0)
 ORDER BY `SiteTree_Versions`.`Sort` ASC
 LIMIT 1
```

In theory, it shouldn't affect the results. I re-ran some tweaked queries on my original examples: 
* Retrieving a single record was 5 times faster than the original and 25% faster than the PR
* Retrieving all records was 25% slower than than the original and 15 times faster than the PR.

It kind of makes sense that multi record query would be a bit slower because you are adding a useless condition check that will always be true. That would be a linear increase, so probably not that bad.

wilr:
Another variation to try is with DISTINCT set to false and doing a GROUP BY instead. We're running that for several queries on a CWP project (~33,000 pages) and it's halved query time for large sitetree joins but it depends on what index is used.

ScopeyNZ:
> ```
> +                AND (`SiteTree_Versions`.`RecordID` = 100)
> +                AND (`SiteTree_Versions`.`WasDeleted` = 0)
> ```

That is a much better solution. 

My assumption is that MySQL is not smart enough to plan this inner join efficiently and instead will just run that query verbatim - essentially finding the latest version for every single record in the `_Versions` table.

maxime-rainville:
On second thought, we might need to be careful what condition we put in there. Otherwise we might end up retrieving any older version that match the provided condition. It might be that we only add the `RecordID` clause to the sub query.

sminnee:
The woes of DISTINCT being tracked here - https://github.com/silverstripe/silverstripe-framework/issues/4923 - perhaps we could look at addressing this in 4.4?

robbieaverill:
@sminnee its worth mentioning that our key project is hoping for this to be in 4.3

sminnee:
OK, while this is a bugfix, I would probably advise against addressing #4923 in 4.3. So maybe tweaking the DISTINCT setting for this specific query is better for 4.3.x

ScopeyNZ:
Ok looking back at the history of this fix the suggestion of adding the filters to the inner join originally caused test failures. I think it makes sense to just limit the inner join by the ID where clause. I'll have a go at adding this in a couple of hours unless Robbie beats me to it.

ScopeyNZ:
> It might be that we only add the `RecordID` clause to the sub query.

In my testing it was difficult to find an example of a query that was generated that filters on a specific record ID. Anyway - in the cases this gets ridiculous the filter is pretty unbound. Even filtering to something as specific as a "PageID" when fetching versions of a HasMany relationship can't be done. The "PageID" is set to zero when deleted, but still needs to show in the result set.

@maxime-rainville can you let me know what page generates a list of "all versions" as you mentioned? Perhaps with that context I can think of a better solution...

At this point it's looking like some horrible "opt-in" solution. 

FYI, dropping the `DISTINCT` had negligible effect.

maxime-rainville:
That's how I got those queries.

```php
        Versioned::reading_archived_date('2020-01-01');

        SiteTree::get()->byID(100);
        /*
SELECT DISTINCT "SiteTree_Versions"."ClassName", "SiteTree_Versions"."LastEdited", "SiteTree_Versions"."Created", "SiteTree_Versions"."CanViewType", "SiteTree_Versions"."CanEditType", "SiteTree_Versions"."Version", "SiteTree_Versions"."URLSegment", "SiteTree_Versions"."Title", "SiteTree_Versions"."MenuTitle", "SiteTree_Versions"."Content", "SiteTree_Versions"."MetaDescription", "SiteTree_Versions"."ExtraMeta", "SiteTree_Versions"."ShowInMenus", "SiteTree_Versions"."ShowInSearch", "SiteTree_Versions"."Sort", "SiteTree_Versions"."HasBrokenFile", "SiteTree_Versions"."HasBrokenLink", "SiteTree_Versions"."ReportClass", "SiteTree_Versions"."RelationFieldsTestPageID", "SiteTree_Versions"."SubsiteID", "SiteTree_Versions"."ParentID", "SiteTree_Versions"."RecordID" AS "ID",
			CASE WHEN "SiteTree_Versions"."ClassName" IS NOT NULL THEN "SiteTree_Versions"."ClassName"
			ELSE 'SilverStripe\\CMS\\Model\\SiteTree' END AS "RecordClassName", "SiteTree_Versions"."RecordID", "SiteTree_Versions"."WasPublished", "SiteTree_Versions"."WasDeleted", "SiteTree_Versions"."WasDraft", "SiteTree_Versions"."AuthorID", "SiteTree_Versions"."PublisherID"
 FROM "SiteTree_Versions" INNER JOIN             (
            SELECT "SiteTree_Versions"."RecordID",
                MAX("SiteTree_Versions"."Version") AS "LatestVersion"
            FROM "SiteTree_Versions"
            WHERE "SiteTree_Versions"."LastEdited" <= '2020-01-01'
                AND "SiteTree_Versions"."WasDraft" = 1
            GROUP BY "SiteTree_Versions"."RecordID"
            ) AS "SiteTree_Versions_Latest" ON             "SiteTree_Versions_Latest"."RecordID" = "SiteTree_Versions"."RecordID"
            AND "SiteTree_Versions_Latest"."LatestVersion" = "SiteTree_Versions"."Version"
 WHERE ("SiteTree_Versions"."RecordID" = 100)
 AND ("SiteTree_Versions"."WasDeleted" = 0)
 ORDER BY "SiteTree_Versions"."Sort" ASC
 LIMIT 1
         */

        SiteTree::get()->getIterator();
        /*
SELECT DISTINCT "SiteTree_Versions"."ClassName", "SiteTree_Versions"."LastEdited", "SiteTree_Versions"."Created", "SiteTree_Versions"."CanViewType", "SiteTree_Versions"."CanEditType", "SiteTree_Versions"."Version", "SiteTree_Versions"."URLSegment", "SiteTree_Versions"."Title", "SiteTree_Versions"."MenuTitle", "SiteTree_Versions"."Content", "SiteTree_Versions"."MetaDescription", "SiteTree_Versions"."ExtraMeta", "SiteTree_Versions"."ShowInMenus", "SiteTree_Versions"."ShowInSearch", "SiteTree_Versions"."Sort", "SiteTree_Versions"."HasBrokenFile", "SiteTree_Versions"."HasBrokenLink", "SiteTree_Versions"."ReportClass", "SiteTree_Versions"."RelationFieldsTestPageID", "SiteTree_Versions"."SubsiteID", "SiteTree_Versions"."ParentID", "SiteTree_Versions"."RecordID" AS "ID", 
			CASE WHEN "SiteTree_Versions"."ClassName" IS NOT NULL THEN "SiteTree_Versions"."ClassName"
			ELSE 'SilverStripe\\CMS\\Model\\SiteTree' END AS "RecordClassName", "SiteTree_Versions"."RecordID", "SiteTree_Versions"."WasPublished", "SiteTree_Versions"."WasDeleted", "SiteTree_Versions"."WasDraft", "SiteTree_Versions"."AuthorID", "SiteTree_Versions"."PublisherID"
 FROM "SiteTree_Versions" INNER JOIN             (
            SELECT "SiteTree_Versions"."RecordID",
                MAX("SiteTree_Versions"."Version") AS "LatestVersion"
            FROM "SiteTree_Versions"
            WHERE "SiteTree_Versions"."LastEdited" <= '2020-01-01'
                AND "SiteTree_Versions"."WasDraft" = 1
            GROUP BY "SiteTree_Versions"."RecordID"
            ) AS "SiteTree_Versions_Latest" ON             "SiteTree_Versions_Latest"."RecordID" = "SiteTree_Versions"."RecordID"
            AND "SiteTree_Versions_Latest"."LatestVersion" = "SiteTree_Versions"."Version"
 WHERE ("SiteTree_Versions"."WasDeleted" = 0)
 AND ("SiteTree_Versions"."SubsiteID" IN (0))
 ORDER BY "SiteTree_Versions"."Sort" ASC
         */
```

ScopeyNZ:
We can get a "quick win" by adding the `WHERE` clause only when the query is limited to 1. It's too hard to introspect the query for relevant conditions to apply to the inner join. I was POCing this and I had things like `strpos($clause, '"' . $baseTable . '"') !== false`. Even then for `hasMany` relations the foreign ID is set to zero when deleted from its parent - which ruins the query. - Too hard.

I really do wonder whether the performance decrease for a relatively "small" site (~500 pages) is worth not having this performance increase for a large one? The difference on a small site is a few hundred ms slower. The difference on a large site is several seconds faster?

Anyway. Parking this for a little to let my brain recover. I'll be back.

ScopeyNZ:
Hey @chillu . Do you have any suggestions about an approach here? I'm not sure that there's any suggestion I haven't investigated so we're basically left with:

- What is there
- What this PR proposes
- Both, with some undefined way of switching between the two (maybe based on the LIMIT of the query).

You're welcome to read the PR for some context but in summary:

Speed of the old method depends on the size of the `_Versioned` table. The smaller it is the faster it goes. Speed of the new method is based on the number of records being pulled. Pull 500 records at a time and it's slower.

I'm not actually sure of anywhere a query is generated that pulls a large list of versioned objects but it did come up as part of Max's testing.

chillu:
I don't see `SiteTree::get()->getIterator()` as a completely unbounded call to be a realistic use case, either in `_versions` tables or in the base table. That's just too many records. If it's still slow when you apply `LIMIT 20`, then we *might* have an issue - because my assumption is that the "version list" interface in the CMS gives you a paginated list of 20 versions at a time.

Priorities from my perspective:
1. Efficient archive view (single version per record) - that's what the underlying project raised, right?
2. Efficient selection of a small number of versions per single record (<50)
3. Efficient selection of a large number of versions per single record (>50). That's not a goal as such, unless somebody can point me to a usage path where the CMS fires off such queries.

maxime-rainville:
FYI just tried Ingo's limit use case with my ~1500 versions. 
* With a 20 limit the PR is about 40% faster
* With a 30 limit the PR is about even with the legacy code
* With a 100 limit the PR is about 2500% slower

chillu:
Please quote percentages only alongside absolut measurements. We all acknowledge those are different based on execution context. 2.5ms is 400% slower than 10ms. It's also irrelevant on this scale.

robbieaverill:
I see there being three options:

1. We merge this as it is and accept that smaller websites may have a performance loss
2. We change this so that it could be documentation for medium-large size websites, and implement as a project specific implementation
3. We go back to the drawing board again and try something else

chillu:
@maxime-rainville @robbieaverill The current solution might be good enough, but I haven't seen any actual measurements - just vague statements. 

> We merge this as it is and accept that smaller websites may have a performance loss

That really depends on *how much* of a loss.

robbieaverill:
@maxime-rainville are you able to provide some time metrics?

maxime-rainville:
I was testing straigh SQL queries. When I got a sec, I'll test an actual query trough a silverstripe task and maybe find a realistic scenario like listing versions of a page in versioned-admin.

src/Versioned.php
```diff
@@ -629,38 +630,31 @@ protected function augmentSQLVersionedArchive(SQLSelect $query, DataQuery $dataQ
...
-                MAX("{$baseTable}_Versions"."Version") AS "LatestVersion"
-            FROM "{$baseTable}_Versions"
-            WHERE "{$baseTable}_Versions"."LastEdited" <= ?
+        // Filter by latest version (based on date)
+        $date = DB::get_conn()->quoteString(Convert::raw2sql($date));
+        $latestVersionSql = <<<SQL
+        (
+            SELECT MAX("{$baseTable}_Versions_Latest"."Version")
+            FROM "{$baseTable}_Versions" AS "{$baseTable}_Versions_Latest"
+            WHERE "{$baseTable}_Versions_Latest"."LastEdited" <= {$date}
```
tractorcow: Why did you de-parameterise this query? Non-parameterised queries aren't cacheable in the DB backend.

src/Versioned.php
```diff
@@ -629,38 +630,31 @@ protected function augmentSQLVersionedArchive(SQLSelect $query, DataQuery $dataQ
...
-            (
-            SELECT "{$baseTable}_Versions"."RecordID",
-                MAX("{$baseTable}_Versions"."Version") AS "LatestVersion"
-            FROM "{$baseTable}_Versions"
-            WHERE "{$baseTable}_Versions"."LastEdited" <= ?
+        // Filter by latest version (based on date)
+        $date = DB::get_conn()->quoteString(Convert::raw2sql($date));
+        $latestVersionSql = <<<SQL
+        (
+            SELECT MAX("{$baseTable}_Versions_Latest"."Version")
```
tractorcow: It's not "latest" if it's a historic archive date is it? Feels misleading.

src/Versioned.php
tractorcow: Why did you de-parameterise this query? Non-parameterised queries aren't cacheable in the DB backend.

src/Versioned.php
```diff
@@ -629,38 +630,31 @@ protected function augmentSQLVersionedArchive(SQLSelect $query, DataQuery $dataQ
...
+            FROM "{$baseTable}_Versions" AS "{$baseTable}_Versions_Latest"
+            WHERE "{$baseTable}_Versions_Latest"."LastEdited" <= {$date}
+                AND "{$baseTable}_Versions_Latest"."RecordID" = "{$baseTable}_Versions"."RecordID"
```
tractorcow: This condition on the outer row turns the inner query from a non-correlated subquery to a correlated subquery. The performance hit is that this query will need to be run for every ROW in the outer query.

See https://en.wikipedia.org/wiki/Correlated_subquery

The non-correlated subquery (previously) would only be executed once and then joined internally.

It's going to be super slow on large datasets.

robbieaverill: Perhaps that’s what Maxime found- many rows in the outer query rather than one. Unfortunately it’s already super slow on large datasets since the existing inner join queries every row without using indexes. Do you have another suggestion for how we could fix this?

tractorcow: You could look at promoting the limit up to the inner query as well; That way instead of joining two large lists, and limiting the outer list only (and thus the results) maybe you would end up joining only two smaller lists?

ScopeyNZ:
Ok I'll close this and raise a new PR with a new approach... When I come up with it.

src/Versioned.php
ScopeyNZ: This is certainly what Max found. The problem is the non-correlated subquery will run for every record in the versions table - this is a pretty big flaw in the query planner in MySQL imo. This means that as the `"{$baseTable}_Versions"` table gets larger the query gets slower. 

The correlated subquery is faster in this case - although it gets run for every row - I haven't found a use case where more than ~20 rows are pulled. 

It's too hard to promote the condition to the inner join - in some cases all the conditions aren't applied, and even if they were it's difficult to introspect those that actually matter. In some cases some parts of the condition shouldn't be promoted to the inner join.

Again this highlights a flaw with the MySQL query planner. Why can't it be intelligent enough to auto-promote the `ID` condition considering it's a join condition...

brynwhyman:
New PR at https://github.com/silverstripe/silverstripe-versioned/pull/225

Issue #216, opened 2019-02-24
versioned 1.3 incompatible with silverstripe 4.1


b-rhodes:
When using versioned with silverstripe-framework 4.1.5, I get the error:
```
PHP Fatal error:  Interface 'SilverStripe\Forms\GridField\GridField_ActionMenuItem' not found in /var/www/summerset.local/vendor/silverstripe/versioned/src/GridFieldArchiveAction.php on line 20
```
And according to the [Silverstripe 4.2 Changelog](https://docs.silverstripe.org/en/4/changelogs/4.2.0/), GridField_ActionMenuItem was only created in silverstripe 4.2.
However the composer.json for versioned just requires `"silverstripe/framework": "^4"`

robbieaverill:
Would you like to make a PR to update the composer dependency to ^4.2?

b-rhodes:
Done, here it is:
https://github.com/silverstripe/silverstripe-versioned/pull/217

robbieaverill:
Thanks - this should prevent people from installing incompatible versions of framework/cms with versioned in future. Cheers @b-rhodes!

PR #217, opened 2019-02-25
Fixed Wrong Silverstripe Requirement

b-rhodes:
Fixed wrong silver-stripe version requirement as detailed here:
https://github.com/silverstripe/silverstripe-versioned/issues/216

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         }
     ],
     "require": {
-        "silverstripe/framework": "^4",
+        "silverstripe/framework": "^4.2",
         "silverstripe/vendor-plugin": "^1"
     },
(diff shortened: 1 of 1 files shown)


robbieaverill:
I suggest we re-target this fix at the 1.2 branch

b-rhodes:
Hmm, that might not have been the right way to change this. Do you want me to close this and create another, forking off of 1.2?

robbieaverill:
It's OK, you just need to rebase the branch now :) I can do this for you if you want

b-rhodes:
That'd be good, I don't really know my way around git/github super well. Thanks 😄

PR #220, opened 2019-03-11
onAfterPublishRecursive extension hook added.

mfendeksilverstripe:
* `onAfterPublishRecursive` extension hook added
* this is useful for situations where parent object is being published and it has a bunch of nested objects that will get published as well
* this extension hook allows executing some actions `after` all nested objects are published as well
* note that `onAfterPublish` extension hook is executed before the nested objects are published
* example use: Page with Blocks, trigger publish, generate static cache after `all` objects are published, not just the page


mfendeksilverstripe:
@chillu This PR needs some feedback

robbieaverill:
Could you re-target the PR at the 1 branch?

mfendeksilverstripe:
@robbieaverill Switched base to `1`

chrispenny:
Hay team, any update on when this one might move through? It's a very useful extension point.

src/RecursivePublishable.php
```diff
@@ -57,20 +57,39 @@ class RecursivePublishable extends DataExtension
...
-        $changeset->addObject($this->owner);
-        return $changeset->publish(true);
+        $changeset->addObject($owner);
+
+        $result = $changeset->publish(true);
+        if (!$result) {
+            return $result;
+        }
+
+        $owner->invokeWithExtensions('onAfterPublishRecursive', $original);
```
robbieaverill: I wonder whether this should still run when publish fails? You might want to handle failures from extensions

chillu: @chrispenny I think this is waiting for a response from you

chrispenny: Perhaps it would be useful to have a separate `failure` extension point.

I'm not sure this needs to be added as part of this change set though.

robbieaverill: Sure, I think what I was suggesting was that you put `$owner->invoke...` before `return $result` and allow it to have a null argument passed

src/RecursivePublishable.php
```diff
@@ -57,20 +57,39 @@ class RecursivePublishable extends DataExtension
      */
     public function publishRecursive()
     {
+        /** @var DataObject|Versioned $owner */
+        $owner = $this->owner;
+
+        // get the last published version
+        $original = null;
+        if ($owner->hasExtension(Versioned::class) && $owner->isPublished()) {
+            $original = Versioned::get_by_stage($owner->baseClass(), Versioned::LIVE)
```
chillu: In case where this version isn't already cached, this impacts publication performance through additional queries - something I think you've been watching out for @chrispenny and @mfendeksilverstripe  ;) Technically we could just pass class+ID to the hook. But given that the existing `onAfterPublish` hook passes `$original` as well, I don't think we should create inconsistent APIs, so this is a necessary performance impact if we need the hook.

src/RecursivePublishable.php
chillu: Yep, agree that we shouldn't invoke this on failed publications. Looking at `ChangeSet->publish()`, there's actually no branch where this could return `false`, but we shouldn't codify that assumption here.

chrispenny: Thanks for the input, team! With EoY coming up very soon, I might not have a chance to take these changes on until the new year.

mfendeksilverstripe:
@chillu `onBeforePublishRecursive` added as per your request

chillu:
Thanks for your patience! :)

PR #225, opened 2019-03-25
FIX Implemented various options to improve archive query performance

ScopeyNZ:
This is a performance "fix" that should have no impact on the existing functionality.

This is a follow-up to #213 . There was concern in that PR that the solution we came up with (changing the non-correlated subquery to a correlated one) had too many performance concerns. The suggestion was provided to "promote" certain conditions from the outer query to the inner join. The only condition that can be promoted is a filter on the primary key of the base table. This PR adds that which adds a small performance improvement.

I've also re-added the functionality from #213 but it relies on a hook added in silverstripe/silverstripe-framework#8915 or a config setting. In my large project I found zero scenarios where the inner join was more performant than the condition - and it is ~5s (browsing to a page with `?archiveDate`) rather than ~200s.

I've edited this to reflect the new state of this PR - it was originally doing more "condition promoting" which was flawed logic.

### DIFF
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -201,6 +201,16 @@ class Versioned extends DataExtension implements TemplateGlobalProvider, Resetta
     private static $prepopulate_versionnumber_cache = true;
 
     /**
+     * Indicates whether augmentSQL operations should add subselects as WHERE conditions instead of INNER JOIN
+     * intersections. Performance of the INNER JOIN scales on the size of _Versions tables where as the condition scales
+     * on the number of records being returned from the base query.
+     *
+     * @config
+     * @var bool
+     */
+    private static $use_conditions_over_inner_joins = false;
+
+    /**
      * Additional database indexes for the new
      * "_Versions" table. Used in {@link augmentDatabase()}.
      *
@@ -611,6 +621,92 @@ class Versioned extends DataExtension implements TemplateGlobalProvider, Resetta
     }
 
     /**
+     * Prepare a sub-select for determining latest versions of records on the base table. This is used as either an
+     * inner join or sub-select on the base query
+     *
+     * @param SQLSelect $baseQuery
+     * @return SQLSelect
+     */
+    protected function prepareMaxVersionSubSelect(SQLSelect $baseQuery)
+    {
+        $baseTable = $this->baseTable();
+
+        // Create a sub-select that we determine latest versions
+        $subSelect = SQLSelect::create(
+            ['LatestVersion' => "MAX(\"{$baseTable}_Versions_Latest\".\"Version\")"],
+            [$baseTable . '_Versions_Latest' => "\"{$baseTable}_Versions\""]
+        );
+
+        $subSelect->renameTable($baseTable, "{$baseTable}_Versions");
+
+        // Determine the base table of the existing query
+        $baseFrom = $baseQuery->getFrom();
+        $baseTable = trim(reset($baseFrom), '"');
+
+        // And then the name of the base table in the new query
+        $newFrom = $subSelect->getFrom();
+        $newTable = trim(key($newFrom), '"');
+
+        // Parse "where" conditions to find those appropriate to be "promoted" into an inner join
+        // We can ONLY promote a filter on the primary key of the base table. Any other conditions will make the
+        // version returned incorrect, as we are filtering out version that may be the latest (and correct) version
+        foreach ($baseQuery->getWhere() as $condition) {
+            $conditionClause = key($condition);
+            // Pull out the table and field for this condition. We'll skip anything we can't parse
+            if (preg_match('/^"([^"]+)"\."([^"]+)"/', $conditionClause, $matches) !== 1) {
+                continue;
+            }
+
+            $table = $matches[1];
+            $field = $matches[2];
+
+            if ($table !== $baseTable || $field !== 'RecordID') {
+                continue;
+            }
+
+            // Rename conditions on the base table to the new alias
+            $conditionClause = preg_replace(
+                '/^"([^"]+)"\./',
+                "\"{$newTable}\".",
+                $conditionClause
+            );
+
+            $subSelect->addWhere([$conditionClause => reset($condition)]);
+        }
+
+        return $subSelect;
+    }
+
+    /**
+     * Inspect a given query that's being "augmented" and determine any conditions that can be "promoted" to an inner
+     * join on the same table. This helps with performance
+     *
+     * @param SQLSelect $query
+     * @param SQLSelect $newQuery
+     */
+    protected function promoteConditions(SQLSelect $baseQuery, SQLSelect $newQuery)
+    {
+
+    }
+
+    /**
+     * Indicates if a subquery filtering versioned records should apply as a condition instead of an inner join
+     *
+     * @param SQLSelect $baseQuery
+     */
+    protected function shouldApplySubSelectAsCondition(SQLSelect $baseQuery)
+    {
+        $baseTable = $this->baseTable();
+
+        $shouldApply =
+            $baseQuery->getLimit() === 1 || Config::inst()->get(static::class, 'use_conditions_over_inner_joins');
+
+        $this->owner->extend('updateApplyVersionedFiltersAsConditions', $shouldApply, $baseQuery, $baseTable);
+
+        return $shouldApply;
+    }
+
+    /**
      * Filter the versioned history by a specific date and archive stage
      *
      * @param SQLSelect $query
@@ -631,29 +727,36 @@ class Versioned extends DataExtension implements TemplateGlobalProvider, Resetta
         $stage = $dataQuery->getQueryParam('Versioned.stage');
         ReadingMode::validateStage($stage);
 
+        $subSelect = $this->prepareMaxVersionSubSelect($query);
+
+        $subSelect->addWhere(["\"{$baseTable}_Versions_Latest\".\"LastEdited\" <= ?" => $date]);
+
         // Filter on appropriate stage column in addition to date
         if ($this->hasStages()) {
             $stageColumn = $stage === static::LIVE
                 ? 'WasPublished'
                 : 'WasDraft';
-            $stageCondition = "AND \"{$baseTable}_Versions\".\"{$stageColumn}\" = 1";
-        } else {
-            $stageCondition = '';
+            $subSelect->addWhere("\"{$baseTable}_Versions_Latest\".\"{$stageColumn}\" = 1");
         }
+
+        if ($this->shouldApplySubSelectAsCondition($query)) {
+            $subSelect->addWhere(
+                "\"{$baseTable}_Versions_Latest\".\"RecordID\" = \"{$baseTable}_Versions\".\"RecordID\""
+            );
+
+            $query->addWhere([
+                "\"{$baseTable}_Versions\".\"Version\" = ({$subSelect->sql($params)})" => $params,
+            ]);
+
+            return;
+        }
+
+        $subSelect->addSelect("\"{$baseTable}_Versions_Latest\".\"RecordID\"");
+        $subSelect->addGroupBy("\"{$baseTable}_Versions_Latest\".\"RecordID\"");
 
         // Join on latest version filtered by date
         $query->addInnerJoin(
-            <<<SQL
-            (
-            SELECT "{$baseTable}_Versions"."RecordID",
-                MAX("{$baseTable}_Versions"."Version") AS "LatestVersion"
-            FROM "{$baseTable}_Versions"
-            WHERE "{$baseTable}_Versions"."LastEdited" <= ?
-                {$stageCondition}
-            GROUP BY "{$baseTable}_Versions"."RecordID"
-            )
-SQL
-            ,
+            '(' . $subSelect->sql($params) . ')',
             <<<SQL
             "{$baseTable}_Versions_Latest"."RecordID" = "{$baseTable}_Versions"."RecordID"
             AND "{$baseTable}_Versions_Latest"."LatestVersion" = "{$baseTable}_Versions"."Version"
@@ -661,7 +764,7 @@ SQL
             ,
             "{$baseTable}_Versions_Latest",
             20,
-            [$date]
+            $params
         );
     }
 
@@ -692,7 +795,7 @@ SQL
 
     /**
      * Return latest version instances, regardless of whether they are on a particular stage.
-     * This provides "show all, including deleted" functonality.
+     * This provides "show all, including deleted" functionality.
      *
      * Note: latest_version ignores deleted versions, and will select the latest non-deleted
      * version.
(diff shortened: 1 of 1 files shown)


src/Versioned.php
```diff
@@ -634,6 +634,35 @@ protected function augmentSQLVersionedArchive(SQLSelect $query, DataQuery $dataQ
...
+            // to some other object in a HasMany relation)
+            if (strpos($conditionClause, "\"{$baseTable}\".") === 0 && strpos($conditionClause, 'ID') !== false) {
+                $innerJoinPromotedConditions[] =
+                    [str_replace($baseTable, $baseTable . '_Versions', $conditionClause) => reset($condition)];
+            }
+        }
+
+        // If there are conditions that are to be "promoted" to the inner join then add them as another subquery
+        if (count($innerJoinPromotedConditions)) {
+            $select = new SQLSelect("\"{$baseTable}_Versions\".\"RecordID\"", "\"{$baseTable}_Versions\"");
```
robbieaverill:
```suggestion
            $select = SQLSelect::create("\"{$baseTable}_Versions\".\"RecordID\"", "\"{$baseTable}_Versions\"");
```

ScopeyNZ:
I've updated the code with the suggestion to add it to the other inner join.

ScopeyNZ:
Does anyone have any thoughts on this considering the older PR this has replaced (#213)?

/cc @tractorcow @maxime-rainville

unclecheese:
@ScopeyNZ Did you do any benchmarks for this? I'm curious what kind of improvements you were getting. I am not seeing any significant gains.

src/Versioned.php
```diff
@@ -603,6 +603,48 @@ protected function augmentSQLVersioned(SQLSelect $query)
...
 
+    /**
+     * Inspect a given query that's being "augmented" and determine any conditions that can be "promoted" to an inner
+     * join on the same table. This helps with performance
+     *
+     * @param SQLSelect $query
+     * @param string $columnMatch Indicate a substring that should be matched to the column condition being promoted
+     * @return array(string, array) Returns a composite of the condition to be added and also the params to be added
+     */
+    protected function createPromotedCondition(SQLSelect $query, $columnMatch = 'ID')
```
unclecheese: Are there uses for this functionality outside of Versioned? Could this be refactored to be part of the `SQLSelect` API?

ScopeyNZ: Interesting question. Possibly. I consider this a bit of a "hack for performance gains". I'm not sure that we really want to promote (pun intended) this kind of way of altering queries. We're trapped into it a bit here because we're already in an `alterDataQuery` extension point.

ScopeyNZ:
I don't have specific reports from something like Blackfire.

On a large project I've been involved in I was getting these queries taking as long as 5s. Now they're down to a few hundred ms max.

ScopeyNZ:
Okay @unclecheese has shown me a situation where this doesn't meet the A/Cs. Relevant conditions aren't promoted if they're on a joined table. This happens when fetching all of a particular subclass of a DataObject.

I guess I'll have to further enhance this to cater for those joined conditions too...

src/Versioned.php
```diff
@@ -603,6 +603,48 @@ protected function augmentSQLVersioned(SQLSelect $query)
...
+
+        foreach ($whereConditions as $condition) {
+            $conditionClause = key($condition);
+            // Only pull conditions that are on the right table and are specific to IDs (RecordID, ParentID, ie. belongs
+            // to some other object in a HasMany relation)
+            if (strpos($conditionClause, "\"{$baseTable}\".") === 0
+                && strpos($conditionClause, $columnMatch) !== false
+            ) {
+                $conditions[] =
+                    [str_replace($baseTable, $baseTable . '_Versions', $conditionClause) => reset($condition)];
```
tractorcow: You can't replace a table name in an entire condition. You'll match and replace values in quoted strings also.

E.g. `WHERE "ClassName" = 'Page'` would become `WHERE "ClassName" = 'Page_Versions'`

tractorcow:
This code looks really dangerous. You're increasing the complexity of an already complex part of core (versioned). I'd like to see benchmarks and before / after queries generated by this code to see the benefits vs the cost in complexity.

src/Versioned.php
tractorcow: Or even `WHERE "MainPageHolder"."Title" = ?` would become `WHERE "MainPage_VersionsHolder" = ?`.

ScopeyNZ:
Yep I can relate that this is some pretty awful looking code. I tried the converting to `WHERE SELECT` (#213) but that had its own problems (although more on this soon). 

The real-life use case comes from an elemental table with 3.5 million records - simply counting this table takes 600ms (locally). Running the archive view queries on `SiteTree` that has the standard elemental extensions & relationships takes several seconds (seeing >10s locally). 

The original solution - converting to a subselect condition reduces that to <.5s - all versioned queries generated during testing were limited in some form.

This update is similar - pulling columns that are IDs from the base query and promoting it to the inner join (of the same table) is really something that the planner should do. It only makes sense that it should be filtered by the at least the same conditions as the inner join will remove records that aren't returned by the left side of the join. Identifying the columns and dealing with the `_Versions` magical suffix that is added is the troubling bit. It's too much to pull in every condition that you identified so it has some pretty ugly logic to identify what should be promoted.

I am all ears for suggestions on implementing a better way to discover conditions to promote.

Even after all of that @unclecheese helped me identify that it's not perfect - and there are some cases where this solution just is not practical. So I'm changing my mind again. __I'm going to re-implement the logic from #213__. However this time it's conditional on a variable I've added to the `augmentSQL` extension point in silverstripe/silverstripe-framework#8915 . Basically it makes sense to use the condition when fetching only one record or when performing an `EXISTS` query. It's going to be faster. 

Unfortunately this doesn't cover all the queries that we've seen _significant_ improvements on. If there's no acceptable solution with this approach then I'll have to figure out some way to hook into this in project code. It's a little frustrating that Versioning falls over when presented a large set of versions though 😢

tractorcow:
How does this code work on queries that filter on tables higher level than the base table? Why do we only prioritise base table fields? I think with a bit of work it could iterate over the hierarchy safely, although you will need to use `LEFT JOIN` in place of `INNER JOIN` in case of missing rows right?

tractorcow:
Good work attacking this by the way @ScopeyNZ. It doesn't look easy. :)

ScopeyNZ:
> How does this code work on queries that filter on tables higher level than the base table? Why do we only prioritise base table fields? 

I've been testing performance improvements on the new history viewer. It doesn't seem to generate any queries that filter on base table fields in the codebase I've been testing on (which is a huge project). In saying that though I've now started testing by trying to just load pages with `?archiveDate=[past date]` and I'm seeing that this is a big problem. I'm going to be tackling this now over the next couple of days

> I think with a bit of work it could iterate over the hierarchy safely, although you will need to use `LEFT JOIN` in place of `INNER JOIN` in case of missing rows right?

I think my unfamiliarity with parts of the ORM has let me down here. I'll try to be more informed with the code now that I have to loop over relations. I don't think we need to use `LEFT JOIN` here otherwise the purpose of the inner join is lost. We only want one record from the left side for every result in the added join. I personally think the logic is much more readable as a sub-select in the `WHERE` clause.

For anyone who might be reading and not following, a simplified representation of this code is doing (written as a sub-select in the where clause not the inner-join variant):

```
SELECT * 
FROM DataObject_Versions a 
WHERE 
  Version = (
    SELECT MAX(Version) 
    FROM DataObject_Versions b 
    WHERE date < :archiveDate AND a.ID = b.ID
  ) 
  AND ID = 1 
```

ScopeyNZ:
Ok I've realised that I'm not as smart as the query planner. This whole "promoting conditions" idea is seriously flawed. I can only ever promote a condition that applies to the foreign key of the base table that is being queried - everything else can make the result incorrect.

Frankly, this query will always be slow. I'll try to figure out some API so you can choose between an inner join or the subselect as depending on the expected results and the size of the versions table the subselect is way quicker.

tractorcow:
It's a real trouble! I've had many battles with this code, and I'm fully aware how sub-optimal the SQL is here. It's accurate and precise, but not optimised for speed, unfortunately.

tractorcow:
If you want to get some more performance try addressing https://github.com/silverstripe/silverstripe-framework/issues/4923 for a global performance boost. :P

ScopeyNZ:
> If you want to get some more performance try addressing [silverstripe/silverstripe-framework#4923](https://github.com/silverstripe/silverstripe-framework/issues/4923) for a global performance boost. :P

Haha thanks for the suggestion. I've got a few more things to pursue related to this - I'll add that to the list.

ScopeyNZ:
Ok I've rewritten this PR. I've made the approach a lot cleaner and I've reintroduced the "sub-query as condition" logic - but it's only used when it is likely to be more effective. See the OP for more details.

src/Versioned.php
ScopeyNZ: I've addressed this now by making the code less awful 🙂

unclecheese:
Tested and working. Significant improvements on _Versions tables with > 3M rows.

src/Versioned.php
```diff
@@ -603,6 +613,93 @@ protected function augmentSQLVersioned(SQLSelect $query)
...
+     * Indicates if a subquery filtering versioned records should apply as a condition instead of an inner join
+     *
+     * @param SQLSelect $baseQuery
+     */
+    protected function shouldApplySubSelectAsCondition(SQLSelect $baseQuery)
+    {
+        $baseTable = $this->baseTable();
+
+        $shouldApply =
+            $baseQuery->getLimit() === 1 || Config::inst()->get(static::class, 'use_conditions_over_inner_joins');
```
robbieaverill: By using `static::class` here you'd be losing base class config settings if the class were ever subclasses (like Fluent does). Perhaps use `$this->config()->get('use_conditions_over_inner_joins')` instead

tractorcow: Wouldn't that only be a problem if you use `Config::UNINHERITED`?

robbieaverill: Well if you initiate it from static::class then it's the LSB class name, so presumably wouldn't load this class's config

tractorcow:
Ok, I have one problem.

If you are promoting the where to the inner query that will get the max version, yes it WILL work (not-error) and produce a smaller set.

However, what will happen is that the extra condition would produce the wrong version ID for records that transitively match / don't match the given condition.

E.g. we have a condition `Name = 'Bob'` If the latest version has `Name = 'Andrew'` then the old code would not match this (row would not appear), because it would get the latest version, and then compare the name to see if it is Bob. It would be false.

However, because we are now applying the where to the grouping condition, it actually applies the condition BEFORE doing the group by. Effectively, instead of finding the latest record, and matching it if it's name is Bob, it will navigate the entire history of that record, and pretend that the latest record that matches the condition is the latest record. In the case above, it would pretend that `Name = 'Bob'` is the latest record, when in fact Bob has actually changed his name to Andrew.

In order to fix this, you can still do this with a bit of work, but you MUST promote the query from a WHERE to a HAVING. This is because HAVING filters the post-grouped result, and WHERE filters the result prior to grouping.

Sorry to be the bearer of bad / hard to implement news, but it's a good thing I gave it a review. :P

ScopeyNZ:
Thanks @tractorcow for the continued review on this. I'm sorry I'm taking so long between looking at this - I've been busy with other work...

> However, because we are now applying the where to the grouping condition, it actually applies the condition BEFORE doing the group by.

I'm a little confused about this statement. With this code the query either remains the same (with an inner join that has a group by) or if the config is set then the query is applied as a condition and there is _no_ group by.

Here's a simple example of how the SQL differs

#### Default:

```sql
SELECT DISTINCT
   "SiteTree_Versions"."ID",
   -- ...
FROM
   "SiteTree_Versions"
   INNER JOIN
      (
         SELECT
            MAX("SiteTree_Versions_Latest"."Version") AS "LatestVersion",
            "SiteTree_Versions_Latest"."RecordID"
         FROM
            "SiteTree_Versions" AS "SiteTree_Versions_Latest"
         WHERE
            "SiteTree_Versions_Latest"."LastEdited" <= '2019-04-01'
            AND "SiteTree_Versions_Latest"."WasDraft" = 1
         GROUP BY
            "SiteTree_Versions_Latest"."RecordID"
      )
      AS "SiteTree_Versions_Latest"
      ON "SiteTree_Versions_Latest"."RecordID" = "SiteTree_Versions"."RecordID"
      AND "SiteTree_Versions_Latest"."LatestVersion" = "SiteTree_Versions"."Version"
WHERE
   "SiteTree_Versions"."ParentID" <> "SiteTree_Versions"."RecordID"
   AND "SiteTree_Versions"."ParentID" = 1
   AND "SiteTree_Versions"."ShowInMenus" = 1
   AND "SiteTree_Versions"."WasDeleted" = 0
ORDER BY
   "SiteTree_Versions"."Sort" ASC
```

### Sub-select

```sql
SELECT DISTINCT
   "SiteTree_Versions"."ID"
   -- ...
FROM
   "SiteTree_Versions"
WHERE
   "SiteTree_Versions"."ParentID" <> "SiteTree_Versions"."RecordID"
   AND "SiteTree_Versions"."ParentID" = 1
   AND "SiteTree_Versions"."ShowInMenus" = 1
   AND "SiteTree_Versions"."WasDeleted" = 0
   AND "SiteTree_Versions"."Version" = 
      (
         SELECT
            MAX("SiteTree_Versions_Latest"."Version") AS "LatestVersion"
         FROM
            "SiteTree_Versions" AS "SiteTree_Versions_Latest"
         WHERE
            "SiteTree_Versions_Latest"."LastEdited" <= '2019-04-01'
            AND "SiteTree_Versions_Latest"."WasDraft" = 1
            AND "SiteTree_Versions_Latest"."RecordID" = "SiteTree_Versions"."RecordID"
      )
   )
ORDER BY
   "SiteTree_Versions"."Sort" ASC
```

The subselect, whether it be an inner join or a condition, can never be filtered by anything other than the primary key on the base table - because otherwise you're right that you'll get incorrect results based on your filters. If we think about the purpose of this query, filter only on the "current" version of each record - depending on the date - the inner join obviously does this, we're doing an aggregate query for all records of the given DataObject (in this case, SiteTree) - the grouping obviously makes sense here because we want the aggregate grouped by the primary key of the base table to join on.

The subselect essentially does the same thing, without needing to aggregate the whole table. No grouping is added as it's sort of doing a JIT subselect to filter out anything matching that's not the "current" version. I assume the planner is smart enough to run the query once for unique `"RecordID"`s. Both of these happen before a grouping on the query as a whole.

Perhaps I'm missing something but I really don't understand how these queries differ right now? In terms of results? Am I missing some use case here? It wouldn't surprise me - the performance disparity is pretty intense on huge projects. In a bunch of automated and manual testing it appears to work as anticipated? This is comparing before and after on complex pages using `archiveDate=date`.

If you describe some scenarios I can try to replicate them in automated tests.

ScopeyNZ:
The `promoteConditions` API should probably be refactored/renamed. I was under the incorrect assumption in the past that I could promote more than just the primary key on the base table. That's not true though...

tractorcow:
@ScopeyNZ you need to consider an example that has an outer condition to be "promoted".

E.g.

```
Versioned::reading_archived_date('2019-01-03', Versioned::DRAFT);
$result = SiteTree::get(); // works fine, because no query to promote to inner
$result = SiteTree::get()->filter('Title', 'Some old page'); // Title filter added to inner query incorrectly
```

tractorcow:
@ScopeyNZ a real world case where this could mess up; A dataobject could be removed from a parent object (ParentID set to 0). Restoring the parent object (recursively) to a point in time after this would incorrectly ignore the ParentID = 0, and see the earlier record matching ParentID = theoldid, incorrectly restoring the deleted record.

ScopeyNZ:
With this PR the conditions are _not_ promoted though. I think I should change the code I've written to remove that confusing API. ONLY the `BaseTable.ID` can (and will) apply to the subselect.

tractorcow:
>  I think I should change the code I've written to remove that confusing API

ok, sorry that makes sense. I'll let you do that. :D

ScopeyNZ:
Ok I've done that. I've removed the confusing `promoteConditions` API and left a comment to be very clear about what can be done for future developers browsing this code.

src/Versioned.php
```diff
@@ -603,6 +613,92 @@ protected function augmentSQLVersioned(SQLSelect $query)
...
+     * Inspect a given query that's being "augmented" and determine any conditions that can be "promoted" to an inner
+     * join on the same table. This helps with performance
+     *
+     * @param SQLSelect $query
+     * @param SQLSelect $newQuery
+     */
+    protected function promoteConditions(SQLSelect $baseQuery, SQLSelect $newQuery)
+    {
+
+    }
```
robbieaverill: Do you still need this function?

ScopeyNZ:
I've rebased onto 1.4 and added an extension point for the subquery.

unclecheese:
Haven't heard from @tractorcow on the latest commit, but it's clear that he's implemented what he asked for, and that was the last unresolved bit, so I'm merging this.

ScopeyNZ:
Thanks @unclecheese . To be clear for people looking back on this in the future: this PR doesn't do anything by default but adds configuration to offer an alternative method of doing archive queries when working with very large versioned tables (several million records).

PR #227, opened 2019-04-08
MINOR: Add test to validate writeWithoutVersioned

sminnee:
Created as part of looking into 
https://github.com/silverstripe/silverstripe-framework/issues/3557
but the test confirmed that no further changes are needed.

Notably, this test validates the creation of an extra Versions entry
on publication as designed behaviour, which appears to be the case.

### DIFF
--- a/tests/php/VersionedTest.php
+++ b/tests/php/VersionedTest.php
@@ -1520,4 +1520,72 @@ class VersionedTest extends SapphireTest
         // See: SapphireTest::createMemberWithPermission()
         $this->assertSame('ADMIN@example.org', $publisher->Email);
     }
+
+    /**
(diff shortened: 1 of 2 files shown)


tests/php/VersionedTest.php
```diff
@@ -1520,4 +1520,73 @@ public function testPublisher()
...
+        $versions = $record->VersionsList()->map('Version', 'WasPublished')->toArray();
+        $this->assertEquals(
+            [
+                1 => 0,
+                2 => 0,
+                3 => 1,
+            ],
+            $versions
+        );
+
```
robbieaverill: This empty line is causing the PHPCS failure =)

sminnee: *sigh* force-pushed a squash fix

Issue #230, opened 2019-05-08
SS4.1 Tagging


nglasl:
This is a follow up to https://github.com/silverstripe/silverstripe-versioned/issues/216 because the tags for this module are incorrect and need updating. My project is running SS4.1.5, and has pulled down 1.3.2 of this module (since that only asks for SS4, and 1.3.2 is the latest stable version for SS4). However, 1.3.2 of this module throws the error outlined in the issue above because the class wasn't introduced until SS4.2. The dependencies were corrected in 1.3.3, but ~~technically 1.3.3 should have replaced 1.3.2~~ this is where it gets tricky because the problem goes back to 1.3.0.

nglasl:
Digging into this a little further, 1.2.4 seems fine https://github.com/silverstripe/silverstripe-versioned/blob/1.2.4/src/GridFieldArchiveAction.php vs https://github.com/silverstripe/silverstripe-versioned/blob/1.3.0/src/GridFieldArchiveAction.php. However, the 1.2 branch has the requirement change to SS4.2 https://github.com/silverstripe/silverstripe-versioned/commits/1.2. This is confusing, and now I don't know what version I *should* be using :)

robbieaverill:
Hey @nglasl, we have a version lock with SilverStripe framework if you remove three minor versions from it - e.g. SS 4.3 = versioned 1.3. Are you noticing this error message in a unit test build? I've seen that before and we usually work around it by specifying which versioned constraint to use. It can pull in incompatible versions if you have a `^1.0` constraint on versioned somewhere in your project, because Composer will install the latest available version even if you're running SilverStripe 4.1 or 4.2 for example

nglasl:
Yeah I'm using https://github.com/symbiote/silverstripe-queuedjobs/blob/4.3.2/composer.json which ends up pulling this down as a dependency. The error goes away if I explicitly define `"silverstripe/versioned": "1.1.5"`, but the reason for creating the issue is:

- others are bound to encounter this
- it's not clear on the cause until you start digging through file or issue history.

Considering this affects so many versions, there's probably not a lot we can do unless the module dependencies are more strict. Even then, others will encounter this on all existing versions of said module.

robbieaverill:
Technically speaking, I think adding the class was a semantic versioning violation, since PHP cannot parse the class without the interface it implements existing. We could probably patch this easily enough with an early return:

```php
if (!interface_exists(GridField_ActionMenuItem::class)) {
    return;
}
[...]

PR #232, opened 2019-05-17
FIX Revert framework constraint bump and add easly returns when non B/C interface does not exist

robbieaverill:
This change must be released as a 1.3.x patch tag, then the Composer constraint can be bumped back again to where it was. This will fix cases where Composer installs silverstripe/versioned ~1.3.0 on older SilverStripe 4.0/4.1 projects, which has a backwards compatibility break in it.

The change here allows the code to be backwards compatible with older versions. Once tagged it can be bumped and re-tagged in future in the same release line without concerns for the same B/C break.

Issue: https://github.com/silverstripe/silverstripe-versioned/issues/230

Note that we've already decided it's OK to stop releasing core modules in lock step with each other - see https://github.com/silverstripe/silverstripe-framework/issues/8970

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         }
     ],
     "require": {
-        "silverstripe/framework": "^4.2",
+        "silverstripe/framework": "^4",
         "silverstripe/vendor-plugin": "^1"
     },
     "require-dev": {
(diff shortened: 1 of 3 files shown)


robbieaverill:
@ScopeyNZ this will also the broken restfulserver builds I saw you looking at yesterday 😆

PR #233, opened 2019-05-17
Revert "FIX Revert framework constraint bump and add easly returns when non B/C interface does not exist"

robbieaverill:
This reverts commit 5df7f4f2708efe98daa6c2be8f0d5a2de6b0742d.

The previous commit has been released in 1.3.4, so we can restore the previous state from now on.

Issue: https://github.com/silverstripe/silverstripe-versioned/issues/230

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         }
     ],
     "require": {
-        "silverstripe/framework": "^4",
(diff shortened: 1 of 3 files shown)


robbieaverill:
Self-merging with @ScopeyNZ's approval

Issue #72, opened 2017-12-04
Propagate "draft" status to owner objects


chillu:
# Overview

Authors deal with more versioned, nested relationships in SilverStripe 4. In order to simplify the mental model, many authors will still think about publication around the "main" types: Pages, and optionally Blocks. With the ownership and cascading publish abilities already in place, authors effectively regard the "main" record as changed if any of it's owned relations are changed (have an unpublished draft). The CMS needs to effectively communicate this draft status at a level that makes sense to authors, without forcing them to click through dozens of editing UIs for owned relations to determine individual draft status of these.

# Example

 * `Page` has_many and owns`ContentBlock`
 * `Page` many_many `Watcher`
 * `GalleryContentBlock` has_many and owns `GalleryItem`
 * `GalleryItem` has_one and owns `Image`

Effect:

 * Changing the title of `Image` (without publishing) would mark `GalleryItem`, `GalleryContentBlock` and `Page` as changed.
 * Adding an unpublished `GalleryItem` would mark `GalleryContentBlock` and `Page` as changed
 * Adding and publishing a `GalleryItem` would not mark any records as changed
 * Adding a `Watcher` would not mark any records as changed

# Acceptance Criteria

 * [x] The draft status is only affected by owned relations
 * [x] The draft status is reflected throughout nested owners, unless the ownership chain is "broken"
 * [x] Only unpublished relations change the status of owners
 * [x] If an owner is published, the draft status of all nested owned relations is reset
 * [x] The draft status is triggered when an item in an owned relation is added or removed
 * [x] If a record is owned by more than one owner ("non-exclusive ownership"), all of them show a draft status
 * [x] The draft status is correctly reflected in the site tree (for pages with nested owned relations)
 * [x] The draft status is correctly reflected in edit views for ~~GridField~~ and Page (GridField update is separate: https://github.com/silverstripe/silverstripe-versioned/pull/222)
 * [x] If an owned relation is changed on an edit screen, the change is reflected after the owner is saved (and the view is refreshed)
 * [x] Showing the draft status does not significantly slow down tree views (for Pages), or GridField views (for other records)
 * [x] Loading and saving data on a record (owner and owned) is not significantly slowed down either.

# Notes

 * A `ChangeSet` adds owns relations (recursively) through`ChangeSet->findReferenced()`, regardless of they're changed or not. 
 * Unchanged `ChangeSetItem` records will have `VersionBefore = VersionAfter`
[...]

PR #235, opened 2019-06-04
FIX lookupReverseOwners to work in deriving classes

dnsl48:
The fix allows classes extending RecursivePublishable to use `lookupReverseOwners`.
Currently, those aren't always explicitly registered as extensions and as such `DataObject::has_extension` may return false even if RecursivePublishable is there, but the particular calling extension is not.
`DataObject::has_extension` checks for deriving classes internally (unless a third parameter given).

# Related Issues

https://github.com/silverstripe/silverstripe-versioned/issues/72

### DIFF
--- a/src/RecursivePublishable.php
+++ b/src/RecursivePublishable.php
@@ -244,7 +244,7 @@ class RecursivePublishable extends DataExtension
         array_shift($classes); // skip DataObject
         foreach ($classes as $class) {
             // Ensure this class is RecursivePublishable
-            if (!DataObject::has_extension($class, static::class)) {
+            if (!DataObject::has_extension($class, self::class)) {
                 continue;
             }
 


ScopeyNZ:
@dnsl48 do you want to target this at 1.4 (or earlier)? It's a bugfix?

dnsl48:
Could do so, but I'm not sure it's worth the risk. Even though it's a bufgix, technically it amends the method behaviour. I'd prefer it to go to the next minor release, just as a safer path to take.

ScopeyNZ:
I'm still thinking it's probably a valid bugfix. I really think that if anyone else was subclassing `RecursivePublishable` then they'll have the same issue.

dnsl48:
I've changed the base to `1.4`

PR #238, opened 2019-06-18
API Add deprecation warnings to Versioned::VersionsList() and allVersions()

robbieaverill:
These methods are already removed in 5.x and are in the changelog as removals, we missed a deprecation notice though.


robbieaverill:
Self merging this since it's a trivial docs change.

PR #239, opened 2019-07-23
MINOR Default WasDraft to true

maxime-rainville:
`WasDraft` didn't exist in SilverStripe 3. In most cases it will be set to `true`. Exception are when archiving a record or when publishing directly to live.

History view breaks in most cases when looking at records from SS3, because WasDraft default to false.

By default it to true we provided a better representation of the SilverStripe 3 data and we fix the history view.

Note that this will only fix new migration, not old one.

# Parent issue
* https://github.com/silverstripe/silverstripe-versioned-admin/issues/100

### DIFF
--- a/lang/eo.yml
+++ b/lang/eo.yml
@@ -1,4 +1,10 @@
 eo:
+  SilverStripe\VersionedAdmin\ArchiveAdmin:
+    MENUTITLE: Arkivoj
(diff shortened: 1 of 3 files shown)


maxime-rainville:
I contemplated creating a fake SS3 database and tend running a fake `dev/build` to test that migrated data get a WasDraft flag set to true, but opted against it as it seemed overkilled.

If someone can think of an elegant way of testing this, I'm all hears.

PR #240, opened 2019-07-24
Default Was Draft to true

maxime-rainville:
`WasDraft` didn't exist in SilverStripe 3. In most cases it will be set to `true`. Exception are when archiving a record or when publishing directly to live.

History view breaks in most cases when looking at records from SS3, because WasDraft default to false.

By default it to true we provided a better representation of the SilverStripe 3 data and we fix the history view.

Note that this will only fix new migration, not old one.

# Parent issue
* https://github.com/silverstripe/silverstripe-versioned-admin/issues/100

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,6 +4,8 @@ env:
   global:
     - COMPOSER_ROOT_VERSION=1.3.x-dev
 
+dist: trusty
+
 matrix:
   fast_finish: true
   include:
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -163,7 +163,7 @@ class Versioned extends DataExtension implements TemplateGlobalProvider, Resetta
         "Version" => "Int",
         "WasPublished" => "Boolean",
         "WasDeleted" => "Boolean",
-        "WasDraft" => "Boolean",
+        "WasDraft" => "Boolean(1)",
         "AuthorID" => "Int",
         "PublisherID" => "Int"
     ];


maxime-rainville:
Travis is being obnoxious ... I'm trying to debug it.

maxime-rainville:
Travis is happy again. For reference the problem was that Travis had switch our build to Xenial because we hadn't explicitely told it to stay on trusty https://changelog.travis-ci.com/xenial-as-the-default-build-environment-99476

maxime-rainville:
I contemplated trying to create a test that build a fake SS3 table and tries to upgrade it to see if `WasDraft` gets set correctly, but I concluded it was more trouble than it's worth.

If someone can think of a more simple way of testing this, I'm all hears.

chillu:
@scott-nz Can you check if this fixes your use case please?

scott-nz:
Yes, this resolves the issue for me.
History viewer preview is displaying after downloading a fresh copy of the SS3 version of the database and running a dev/build.

chillu:
I can see how this is fixed when I manually `UPDATE` the `*_versions` tables in 4.x, but not when I just run a dev build. I've created some draft and publish versions on a standard 3.x install to test this.

After importing the 3.x snapshot into 4.x and running a dev/build, I see the newly created column with a default of `0` (not `NULL`).

```sql
CREATE TABLE `SiteTree_Versions` (
-- ...
`WasDraft` tinyint(1) unsigned NOT NULL DEFAULT '0',
) ENGINE=InnoDB AUTO_INCREMENT=18 DEFAULT CHARSET=utf8;
```

As expected, all of the existing versions take on that new default:

```sql
SELECT COUNT(*) FROM SiteTree_Versions WHERE WasDraft = NULL; => 0
SELECT COUNT(*) FROM SiteTree_Versions WHERE WasDraft = 0; => 17
SELECT COUNT(*) FROM SiteTree_Versions WHERE WasDraft = 1; => 0
```

After applying the pull request and running dev/build, the default changes to `1`.

```sql
CREATE TABLE `SiteTree_Versions` (
-- ...
  `WasDraft` tinyint(1) unsigned NOT NULL DEFAULT '1',
) ENGINE=InnoDB AUTO_INCREMENT=18 DEFAULT CHARSET=utf8;
```

None of the existing data takes on the new default, since they already have a value:

```sql
SELECT COUNT(*) FROM SiteTree_Versions WHERE WasDraft = NULL; => 0
SELECT COUNT(*) FROM SiteTree_Versions WHERE WasDraft = 0; => 17
SELECT COUNT(*) FROM SiteTree_Versions WHERE WasDraft = 1; => 0
```

So from what I can tell, we don't have a way to tell apart "versions created under 3.x" from "versions created under 4.x". The [DBBoolean::$defaultVal](https://github.com/silverstripe/silverstripe-framework/blob/4/src/ORM/FieldType/DBBoolean.php#L15) has been in place for ages (and hasn't changed since [3.x](https://github.com/silverstripe/silverstripe-framework/blob/3/model/fieldtypes/Boolean.php#L10)). I'm suspecting that this pull request won't be effective, and we'll need explicit `UPDATE` statements, filtering for versions younger than the date when the 4.x site went live - so a lot more impactful. What am I missing?

maxime-rainville:
As mentioned, this only fixes new migrations.

There's not an elegant way of knowing which version got created on SilverStripe 3 and which version got created on SilverStripe 4. We could write a post-migration task that retroactively set `WasDraft` to true for versions created before a given date, but that will require site owners to provide a date.

maxime-rainville:
@chillu Sorry, this dropped off my radar until someone hit me up about hit on the public slack. Do we still have objections to merging my fix. From the discussion we had off line, it sounded like your concerns had been alleviated.

chillu:
Yeah I think we sorted this out offline, was a misunderstanding on my part. Feel free to self-merge. I haven't merged yet because I didn't know how this would affect your current in flight release

PR #241, opened 2019-08-22
FIX Ensure dataobjects are unpublished during the delete mutation

ScopeyNZ:
Fixes dnadesign/silverstripe-elemental#678

### DIFF
--- a/src/GraphQL/Extensions/DeleteExtension.php
+++ b/src/GraphQL/Extensions/DeleteExtension.php
@@ -1,0 +1,44 @@
+<?php
+
+namespace SilverStripe\Versioned\GraphQL\Extensions;
+
+use Exception;
+use SilverStripe\Core\Extension;
+use SilverStripe\GraphQL\Scaffolding\Scaffolders\CRUD\Delete;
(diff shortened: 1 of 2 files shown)


src/GraphQL/Extensions/DeleteExtension.php
```diff
@@ -0,0 +1,44 @@
...
+ * Extends the @see Delete CRUD scaffolder to unpublish any items first
+ */
+class DeleteExtension extends Extension
+{
+    /**
+     * Hooks into the `augmentMutation` extension point in @see Delete::resolve
+     *
+     * @param DataList $objects
+     * @param $args
+     * @param $context
```
robbieaverill: Can you fix this syntax? It's not valid

unclecheese:
@ScopeyNZ / @robbieaverill Is there a reason why this just doesn't use `doArchive`? Is that because `delete()` has already been called by the owner?

Wondering if we need a new `archiveXXX` operation to be provided by versioned.

robbieaverill:
I believe that if you delete without unpublish the object will still exist in the live stage with no way to remove it

PR #243, opened 2019-08-28
Allow actions to be taken after versioned has recorded a delete

adrhumphreys:
There are instances where you want to do actions after versioned has written the delete version.

An example of this is wanting to take a snapshot of a dataobject after it's been deleted. You'd currently get the wrong/correct version number depending on when the extension runs


tractorcow:
Looks good to me.

robbieaverill:
Thanks @adrhumphreys

PR #245, opened 2019-09-08
Dupe: Allow actions to be taken after versioned has recorded a delete

adrhumphreys:
This is a dupe of https://github.com/silverstripe/silverstripe-versioned/pull/243
I just was hoping I could get this merged into 1.4 and tagged for a release of 1.4.x


ScopeyNZ:
I'm not sure we can classify this as a patch change to be honest. I'll leave this open for other opinions though!

Cheddam:
The original change is now present in two minor releases, so I imagine this is no longer necessary 😅

PR #246, opened 2019-10-08
FIX archive relationships, not related objects

### LINKED ISSUE #200: GridFieldArchiveAction archives the end object in a versioned many_many relation
## Affected Version

silverstripe/recipe-cms 4.3.0
silverstripe/versioned 1.3.0

## Description

In SilverStripe 4.3.0 (at least) when you have a many_many through relationship that is versioned (on both the junction and the end objects) the "Archive" button provided through `GridFieldArchiveAction` does not archive the junction object it archives the end object. As well the junction object's row remains in the database.

As a side (not sure if it's my install or not it's possible found on a very big site) the alert to confirm the action is empty in Firefox 64 at least.

## Expected
The junction object would be removed but the end object of the relationship would not be.

## Steps to Reproduce

1. Have a many_many through relationship that has the junction and end objects versioned
2. Add one of the end objects to the relationship
3. Click the archive button
4. Check the end object and it is now archived

# Pull requests
* [ ] https://github.com/silverstripe/silverstripe-versioned/pull/246

NightJar:
The archive action is set to take the place of the delete button,
considering that when a published item is deleted (i.e. from the
draft table) then it cannot be unpublished as no interface exists
to do so.

However this was having the unintended affect of removing objects
people wished to only remove from a relationship list, as the
archive action did not account for whether the delete action intends
to remove the related object, or the link to it (the relationship).

This is most easily remedied by providing the user with both options.
It may not be the best user experience, but is the most flexible.

Resolves #200

### DIFF
--- a/src/GridFieldArchiveAction.php
+++ b/src/GridFieldArchiveAction.php
@@ -73,7 +73,16 @@ class GridFieldArchiveAction implements GridField_ColumnProvider, GridField_Acti
         $model = $gridField->getModelClass();
         $isModelVersioned = $model::has_extension(Versioned::class);
         if ($isModelVersioned) {
-            $gridField->getConfig()->removeComponentsByType(GridFieldDeleteAction::class);
+            $config = $gridField->getConfig();
+            $deleteComponents = $config->getComponentsByType(GridFieldDeleteAction::class);
+            foreach ($deleteComponents as $deleteComponent) {
+                if ($deleteComponent->getRemoveRelation()) {
+                    // The 'delete' button will "unlink" the relationship, NOT delete the item.
+                    continue;
+                }
+                // Deleting an item that is published would leave no way to unpublish it.
+                $config->removeComponent($deleteComponent);
+            }
         }
         if (!in_array('Actions', $columns)) {
             $columns[] = 'Actions';


NightJar:
:thinking: This only works when the Through object is _not_ versioned.
else we end up with the same problem the altered offending code was originally trying to solve - being unable to delete the published record.

NightJar:
# Summary of options

1. Create a new versioned unlink action grid field component to replace any "unlink" type delete action components, in a similar manner to what used to happen with archive action & delete actions of any type.
2. Modify `ChangeSet::findReferenced` to _always_ include many many through style relationships if the object it represents defines any, these are _always_ implicitly owned, as the relationship affects nothing else. New `ChangeSetItem`s should then be made for this link object of type `ChangeSetItem::CHANGE_DELETED` for all records that have been deleted but are yet to be unpublished.

# Pros

1. Easier to implement. Works like many many style relationships did in SilverStripe 3 (and unversioned ones still do in SilverStripe 4).
2. More complete. Covers more expected use case.

# Cons

1. Counterintuitive to versioning; Records are versioned, but relationship changes are instant. _May_ lead to uncertainty about version history.
2. Will take longer to implement. Can hurt brain in attempts to keep track of the application state.

maxime-rainville:
Maybe we should just include all possible actions as component of the generic grid field config (archive/delete, unlink, unpublish). Then, it's the responsibility of developers to remove what they doen't want based on their specific context.

NightJar:
This is a reasonable idea, and the approach I've taken the quicker path to implementing (show both unlink and archive). However this would require a representative row in the relevant gridfield(s) to continue to display the item, so long as it maintains a link through that relationship in any state (draft/published/deleted).

tractorcow:
I think this makes an improvement; Users who had an unlink action will now be able to use it properly, even if there is an extraneous extra option that they may not use.

NightJar:
@mfendeksilverstripe - you may be interested in this problem set (still outstanding, although this PR is merged) while adjusting https://github.com/silverstripe/silverstripe-framework/pull/8856 :)

PR #248, opened 2019-11-04
NEW migration task to publish unVersioned objects

NightJar:
DataObjects with the Versioned extension may not have always had this
setting applied to them. What was previously "always live" by nature of
not having a "draft" version now are by default in draft, as it is the
"base table" that is used as the draft with the Versioned extension.

We need to migrate all objects that exist in the draft table and have no
recorded version at all to the live table, in order that they may continue
to function as they had before the Versioned extension was applied.

This is a new build task that does exactly that. It searches for any
DataObject in this state, and publishes it - but care is taken to not
cascade that publish to cause issues with any new owns relationship
that may have been set up.

### DIFF
--- a/src/Task/PublishNewlyVersionedDataObjectsTask.php
+++ b/src/Task/PublishNewlyVersionedDataObjectsTask.php
@@ -1,0 +1,183 @@
+<?php
+
+namespace SilverStripe\Versioned\Task;
+
+use Exception;
+use Monolog\Handler\StreamHandler;
+use Monolog\Logger;
+use Psr\Log\LoggerInterface;
+use SilverStripe\Assets\File;
+use SilverStripe\Control\Director;
+use SilverStripe\Control\HTTPRequest;
+use SilverStripe\Core\ClassInfo;
+use SilverStripe\Core\Config\Config;
+use SilverStripe\Core\Injector\Injector;
+use SilverStripe\Dev\BuildTask;
+use SilverStripe\Logging\PreformattedEchoHandler;
+use SilverStripe\ORM\DataObject;
+use SilverStripe\ORM\DataObjectSchema;
+use SilverStripe\Versioned\Versioned;
+
+/**
+ * Migrates files that have been changed from unversioned to versioned by application of the Versioned extension.
+ *
+ * Any existing DataObject created before the Versioned extension is applied was previously "always live", but now is
+ * in the "draft" state, as Versioned uses the 'base table' for draft content. This has the side effect of removing all
+ * previously unversioned data on the website from public view.
+ *
+ * This task fixes this by publishing records with no version information.
+ *
+ * NOTE: This task relies on classes listed within the database to still exist. This means any
+ * `SilverStripe\ORM\DatabaseAdmin.classname_value_remapping` definitions should always be run _first_. This happens as
+ * part of a normal `dev/build` process, but is important to mention here as valuable debugging information. Ensure
+ * that all modules & classes have a valid e.g. `_config/legacy.yml` defined for all DataObjects.
+ */
+class PublishNewlyVersionedDataObjects extends BuildTask
+{
+    /**
+     * Allows classes to be ingored by this automated procedure
+     * e.g. in the case that there is a more complex migration task performed elsewhere.
+     *
+     * @config
+     * @var array
+     */
+    private static $ignore_classes = [
+        File::class, // Handled by the File Migration task in `silverstripe/assets`
+    ];
+
+    /**
+     * Apply extra filters to select what will be published from the unversioned records
+     * Map of:
+     *   PHP class name (fully qualified with namespace) => WHERE clasue (escaped SQL string)
+     *
+     * @config
+     * @var array
+     */
+    private static $extra_filters = [];
+
+    /**
+     * Prefix all logs output with this, for easier location in a log stream (e.g. syslog)
+     *
+     * @config
+     * @var string
+     */
+    private static $log_prefix = __CLASS__ . ': ';
+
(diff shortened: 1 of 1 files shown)


src/Task/PublishNewlyVersionedDataObjectsTask.php
```diff
@@ -0,0 +1,183 @@
...
+     * @return void
+     */
+    public function run($request)
+    {
+        $this->configureLogger();
+        $loggerMessagePrefix = $this->config()->get('log_prefix');
+        $this->logger->info($loggerMessagePrefix . 'Beginning');
+
+        $ignoredClasses = $this->config()->get('ignore_classes');
+        $versionedClasses = array_filter($this->findVersionedClasses(), function ($class) use ($ignoredClasses) {
```
ScopeyNZ: Considering `findVersionedClasses` already does `array_filter` why don't you combine the two filters into one?

src/Task/PublishNewlyVersionedDataObjectsTask.php
```diff
@@ -0,0 +1,183 @@
...
+            return !in_array($class, $ignoredClasses);
+        });
+        $extraFilters = $this->config()->get('extra_filters');
+
+        try {
+            foreach ($versionedClasses as $class) {
+                $table = Injector::inst()->get(DataObjectSchema::class)->baseDataTable($class);
```
ScopeyNZ: `DataObjectSchema` could be defined in the `dependencies` config rather than using Injector in every iteration of the loop.

src/Task/PublishNewlyVersionedDataObjectsTask.php
```diff
@@ -0,0 +1,183 @@
...
+                $versionTable = "${table}_Versions";
+                $unversioned = Versioned::get_by_stage($class, Versioned::DRAFT)
+                    ->leftJoin($versionTable, "\"$versionTable\".\"RecordID\" = \"$table\".\"ID\"")
+                    ->where("\"$versionTable\".\"ID\" IS NULL");
```
ScopeyNZ: This will result in `SELECT [attributes] FROM table LEFT JOIN table_versions USING (ID) WHERE table_versions.ID IS NULL`. I would be interested to see if `SELECT [attributes] FROM table WHERE NOT EXISTS(SELECT * FROM table_versions WHERE table_versions.ID = table.ID)` is more efficient.

src/Task/PublishNewlyVersionedDataObjectsTask.php
```diff
@@ -0,0 +1,183 @@
...
+
+                if (isset($extraFilters[$class])) {
+                    $unversioned = $unversioned->where($extraFilters[$class]);
+                }
+
+                $count = $unversioned->count();
+                if ($count) {
```
ScopeyNZ: Maybe:
```
if (!$count) {
	continue;
}
```
?

src/Task/PublishNewlyVersionedDataObjectsTask.php
ScopeyNZ: It might also be nice to extract this into a `getUnpublishedRecords($class)` private method for readability? Note "unpublished" vs "unversioned".

src/Task/PublishNewlyVersionedDataObjectsTask.php
```diff
@@ -0,0 +1,183 @@
...
+    {
+        return array_filter(
+            ClassInfo::subclassesFor(DataObject::class),
+            function ($class) {
+                return (
+                    DataObject::has_extension($class, Versioned::class)
+                    && in_array(
+                        Versioned::class,
+                        (array) Config::forClass($class)->uninherited('extensions')
+                    )
```
ScopeyNZ: I assume you're checking that the class has `Versioned` defined explicitly on it (and not a parent)? Is this a safe assumption? Can't I create a base class with `Versioned` and then extend DataObjects off it with their own DB tables configured? If not, is the first condition actually required?

ScopeyNZ:
Fancy a test?

src/Task/PublishNewlyVersionedDataObjectsTask.php
```diff
@@ -0,0 +1,183 @@
...
+ * previously unversioned data on the website from public view.
+ *
+ * This task fixes this by publishing records with no version information.
+ *
+ * NOTE: This task relies on classes listed within the database to still exist. This means any
+ * `SilverStripe\ORM\DatabaseAdmin.classname_value_remapping` definitions should always be run _first_. This happens as
+ * part of a normal `dev/build` process, but is important to mention here as valuable debugging information. Ensure
+ * that all modules & classes have a valid e.g. `_config/legacy.yml` defined for all DataObjects.
+ */
+class PublishNewlyVersionedDataObjects extends BuildTask
```
JakubDolba: looking on this in project and I would choose different naming (also for `$segment`) to know by a name this is Migration task related to upgrading SS3 -> SS4

maybe something like 
`UpgradeMigrationPublishNewlyVersionedDataObjectsTask`

JakubDolba: I realised this might be used also in later minor upgrades (by introducing new versioned objects which werent versioned before)
so `PublishNewlyVersionedDataObjects` is probably fine

emteknetnz:
@NightJar This pull request hasn't had any activity for a while.  Looks like there's quite a bit of peer review feedback that hasn't been addressed. Are you going to be doing further work on this PR, or would you prefer to close this PR?

emteknetnz:
It seems like there's going to be no further activity on this pull request, so we've closed it for now. Please open a new pull-request if you want to re-approach this work, or for anything else you think could be fixed or improved.

PR #253, opened 2020-01-28
Added missing table versioning

mfendeksilverstripe:
## Versioned table fix

* `DataQuery::applyRelation` now supports versioned tables
* `ManyManyThroughList` no longer leaks draft content

## Related issues

https://github.com/silverstripe/silverstripe-versioned/issues/254

## Dependencies

https://github.com/silverstripe/silverstripe-framework/pull/9386

## Notes

Split from https://github.com/silverstripe/silverstripe-framework/pull/8856

### DIFF
--- a/_config/versionedextension.yml
+++ b/_config/versionedextension.yml
@@ -14,3 +14,9 @@ SilverStripe\Core\Injector\Injector:
       mode: StagedVersioned
   # Default is alias for .stagedversioned
   SilverStripe\Versioned\Versioned: '%$SilverStripe\Versioned\Versioned.stagedversioned'
+---
+Name: versioned-table
+---
+SilverStripe\ORM\DataQuery:
+  extensions:
+    - SilverStripe\Versioned\VersionedTableDataQueryExtension
--- a/src/VersionedTableDataQueryExtension.php
+++ b/src/VersionedTableDataQueryExtension.php
@@ -1,0 +1,35 @@
+<?php
+
+namespace SilverStripe\Versioned;
+
+use SilverStripe\Core\Extensible;
+use SilverStripe\Core\Extension;
+use SilverStripe\Core\Injector\Injector;
+use SilverStripe\ORM\DataQuery;
+
+/**
+ * Applies correct stage to tables
+ *
+ * @property DataQuery $owner
+ */
+class VersionedTableDataQueryExtension extends Extension
+{
+    /**
+     * Extension point in @see DataQuery::getJoinTableName()
+     *
+     * @param string $class
+     * @param string $table
+     * @param string $updated
+     */
+    public function updateJoinTableName($class, $table, &$updated)
+    {
+        if (!Extensible::has_extension($class, Versioned::class)) {
+            return;
+        }
+
+        /** @var Versioned $versioned */
+        $versioned = Injector::inst()->get(Versioned::class);
+        $stage = $versioned->get_stage();
+        $updated = $versioned->stageTable($table, $stage);
+    }
+}
--- a/tests/php/VersionedTableTest/House.php
+++ b/tests/php/VersionedTableTest/House.php
@@ -1,0 +1,66 @@
+<?php
+
+namespace SilverStripe\Versioned\Tests\VersionedTableTest;
+
+use SilverStripe\Dev\TestOnly;
+use SilverStripe\ORM\DataObject;
+use SilverStripe\ORM\HasManyList;
+use SilverStripe\ORM\ManyManyThroughList;
+use SilverStripe\Versioned\Versioned;
+
+/**
+ * Class House
+ *
+ * @property string $Title
+ * @property int $RoofID
+ * @method WoodenRoof Roof()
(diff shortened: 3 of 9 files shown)


emteknetnz:
Is the travis failure on this related to the 9386 framework PR being unmerged?

mfendeksilverstripe:
@emteknetnz Correct, this PR depends on the other, hence I listed it in the dependencies. Note that this change was originally one single change-set (https://github.com/silverstripe/silverstripe-framework/pull/8856) but I was advised to split it apart across two modules.

emteknetnz:
@mfendeksilverstripe are you able to rebase this off the tip of 1.5 and target 1.5 instead?

_config/versionedextension.yml
```diff
@@ -14,3 +14,9 @@ SilverStripe\Core\Injector\Injector:
       mode: StagedVersioned
   # Default is alias for .stagedversioned
   SilverStripe\Versioned\Versioned: '%$SilverStripe\Versioned\Versioned.stagedversioned'
+---
+Name: versioned-table
+---
+SilverStripe\ORM\DataQuery:
+  extensions:
+    - SilverStripe\Versioned\VersionedTableExtension
```
emteknetnz: Seems like this should be renamed to something along the lines of `VersionedDataQueryExtension`

src/VersionedTableExtension.php
```diff
@@ -0,0 +1,33 @@
...
+ * @property DataQuery $owner
+ */
+class VersionedTableExtension extends Extension
+{
+    /**
+     * @param string $class
+     * @param string $table
+     * @param string $augmented
+     */
+    public function augmentJoinTable($class, $table, &$augmented)
```
emteknetnz: I'm not sure if "augment" is quite the right name for this, according the the interweb the definition for it is `make (something) greater by adding to it; increase.`

What this function is doing is getting the correct stage table name, rather than adding to the existing one

Maybe rename this to `updateJoinTableName` ?

Note to other peer reviewers: this function is called from here in a separate new PR
https://github.com/silverstripe/silverstripe-framework/pull/9386/files#diff-05310065482eac07cac9003bc30b7cacR1107

Could this be changed togetStageTableName'

emteknetnz:
@mfendeksilverstripe could you rebase this on top of the `1` branch and target that branch instead of `1.4` - this is for semver reasons since it's adding new API due to the addition of a new class.  This means that this enhancement will be released in `1.6`.

One this is merged you should be able to include this without using a fork in on your current project by setting your composer.json along the lines of
"silverstripe/versioned": "1.x-dev#[hash-of-merge-commit] as 1.4"

mfendeksilverstripe:
@emteknetnz Thanks for feedback, it's now rebased on `1`

emteknetnz:
The travis failure above is because we need to merge https://github.com/silverstripe/silverstripe-framework/pull/9386 first.  The tests pass locally when that framework code is installed i.e. `vendor/bin/phpunit vendor/silverstripe/versioned/tests`

composer.json
```diff
@@ -19,7 +19,7 @@
         }
     ],
     "require": {
-        "silverstripe/framework": "^4.2",
+        "silverstripe/framework": "4.4.x-dev as 4.4.4",
```
mfendeksilverstripe: This is just a temporary change to provide the needed dependency. This should be updated after the dependency is properly released.

.travis.yml
```diff
@@ -30,7 +30,7 @@ before_script:
 # Install composer
   - composer validate
   - composer install --prefer-dist
-  - composer require silverstripe/recipe-core:4.x-dev silverstripe/admin:1.x-dev silverstripe/graphql:3.x-dev --prefer-dist --no-update
+  - composer require silverstripe/recipe-core:4.4.x-dev silverstripe/admin:1.4.x-dev silverstripe/graphql:3.x-dev --prefer-dist --no-update
```
mfendeksilverstripe: This is just a temporary change to get the needed dependency. This should be reverted after the dependency is properly released.

mfendeksilverstripe:
@emteknetnz The tests are now passing as I temporarily updated the test build to include the needed dependency.

.travis.yml
emteknetnz: ok great ta.  could you please revert it back, then squash and push.  we should be good to merge after that

mfendeksilverstripe: Done :)

chillu:
This appears to have caused a critical regression: https://github.com/silverstripe/silverstripe-versioned/issues/283

mfendeksilverstripe:
Why wasn't the regression picked up by unit tests?

maxime-rainville:
It went through our entire regression test suite undetected as well. I think the problem is that this issue only occurs when you have a query joining in a stageless versioned table on a "Live" context.

That's not a very common scenario on any of our supported modules. However, it's probably relatively common in a lot of third party projects.

I added an extra unit tests to the patch to fix the issue, so next time we break this we'll hopefully pick it up.

mfendeksilverstripe:
Thanks, much appreciated.

PR #256, opened 2020-02-10
FIX Update core requirement to 4.5 series

Cheddam:
See [silverstripe-cms#2521](https://github.com/silverstripe/silverstripe-cms/issues/2521) for context.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         }
     ],
     "require": {
-        "silverstripe/framework": "^4.2",
+        "silverstripe/framework": "^4.5",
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -19,7 +19,7 @@
         }
     ],
     "require": {
-        "silverstripe/framework": "^4.2",
+        "silverstripe/framework": "^4.5",
```
emteknetnz:
```suggestion
        "php": ">=7.1",
        "silverstripe/framework": "^4.5",
```

PR #267, opened 2020-03-26
Upgrade to phpunit 8

dhensby:
Bump dependency to phpunit 8


robbieaverill:
Merging because this is required to get framework updated, which is required to get this green. We can make follow up PRs if required.

PR #268, opened 2020-03-29
Update tests for phpunit8

dhensby:
Depends on https://github.com/silverstripe/silverstripe-framework/pull/9443

### DIFF
--- a/tests/php/VersionedGridFieldItemRequestTest.php
+++ b/tests/php/VersionedGridFieldItemRequestTest.php
@@ -44,8 +44,8 @@ class VersionedGridFieldItemRequestTest extends SapphireTest
         $actions = $form->Actions();
 
         // No publish action
-        $this->assertNull($actions->fieldByName('action_doPublish'));
-        $this->assertInstanceOf(FormAction::class, $actions->fieldByName('action_doSave'));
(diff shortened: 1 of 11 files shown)


tests/php/VersionedGridFieldItemRequestTest.php
```diff
@@ -44,8 +44,8 @@ public function testActionsUnversionedOwner()
         $actions = $form->Actions();
 
         // No publish action
-        $this->assertNull($actions->fieldByName('action_doPublish'));
-        $this->assertInstanceOf(FormAction::class, $actions->fieldByName('action_doSave'));
+        $this->assertNull($actions->fieldByName('MajorActions.action_doPublish'));
+        $this->assertInstanceOf(FormAction::class, $actions->fieldByName('MajorActions.action_doSave'));
```
robbieaverill: Interesting - is this something that changed elsewhere and needed to be updated here?

dhensby: I'm not sure where this has changed, but the `Actions` is returning a `FieldList` with a single `CompositeField`, so this assertion was failing without this change

kinglozzer: Looks like this was changed in https://github.com/silverstripe/silverstripe-framework/pull/9200

PR #272, opened 2020-04-25
Move VersionedGridFieldState into correct folder

dhensby:
Composer 2 forward compatibility fix


dhensby:
> Not sure what’s going on with the PGSQL run

It's the CS test which seems to be running `composer validate` but after travis has munged-up the composer.json file. It's also redundant because `composer validate` is run in the install step - I think we should remove it.

> I can’t seem to restart builds in Travis anymore either :(

That happened to me, I had to disconnect my github account from travis and then re-auth it and I could agian.

PR #276, opened 2020-06-18
BUG: Write to stage extension point params.

mfendeksilverstripe:
# Write to stage extension point params.

* extension point uses undefined variable
* this doesn't throw only because it's passed as a reference
* in current situation this is effectively the same as always passing `null`

## Short term workaround

* following line of code can be used as a short term workaround until this fix comes through
* put following line into your extension code to determine the stage

```
$stage = Versioned::get_stage()
```

## Related issues

https://github.com/silverstripe/silverstripe-versioned/issues/277

### DIFF
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -2619,11 +2619,11 @@ SQL
                 $owner->setSourceQueryParam('Versioned.stage', $stage);
 
                 // Write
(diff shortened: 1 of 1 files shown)


robbieaverill:
Thanks @mfendeksilverstripe

PR #279, opened 2020-07-03
Show the type of save that was performed for the record instead of just showing either Saved/Published.

bergice:
See https://github.com/silverstripe/silverstripe-cms/issues/2448

* Save fields to archived version entries (previously these entries would save without any fields on them)
* Show archived/deleted version entries instead of filtering them out

### DIFF
--- a/src/GraphQL/Resolvers/ApplyVersionFilters.php
+++ b/src/GraphQL/Resolvers/ApplyVersionFilters.php
@@ -30,6 +30,7 @@ class ApplyVersionFilters
             case Versioned::LIVE:
             case Versioned::DRAFT:
             case 'latest_versions':
+            case 'all_versions':
                 Versioned::set_stage($mode);
                 break;
             case 'archive':
@@ -77,6 +78,9 @@ class ApplyVersionFilters
                 $list = $list
                     ->setDataQueryParam('Versioned.mode', 'archive')
                     ->setDataQueryParam('Versioned.date', $date);
+                break;
+            case 'all_versions':
+                $list = $list->setDataQueryParam('Versioned.mode', 'all_versions');
                 break;
             case 'latest_versions':
                 $list = $list->setDataQueryParam('Versioned.mode', 'latest_versions');
@@ -182,6 +186,8 @@ class ApplyVersionFilters
                         $date
                     ));
                 }
+                break;
+            case 'all_versions':
                 break;
             case 'latest_versions':
                 break;
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -576,8 +576,9 @@ class Versioned extends DataExtension implements TemplateGlobalProvider, Resetta
      * Augment SQL to select from `_Versions` table instead.
      *
      * @param SQLSelect $query
+     * @param bool $filterDeleted Whether to exclude deleted entries or not
      */
-    protected function augmentSQLVersioned(SQLSelect $query)
+    protected function augmentSQLVersioned(SQLSelect $query, bool $filterDeleted = true)
     {
         $baseTable = $this->baseTable();
         foreach ($query->getFrom() as $alias => $join) {
@@ -615,6 +616,11 @@ class Versioned extends DataExtension implements TemplateGlobalProvider, Resetta
             "count(DISTINCT \"{$baseTable}_Versions\".\"RecordID\")",
             "count(DISTINCT \"{$baseTable}_Versions\".\"ID\")"
         );
+
+        // Filter deleted versions, which are all unqueryable
+        if ($filterDeleted) {
+            $query->addWhere(["\"{$baseTable}_Versions\".\"WasDeleted\"" => 0]);
+        }
     }
 
     /**
@@ -807,6 +813,8 @@ SQL
         $baseTable = $this->baseTable();
         $subSelect = $this->prepareMaxVersionSubSelect($query, $dataQuery);
 
+        $subSelect->addWhere("\"{$baseTable}_Versions_Latest\".\"WasDeleted\" = 0");
+
         if ($this->shouldApplySubSelectAsCondition($query)) {
             $subSelect->addWhere(
                 "\"{$baseTable}_Versions_Latest\".\"RecordID\" = \"{$baseTable}_Versions\".\"RecordID\""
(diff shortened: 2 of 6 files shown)


bergice:
> Tested locally, seems to damage the archive process for Pages. Steps to reproduce:
> 
> 1. Create Page
> 2. Archive Page
> 3. Observe:
>    
>    * Page data is wiped out (Title, Content, etc.)
>    * CMS immediately redirects to the Home page editor, archived Page cannot be loaded in editor unless you navigate to the URL directly

I've made some changes to:
* No longer filter out deleted versions when retrieving record versions (I don't see why we should hide data when reading versions. This may seem like unexpected behaviour.)
* No longer save deleted versions with blank fields (I'm not sure why we are doing this, but to me it almost seems like a bug.)

There may be some unintended consequences with making these changes, so you should probably do a thorough review.

The history page still won't load but I'm pretty sure this is an unrelated bug.

src/Versioned.php
```diff
@@ -1902,7 +1899,7 @@ public function doUnpublish()
         return true;
     }
 
-    public function onAfterDelete()
+    public function onBeforeDelete()
```
emteknetnz, line 1902: Cannot delete `onAfterDelete()`, would be a breaking change requiring a major version release

src/Versioned.php
```diff
@@ -1856,13 +1851,15 @@ public function doArchive()
...
         $this->suppressDeletedVersion(function () use ($owner) {
             $owner->doUnpublish();
-            $owner->deleteFromStage(static::DRAFT);
         });
         // Create deleted version in both stages
         $this->createDeletedVersion([
             static::LIVE,
             static::DRAFT,
         ]);
+        $this->suppressDeletedVersion(function () use ($owner) {
```
emteknetnz, line 1860: How come this is moving?  Was the 'WasDeleted' Version on the Versions table not being created previously?

src/Versioned.php
```diff
@@ -615,9 +615,6 @@ protected function augmentSQLVersioned(SQLSelect $query)
             "count(DISTINCT \"{$baseTable}_Versions\".\"RecordID\")",
             "count(DISTINCT \"{$baseTable}_Versions\".\"ID\")"
         );
-
-        // Filter deleted versions, which are all unqueryable
-        $query->addWhere(["\"{$baseTable}_Versions\".\"WasDeleted\"" => 0]);
```
emteknetnz, line 620: This seems like a pretty massive change that's likely to have side-effects

src/Versioned.php
bergice: It is necessary for the archived entry to be able to query against the current record. If we delete the record first, we won't be able to fetch the fields for the archived one. 🤷

src/Versioned.php
bergice: Turns out this change is not necessary for this to work, so I'll revert it.

bergice:
@emteknetnz I've exposed the `all_versions` query mode through GraphQL and I'll use that instead of `latest_versions`. It seems to interfere less with the current structure.

emteknetnz:
There's some fails in kitchen-sink on assets which related to another thing that was recently merged https://travis-ci.org/github/creative-commoners/cwp-recipe-kitchen-sink/jobs/717066474

Could you rebase silverstripe/versioned off the latest `1` and repush and ping me, cheers

src/Versioned.php
```diff
@@ -1153,18 +1156,16 @@ protected function augmentWriteVersioned(&$manipulation, $class, $table, $record
         ];
 
         // Add any extra, unchanged fields to the version record.
-        if (!$isDelete) {
             $data = DB::prepared_query("SELECT * FROM \"{$table}\" WHERE \"ID\" = ?", [$recordID])->record();
```
emteknetnz, line 1159: This line would need to be unindented

src/Versioned.php
```diff
@@ -1153,18 +1156,16 @@ protected function augmentWriteVersioned(&$manipulation, $class, $table, $record
         ];
 
         // Add any extra, unchanged fields to the version record.
-        if (!$isDelete) {
```
emteknetnz, line 1156: Why are we removing this line? Based on the inline comment inside the block `// If the value is not set at all in the manipulation currently, use the existing value from the database` Was something being written to the _Versions table incorrectly that is needed later?

tests/php/VersionedTest.php
```diff
@@ -1589,6 +1589,27 @@ public function testWriteWithoutVersion()
...
+        $record = new VersionedTest\TestObject();
+        $record->Title = 'My record';
+        $record->write();
+
+        // then archive it
+        $record->doArchive();
+
+        // ensure we can retrieve the archived entry when retrieving versioned entries
+        $versions = $record->VersionsList()->toArray();
+        $this->assertCount(2, $versions, 'All versions are retrieved');
```
emteknetnz, line 1604: This unit test should be a bit more specific?  It should test that one of the version is the draft version, another or the live version?

tests/php/VersionedTest.php
```diff
@@ -1589,6 +1589,27 @@ public function testWriteWithoutVersion()
         );
     }
 
+    public function testArchivedEntries()
+    {
+        // create a new record and save it
+        $record = new VersionedTest\TestObject();
+        $record->Title = 'My record';
+        $record->write();
```
emteknetnz, line 1597: Should this be a publish instead of just a write()?

src/GraphQL/Resolvers/ApplyVersionFilters.php
```diff
@@ -30,6 +30,7 @@ public function applyToReadingState($versioningArgs)
             case Versioned::LIVE:
             case Versioned::DRAFT:
             case 'latest_versions':
+            case 'all_versions':
```
emteknetnz, line 33: Need to add a corresponding unit test in ApplyVersionFiltersTest.php

src/Versioned.php
emteknetnz, line 1159: This line would need to be unindented if we stick with removing the if(!$isDeleted) { line above

tests/php/VersionedTest.php
robbieaverill: Hey team, by the way the message here is shown when the assertion fails, so it's usually a good idea to write a message that explains why it failed eg. incorrect number of versions returned

src/Versioned.php
bergice: It's necessary for populating the archived entry fields correctly. Otherwise they all end up blank..

emteknetnz:
@bergice I've tested this an it looks good, though I'm getting this error when I visit /admin/archive - do you know if this is related?

image

PR #280, opened 2020-07-06
FIX: Fix semantics of isPublished(), isArchived(), isOnDraft() for objects without draft/published stages

sminnee:
The existing code did nothing useful - returning true all the time for
these cases. This patch fixes them be a more meaningful semantic, where
draft and published are treated as being the same stage.

isOnLiveOnly() will never be true in such cases, and isPublished() will
always be the same as isOnDraft().

To do:

* [x] add tests for single-stage case

### DIFF
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -2810,6 +2810,7 @@ SQL
 
     /**
      * Check if this record exists on live
+     * On objects with only 1 stage, check if the record exists on that stage.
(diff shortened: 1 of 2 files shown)


sminnee:
I guess I should add some tests for the single-stage case? 🤔

sminnee:
OK I've added tests now.

PR #286, opened 2020-07-23
FIX Fetch live record during unpublish to handle modified state

emteknetnz:
Fixes https://github.com/silverstripe/silverstripe-assets/issues/371

Unit test PR is in assets - https://github.com/silverstripe/silverstripe-assets/pull/412

### DIFF
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -1433,6 +1433,24 @@ SQL
     }
 
     /**
+     * Helper method to safely suppress delete callback on a separate DataObject
+     *
+     * @param DataObject $dataObject
+     * @param callable $callback
+     * @return mixed Result of $callback()
+     */
+    private function suppressDeletedVersionOnDataObject($dataObject, $callback)
+    {
+        $original = $dataObject->getField(self::DELETE_WRITES_VERSION_DISABLED);
+        try {
+            $dataObject->setField(self::DELETE_WRITES_VERSION_DISABLED, true);
+            return $callback();
+        } finally {
+            $dataObject->setField(self::DELETE_WRITES_VERSION_DISABLED, $original);
+        }
+    }
+
+    /**
      * If a write was skipped, then we need to ensure that we don't leave a
      * migrateVersion() value lying around for the next write.
      */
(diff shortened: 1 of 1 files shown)


emteknetnz:
@maxime-rainville I’m fairly confident it solve the original issue.  I’ve implemented the solution on `Versioned::doArchive()` as I believe that’s what the Delete button calls in asset-admin

Possibly could make the argument that it could be on `File::delete()`, though I'm not personally sure if you're supposed to be able to delete files on different stages independently - I'm assuming you are?

I'd also be much more concerned with unforseen regressions with this on `File::delete()` since that seems as though it would have far more wide ranging impacts

I have not tested this on `File::delete()` to see what would happen

emteknetnz:
I plugged this PR into the cwp-recipie-kitchen-sink travis build, it went green so we should be fairly confident
https://travis-ci.org/github/creative-commoners/cwp-recipe-kitchen-sink/builds/711374504

This is the composer.json file https://github.com/creative-commoners/cwp-recipe-kitchen-sink/blob/pulls/2.5/modified-delete-physical-files/composer.json

emteknetnz:
The reason I went with Versioned was because the wrong File->Hash was going to AssetControlExtension. 

If we were to not go with the Versioned route because of risk (understandable), and wanted to localise the solution more, then we'd probably need to do something within AssetControlExtension such as repeat my solution there i.e test if stagesDifferent, if the record is live, etc.  It would probably be a little messy there given that AssetControlExtension seems pretty abstract.

I've run this branch through travis cwp-recipe-kitchen-sink and it's green, so we should be fairly confident there aren't any glaring side-effects

https://travis-ci.org/github/creative-commoners/cwp-recipe-kitchen-sink/builds/711401917

https://github.com/creative-commoners/cwp-recipe-kitchen-sink/blob/pulls/2.5/modified-delete-physical-files/composer.json

Another benefit with this solution is that is should prevent other future issues where something needs to get a field which isn't just ID in modified DataObjects

emteknetnz:
@maxime-rainville I'm starting to remember why I fixed this in versioned rather than assets

This is the call-stack from `Versioned::onArchive()` to `AssetControlExtension::onAfterDelete()`

image

If we were to fix this in AssetControlExtension rather than Versioned, then we'd need to query the File_Live record in either `AssetControlExtension::onAfterDelete` (logical place), or in `File->delete()/onAfterDelete()` (doesn't really make sense given that's what AssetControlExtension is for)

However, because this in `onAfterDelete()` it means the File_Live database record no longer exists, so we cannot get the relevant File->Hash

Therefore the logical place to put this would be `AssetControlExtension::onBeforeDelete()` which does not yet exist.  Since this in new API, we would need to release this in 1.7 which is several months away, and we could not backport this to 1.5 or 1.6

I'm happy to move this to redo this in `silverstirpe-assets:1` `AssetControlExtension::onBeforeDelete()` if you want, it just means it will be much slower getting delivered to the real world.  Let me know how you want to proceed.

maxime-rainville:
I would rather keep the hook on `onAfterDelete` because if there's something that prevents the deletion from taking place we'll end up deleting the file. You should be able to look at the archived version to see what the latest live versioned was.

Maybe add some timestamp check, so that if the live version was deleted some time ago, you don't preform the delete.

There could be a case for adding a `onAfterLiveDelete()` hook for this purpose. But I don't think we need this to fix this issue.

emteknetnz:
@maxime-rainville I made an attempt at using the File_Versions table https://github.com/silverstripe/silverstripe-assets/pull/416/files

However I'm not sure how to create File DataObject that represents the File_Live table from the File_Versions table

I've left inline comments on the linked PR

src/Versioned.php
```diff
@@ -1853,11 +1871,31 @@ public function doArchive()
...
+        $id = $this->owner->ID;
+
+        // Unpublish (including delete) from live stage, without creating deleted record on version table
+        // Explicitly use the live record instead of just using $owner which may be the draft record
+        // to handle cases where the record is in a modified state and an extension somewhere needs to
+        // use a value specifically from the live record, for example \Silverstripe\Assets\File::Hash
+        $liveRecord = $isLiveVersion ? $owner : static::get_by_stage($class, Versioned::LIVE)->byID($id);
+        if ($liveRecord) {
+            $this->suppressDeletedVersionOnDataObject($liveRecord, function () use ($liveRecord) {
+                $liveRecord->doUnpublish();
```
emteknetnz, line 1886: Max did some testing to replicate issue, he noted that unpublishing modified files has a similar issue.  This implies that if we are to fix this in versioned, then the fix needs to be in `doUnpublish()`, not `doArchive()`

I originally had it in `doUnpublish()`, though it failed a unit test to do with the count of the Versions table, hence I moved the fix to `doArchive()`

Probably need to keep the fix in `doArchive()` so that Versions count stays correct, though have an additional check within `doUnpublish()` if the current stage is LIVE that we are correctly using the Live record so that we get the correct File->Hash

emteknetnz:
Merged this PR instead https://github.com/silverstripe/silverstripe-versioned/pull/291

PR #291, opened 2020-08-04
FIX Re-fetch DataObject when unpublishing to handle modified state

emteknetnz:
Fix for https://github.com/silverstripe/silverstripe-assets/issues/371

Co-authored-by: Maxime Rainville <maxime@silverstripe.com>

### DIFF
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -1876,6 +1876,7 @@ SQL
      */
     public function doUnpublish()
     {
+        /** @var DataObject|Versioned $owner */
         $owner = $this->owner;
         // Skip if this record isn't saved
         if (!$owner->isInDB()) {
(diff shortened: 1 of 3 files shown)


emteknetnz:
Which `onAfterDelete` are you refering to?

If it's the one in Versioned, that doesn't specificy a $record

If it's the one in Assets, that should already be covered by the other PR https://github.com/silverstripe/silverstripe-assets/pull/412/files#diff-83d5251bdbd398189e341aed1abec995R1037

maxime-rainville:
The point of this change is to make sure that the actual delete happens on the LIVE version so that all the hooks get the live data.

The way I would approach this is create a dummy data object that records hook call in a static variable so we can recover the value after the delete.

I guess we could live with testing this functionality in asset, but it's not ideal, because you could break something in versioned and you wouldn't know until the asset unit tests are run against your merged versioned PR.

emteknetnz:
@maxime-rainville Have add unit-test as required.

Have run this PR along with the assets PR through [travis kitchen-sink](https://travis-ci.org/github/creative-commoners/cwp-recipe-kitchen-sink/builds/715358218), is green

Issue #289, opened 2020-08-02
Version timing


mfendeksilverstripe:
# Version timing

This is an issue which mostly affects larger projects with lot of nested data models. When calling `publishRecursive()` on a top level object like page we assume that the core part of the operation (writes) finish within the same second. This may not be the case however for large projects - publish operation may take several seconds. This results in inconsistent version timestamps which may break the rollback feature.

## Example

Page version timestamp is `2020-08-03 10:51:01`
Block version timestamp is `2020-08-03 10:51:03`

Reverting or viewing version for  `2020-08-03 10:51:01` timestamp will show page version which is correct but the block version will be outdated.

One way to test this is to add `sleep(2)` to `onBeforeWrite()` for both the page and the block.

## Solution

Make sure all write use `$now = DBDatetime::now()->Rfc2822();` and use the `DBDatetime::set_mock_now()` to set the timestamp to match the page (as this happens first). All nested objects need to inherit the timestamp from the page when running `publishRecursive()`.

Fixed in https://github.com/silverstripe/silverstripe-versioned/pull/292

maxime-rainville:
I'm marking this has impact/high because it could lead to data inconsistencies that can not easily be remedied.

maxime-rainville:
All related PR have been merged

Issue #195, opened 2018-11-30
Versions aren't created when owned relations are updated


ScopeyNZ:
## Description

When writing a relation that is owned by a versioned DataObject, a new version is _not_ created. There's a few ways this can be observed and I've written a test that shows this (#196).

## Steps to reproduce

- Have a versioned dataobject that "owns" some relation
- Update the relation (and write)
- Note that the parent dataobject has no new version

## Acceptance Criteria

 - [x] Allows to list all changes between two versions of an owner (incl. direct and indirect owned relations)
 - [x] Allows to list unpublished changes relative to current owner version (incl. direct and indirect owned relations)
 - [x] Allows to determine if an owner is modified if any (direct or indirect) owned relations are modified
 - [x] Allows to determine if intermediate owners are modified
 - [x] Does not mark siblings and unrelated graph nodes as modified
 - [x] Allows to determine modification status after changes have been rolled back
 - [x] Does not mark owners as modified if an owned relation has been archived
 - [x] Does mark owners as modified if an owned relation has been unpublished
 - [x] Each entry uniquely points to a version of the original record
 - [x] Groups snapshot items for each original action which triggered the snapshot
 - [x] Tracks author and date on each group
 - [x] Supports deep nested ownership graphs (5+ levels)
 - [x] Supports propagating changes to multiple owners
 - [x] Supports all owner relationships: `has_one`, `has_many`, ~~many_many~~, `many_many_through`
 - [x] Supports "intermediate" owners which aren't versioned (in order to reach actual owners which are versions)
 - [x] Supports moving of owned relations to different owners ~~without falsifying past "snapshots"~~
 - [x] Does not significantly slow down save operations
 - [x] Does not rely on creating new versions on owners
 - [x] Is resilient to infinite loops on potentially recursive owner structures
 - [x] Is resilient to partial data sets (can't retroactively create this data)
 - [x] Tracks changes in subsites and fluent languages (and can filter by them)
 - [ ] It's obvious to developers reading up about versioned that this option exists (as an opt-in for now)

## Subtasks

* Get buy-in on overall implementation from one or more bespoke teams
* Test on a project with a large number of DataObjects (> 50) and dense version histories (> 100,000 rows)
* Resolve rollback bug (cannot currently roll back owned changes that are strictly draft, i.e. no publish event in between)
* Investigate viabilty of overriding versioned-admin versus augmenting it (e.g. with injectable graphql queries)
* Implement snapshot-admin graphql/react UI
[...]

PR #292, opened 2020-08-04
BUG: Consistent timestamp fix for versions during publish recursive.

mfendeksilverstripe:
# Consistent timestamp fix for versions during publish recursive

## Problem

Calling` publishRecursive()` on a top level object like `Page` can produce inconsistent timestamps on version of nested objects. This breaks the version viewer and rollback functionality as they rely on correct timestamps of the versions.

## Example

* Page has 10 blocks
* Content author publishes the page which triggers recursive publish
* this operation creates published version of all related object starting from page going down to blocks
* let's assumes that operation started and ended with different timestamps (this may happen if the operations are slow or there are simply many objects to go through)
* the resulting versions of related objects will have different timestamps
* this can be emulated by adding `sleep()` to `onBeforeWrite()` calls

### Timestamps

Page - `2020-01-01 10:00:00`
Block 1 - `2020-01-01 10:00:00`
Block 2 - `2020-01-01 10:00:00`
Block 3 - `2020-01-01 10:00:00`
Block 4 - `2020-01-01 10:00:00`
Block 5 - `2020-01-01 10:00:00`
Block 6 - `2020-01-01 10:00:01`
Block 7 - `2020-01-01 10:00:01`
Block 8 - `2020-01-01 10:00:01`
Block 9 - `2020-01-01 10:00:02`
Block 10 - `2020-01-01 10:00:02`

### Why is this a problem

Version history viewer and rollback functionality rely on timestamps to retrieve the correct version. In this specific example the Page dictates the timestamps of the versions. This doesn't work however, because page was the first object to have the version written, so the timestamp used will exclude the other nested objects.

When viewing page version `2020-01-01 10:00:00` the blocks 1 through 5 will show correct content but blocks 6 through 10 will show outdated content as their version was written later.

Same applies to rollback.

## Solution

Lock the time at the start of the operation for the duration of the whole operation.

## Related issues

https://github.com/silverstripe/silverstripe-versioned/issues/289

## Notes

The diff is not very readable but there are no changes other than the callback wrap. You can view the [real diff here](https://github.com/silverstripe/silverstripe-versioned/pull/292/files?w=1).

## Dependencies

https://github.com/silverstripe/silverstripe-framework/pull/9630


ScopeyNZ:
> The diff is not very readable but there are no changes other than the callback wrap.

Putting `&w=1` on the end of the PR URL makes it much more readable 😅

mfendeksilverstripe:
This is amazing, thanks @ScopeyNZ

michalkleiner:
It's the equivalent of [Hide whitespace changes](https://user-images.githubusercontent.com/233342/89355477-94412500-d70f-11ea-8aab-2ab3ab6542f1.png) setting. TIL.

ScopeyNZ:
Ah yeah, the `?w=1` was around for a while before they added UI for it.

Back to this PR

I'm a little surprised this is actually relevant for rollbacks. Yes it does use timestamps but, although it's been a while since I looked at this code in-depth, I remember that it was relying on a timestamp on a record that was saved after related records. 

Maybe this is a new bug that came out of trying to prevent "unnecessary" saves? I vaguely remember seeing something like this in core.

src/RecursivePublishable.php
```diff
@@ -57,41 +57,45 @@ class RecursivePublishable extends DataExtension
...
-        $owner = $this->owner;
+        $now = DBDatetime::now()->Rfc2822();
 
-        // get the last published version
-        $original = null;
-        if ($owner->hasExtension(Versioned::class) && $owner->isPublished()) {
-            $original = Versioned::get_by_stage($owner->baseClass(), Versioned::LIVE)
-                ->byID($owner->ID);
-        }
+        return DBDatetime::withMockNow($now, function () {
```
maxime-rainville, line 62: I don't think this method exists.

tractorcow: You could always create a `::withFixedNow()` method that automatically takes the current now(), and sets it as fixed for the duration. It'd look a lot more professional and less "should be a test-only method" :P

ScopeyNZ: See https://github.com/silverstripe/silverstripe-framework/pull/9630

ScopeyNZ:
I tested this pretty in depth a couple of years ago, and was only let down by the versions being created in the first place (see https://github.com/silverstripe/silverstripe-versioned/issues/195)

mfendeksilverstripe:
@ScopeyNZ I reviewed the issue which you linked but I don't think it's related to my change set. The missing versions can be provided by using the [Versioned Snapshots](https://github.com/silverstripe/silverstripe-versioned-snapshots) module which lets you gain control of when versions are created.

mfendeksilverstripe:
Pushed new changes - added unit tests & PR fixes.

src/RecursivePublishable.php
maxime-rainville: Duh ... just ignored me.

src/RecursivePublishable.php
```diff
@@ -57,41 +57,45 @@ class RecursivePublishable extends DataExtension
...
-        $owner = $this->owner;
+        $now = DBDatetime::now()->Rfc2822();
 
-        // get the last published version
-        $original = null;
-        if ($owner->hasExtension(Versioned::class) && $owner->isPublished()) {
-            $original = Versioned::get_by_stage($owner->baseClass(), Versioned::LIVE)
-                ->byID($owner->ID);
-        }
+        return DBDatetime::withFixedNow($now, function () {
```
maxime-rainville, line 62: The only problem is that `withFixedNow` will need to be in a new minor release. If you want to fix this in a patch release, you'll need to directly call `set_mock_now`.

mfendeksilverstripe: I can't use `set_mock_now` as that approach will cause the previous mock value to be lost. Note that this functionality is used in many unit tests (already tried).

TLDR:

`clear_mock_now` sets the value to `null` not the previous value.

We will need the dependent PR to be merged first.

mfendeksilverstripe:
@maxime-rainville Retargeted to `1`

maxime-rainville:
Cool ... we'll just need to update the composer file to explicitly require framework `^4.7`.

mfendeksilverstripe:
@maxime-rainville Right, done 👍

src/RecursivePublishable.php
sminnee: Dependent PR merged to `4`

maxime-rainville:
Just restarted the build. Once its green, I'll have a go at running it locally as a final check to make sure it works.

tests/php/PublishRecursive/SlowDummyPage.php
```diff
@@ -0,0 +1,55 @@
...
+use SilverStripe\Dev\TestOnly;
+
+/**
+ * Class SlowDummyPage
+ *
+ * @property int $NestedObjectID
+ * @method SlowDummyObject NestedObject()
+ * @package SilverStripe\Versioned\Tests\PublishRecursive
+ */
+class SlowDummyPage extends SiteTree implements TestOnly
```
maxime-rainville, line 15: CMS doesn't get installed by the build. So SiteTree doesn't exist in this context. Do you mind re-implementing your test with a Dummy DataObject instead?

mfendeksilverstripe: Yup, should be fine.

mfendeksilverstripe:
@mfendeksilverstripe Pushed up unit test fixes.

tests/php/PublishRecursive/SlowDummyParent.php
```diff
@@ -0,0 +1,63 @@
...
+    /**
+     * @var array
+     */
+    private static $cascade_duplicates = [
+        'NestedObject',
+    ];
+
+    protected function onBeforeWrite()
+    {
+        sleep(2);
```
tractorcow, line 60: Adding sleeps to the unit tests is a very bad idea. I realise that this is time-critical, but this is going to only add incremental slowdowns to an already massive test suite.

A possible alternative; Clear mock now in your unit tests. Hook into `onBeforePublishRecursive` extension, and add a check to ensure that `DBDatetime` field has a specific mock assigned (i.e. fail if unmocked). You may need to use reflection to expose the protected static var.

If your code is behaving correctly, datetime will have a mock assigned during this process.

If the code is not behaving correctly, and the mock is unset (i.e. during onBeforePublishRecursive) then fail the test.

mfendeksilverstripe: Fair enough will try it out, thanks :)

mfendeksilverstripe:
@maxime-rainville @tractorcow Pushed up sleep emulation (unit test fixes).

tests/php/PublishRecursive/SlowDummyParent.php
```diff
@@ -0,0 +1,101 @@
...
+        DBDatetime::set_mock_now($now);
+    }
+
+    /**
+     * @return DBDatetime|null
+     * @throws ReflectionException
+     */
+    private function getMockNow()
+    {
+        $propertyMockNow = new ReflectionProperty(DBDatetime::class, 'mock_now');
```
tractorcow, line 96: 👍

tests/php/PublishRecursive/SlowDummyParent.php
```diff
@@ -0,0 +1,101 @@
...
+    {
+        $this->emulateSleep(2);
+        parent::onBeforeWrite();
+    }
+
+    /**
+     * @param int $seconds
+     * @throws Exception
+     */
+    private function emulateSleep($seconds)
```
tractorcow, line 73: 👍

maxime-rainville:
The Travis failure is caused by our old friend postgresql. I'm having a look to try to debug what's going on.

tests/php/PublishRecursive/PublishRecursiveTest.php
```diff
@@ -0,0 +1,93 @@
...
+        $parent = SlowDummyParent::create();
+        $parent->Title = 'Slow Parent';
+        $parent->NestedObjectID = (int) $object->ID;
+        $parent->write();
+
+        $parent->publishRecursive();
+
+        $versionsSufix = '_Versions';
+        $parentTable = SlowDummyParent::config()->get('table_name');
+        $parentVersionedTable = $parentTable . $versionsSufix;
```
maxime-rainville, line 62:
```suggestion
        $parentVersionedTable = '"' . $parentTable . $versionsSufix . '"';
```

tests/php/PublishRecursive/PublishRecursiveTest.php
```diff
@@ -0,0 +1,93 @@
...
+        $objectTable = SlowDummyObject::config()->get('table_name');
+        $objectVersionedTable = $objectTable . $versionsSufix;
```
maxime-rainville, line 64:
```suggestion
        $objectVersionedTable = '"' . $objectTable . $versionsSufix . '"';
```

tests/php/PublishRecursive/PublishRecursiveTest.php
```diff
@@ -0,0 +1,93 @@
...
+        $tables = [
+            $parentVersionedTable => $parent->ID,
+            $objectVersionedTable => $object->ID,
+        ];
+
+        $results = [];
+
+        foreach ($tables as $table => $id) {
+            $query = SQLSelect::create(
+                'LastEdited',
```
maxime-rainville, line 74:
```suggestion
                '"LastEdited"',
```

tests/php/PublishRecursive/PublishRecursiveTest.php
```diff
@@ -0,0 +1,93 @@
...
+                $table,
+                ['RecordID' => $id],
+                ['Version' => 'DESC'],
```
maxime-rainville, line 77:
```suggestion
                ['"RecordID"' => $id],
                ['"Version"' => 'DESC'],
```

mfendeksilverstripe:
Thanks for finding the issue @maxime-rainville , I pushed up fixes and unit tests are now all green 👍

PR #293, opened 2020-08-05
GraphQL 4 Compatability

unclecheese:


ScopeyNZ:
Can you leave all the old stuff and manage to keep compatibility with GraphQL <4. As it stands, all these removals of classes cannot happen without a major release.

unclecheese:
Yup, fair point. I was just thinking that those classes rely on ancestors that won't exist anymore, but we can put `class_exists()` guards around them like the rest of the module. I'll move the deletion commits to the master branch.

composer.json
```diff
@@ -28,9 +28,6 @@
         "silverstripe/admin": "^1",
         "squizlabs/php_codesniffer": "^3"
     },
-    "conflict": {
```
sminnee: Would it be better to say "conflict: silverstripe/graphql < 5"?

unclecheese:
Moved to: https://github.com/silverstripe/silverstripe-versioned/pull/308

PR #294, opened 2020-08-16
FIX: Decouple test setup.

sminnee:
Decoupling modules so that versioned doesn’t rely on an arbitrary and
undefined subset of the CMS recipe nudges our architecture in the right
direction. A more successful decomposition of the recipe into packages
would ensure that there aren’t bidirectional or cyclic relationships
between packages, and instead they are arranged more into “layers”.

This patch means that you can test the module independently, like so:

 * (clone module)
 * composer install
 * vendor/bin/phpunit

The test suit not longer relies on HTMLEditorField, which in practise
is not fully functional without admin. It also skips 

There is 1 test that appears to test a piece of LeftAndMain behaviour.
Arguably it should be refactored and placed there, but I’ve made it
skippable instead.

With that in place it tightens the require-dev dependencies to only
include graphql, as some of the test suite tests the graphql/versioned
behaviour.

I haven’t changed the travis.yml configuration as is appears these are
being auto-generated to be consistent across all modules that make up
the CMS recipe (44fb86acffcd3f4041b03f12299b07348439b2f0).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,2 +1,6 @@
 node_modules/
 /**/*.js.map
+vendor/
+resources/
+composer.lock
+assets/
--- a/composer.json
+++ b/composer.json
@@ -25,8 +25,8 @@
     },
     "require-dev": {
         "phpunit/phpunit": "^5.7",
-        "silverstripe/admin": "^1",
-        "squizlabs/php_codesniffer": "^3"
+        "squizlabs/php_codesniffer": "^3",
+        "silverstripe/graphql": "^3"
     },
     "extra": {
         "expose": [
--- a/tests/php/VersionedTest/TestObject.php
+++ b/tests/php/VersionedTest/TestObject.php
@@ -34,7 +34,7 @@ class TestObject extends DataObject implements TestOnly
     private static $db = [
         "Name" => "Varchar",
         'Title' => 'Varchar',
-        'Content' => 'HTMLText',
+        'Content' => 'Text',
     ];
 
     private static $extensions = [
--- a/tests/php/VersionedStateExtensionTest.php
+++ b/tests/php/VersionedStateExtensionTest.php
@@ -65,6 +65,11 @@ class VersionedStateExtensionTest extends SapphireTest
 
     public function testDontUpdateLeftAndMainLinks()
     {
+        if (!class_exists(LeftAndMain::class)) {
+            $this->markTestSkipped('silverstripe/cms not installed');
+            return;
+        }
+
         $controller = new LeftAndMain();
 
         $liveClientConfig = $controller->getClientConfig();


sminnee:
Hey @emteknetnz following on from that other commit, I'd note that you don't actually need the whole recipe to run the tests; just the framework. I wasn't sure whether the team had decided that pulling the recipe in as a test dependency was a bug or a feature. ;-)

emteknetnz:
>> I wasn't sure whether the team had decided that pulling the recipe in as a test dependency was a bug or a feature. ;-)

I'd say it was closer to a bug, though intended as a feature to get more consistency on .travis.yml files

Previously it has recipe-core, though it was changed to recipe-cms which contains recipe-core, so nothing broke, though contains some extra stuff that's unrequired as you say.

sminnee:
Yeah, it's not going to break tests, but it's more the opposite problem: If there is code that inadvertently relies on random other parts of recipe-cms, this won't show up as test failures. It's not usually a big deal but it can mean that for example you occasionally trip over bugs when building a modeladmin-only app.

The second problem, which is that if you're weird like me and run your silverstripe-versioned test suite by checking out the module, and running "composer install; vendor/bin/phpunit" inside that, it can be a little harder to do. This is solved by my changes in this PR which fills out the require-dev dependencies.

sminnee:
Basically I didn't want to "fix" the travis.yml config to pull in only the dependencies necessary for testing the module, only to have it reverted again in the future. ;-)

composer.json
```diff
@@ -25,8 +25,8 @@
     },
     "require-dev": {
         "phpunit/phpunit": "^5.7",
-        "silverstripe/admin": "^1",
-        "squizlabs/php_codesniffer": "^3"
+        "squizlabs/php_codesniffer": "^3",
+        "silverstripe/graphql": "^3"
```
emteknetnz, line 29: Was the only thing that `silverstripe/admin` was included for the `VersionedStateExtensionTest.php`?

Why is `silverstripe/graphql` is being included here?

tests/php/VersionedTest/TestObject.php
```diff
@@ -34,7 +34,7 @@ class TestObject extends DataObject implements TestOnly
     private static $db = [
         "Name" => "Varchar",
         'Title' => 'Varchar',
-        'Content' => 'HTMLText',
+        'Content' => 'Text',
```
emteknetnz, line 37: Why is this changing?

.gitignore
```diff
@@ -1,2 +1,6 @@
 node_modules/
 /**/*.js.map
+vendor/
+resources/
+composer.lock
+assets/
```
emteknetnz, line 6: a bit non-standard having these things here, though not harmful in any way

How come assets is here though, that's quite weird?  Does a public folder get installed on your test setup?

tests/php/VersionedTest/TestObject.php
sminnee: HTMLText requires admin, but versioned doesn't.

.gitignore
sminnee: You can run tests of a module like this, which is how they get run in travis – you don't need to manufacture a project to run them inside. This capability was something introduced in SS4 and was part of the general decoupling.

```
git clone <this repo>
cd silverstripe-versioned
composer install
vendor/bin/phpunit
```

But if you do that you get some folders hanging around. Resources and assets are a bit of a hangover of the fact that framework sill requires silverstripe/assets (which I don't like) but listing them here is fairly harmless.

composer.json
sminnee: * Without admin, HTMLText fields don't work very well. Testing with Text fields was just as good, though.
 * VersionedStateExtensionTest::testDontUpdateLeftAndMainLinks() requires admin, yes. It seemed a bit of weird coupling, and kind of feels like it's more a test for the silverstripe/admin suite. But I opted to mark it as skipped.
 * Versioned optionally supports silverstripe/graphql and so this ensures that you can test both the optional GraphQL plugins  and the core.

PR #299, opened 2020-09-25
ENH Replace E_USER_ERROR errors with exceptions

robbieaverill:
Issue: https://github.com/silverstripe/silverstripe-framework/issues/1915

### DIFF
--- a/src/ChangeSetItem.php
+++ b/src/ChangeSetItem.php
@@ -338,7 +338,7 @@ class ChangeSetItem extends DataObject implements Thumbnail
     /** Reverts this item, then close it. **/
     public function revert()
     {
-        user_error('Not implemented', E_USER_ERROR);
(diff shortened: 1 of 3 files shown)


src/ChangeSetItem.php
```diff
@@ -338,7 +338,7 @@ public function unlinkDisownedObjects()
     /** Reverts this item, then close it. **/
     public function revert()
     {
-        user_error('Not implemented', E_USER_ERROR);
+        throw new \Exception('Not implemented');
```
dhensby, line 341: I think a `RuntimeException` would be more appropriate? I think we should avoid using the base `Exception` class where we can.

dhensby (changes requested):

ScopeyNZ (approved):

PR #302, opened 2020-10-01
Improvement: Update Versioned Admin menu title to "Archive"

chrispenny:
Fixes: https://github.com/silverstripe/silverstripe-versioned-admin/issues/156

> This was a design mishap where it was updated in one place and not another causing inconsistent across the board when it was developed.

### DIFF
--- a/lang/de.yml
+++ b/lang/de.yml
@@ -1,10 +1,4 @@
 de:
-  SilverStripe\VersionedAdmin\ArchiveAdmin:
-    MENUTITLE: Archiv
-  SilverStripe\VersionedAdmin\Controllers\CMSPageHistoryViewerController:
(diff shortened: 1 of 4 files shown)


robbieaverill (commented):
It looks to me like these translations should be in the versioned-admin module. Can we move them there instead?

chrispenny:
Yup, definitely.

I'll update: https://github.com/silverstripe/silverstripe-versioned-admin/pull/175

chrispenny:
Updated the PR here to include the addition of these translations:
https://github.com/silverstripe/silverstripe-versioned-admin/pull/175/files

robbieaverill (approved):
Lovely, thanks!

PR #308, opened 2020-11-02
NEW: GraphQL 4 Compatibility

unclecheese:
Moved from: https://github.com/silverstripe/silverstripe-versioned/pull/293


_config/graphql_schema.yml
```diff
@@ -0,0 +1,11 @@
+---
+Name: versioned-graphql-schema
+Only:
+  moduleexists: 'silverstripe/graphql'
+  classexists: 'SilverStripe\GraphQL\Schema\Schema'
+---
+SilverStripe\GraphQL\Schema\Schema:
+  schemas:
+    '*':
```
chillu, line 9: Adding this to all schemas by default encourages the use of previews in your implementation, but also increases the security surface of your API. Given it's non-trivial to opt out of this, I think we need to add more docs on an "opt out config"

src/GraphQL/Resolvers/VersionFilters.php
```diff
@@ -0,0 +1,238 @@
...
+    public function applyToReadingState(array $versioningArgs)
+    {
+        if (!isset($versioningArgs['mode'])) {
+            return null;
+        }
+
+        $this->validateArgs($versioningArgs);
+
+        $mode = $versioningArgs['mode'];
+        switch ($mode) {
```
chillu, line 31: I think this needs to check permissions before setting stages, it's functionally identical to `VersionedHTTPMiddleware` which does that (via `Versioned::can_choose_site_stage()`)

src/GraphQL/Resolvers/VersionFilters.php
```diff
@@ -0,0 +1,238 @@
...
+            ));
+        }
+        if (!isset($versioningArgs['mode'])) {
+            return $list;
+        }
+
+        $this->validateArgs($versioningArgs);
+
+        $mode = $versioningArgs['mode'];
+        switch ($mode) {
```
chillu, line 79: Same here, needs permission checks

src/GraphQL/Resolvers/VersionFilters.php
```diff
@@ -0,0 +1,238 @@
...
+            case 'all_versions':
+                $list = $list->setDataQueryParam('Versioned.mode', 'all_versions');
+                break;
+            case 'latest_versions':
+                $list = $list->setDataQueryParam('Versioned.mode', 'latest_versions');
+                break;
+            case 'status':
+                // When querying by Status we need to ensure both stage / live tables are present
+                $baseTable = singleton($list->dataClass())->baseTable();
+                $liveTable = $baseTable . '_Live';
```
chillu, line 101: Should use `Versioned->stageTable()`

src/GraphQL/Resolvers/VersionFilters.php
```diff
@@ -0,0 +1,238 @@
...
+                $statuses = $versioningArgs['status'];
+
+                // If we need to search archived records, we need to manually join draft table
+                if (in_array('archived', $statuses)) {
+                    $list = $list
+                        ->setDataQueryParam('Versioned.mode', 'latest_versions');
+                    // Join a temporary alias BaseTable_Draft, renaming this on execution to BaseTable
+                    // See Versioned::augmentSQL() For reference on this alias
+                    $draftTable = $baseTable . '_Draft';
```
chillu, line 110: Should use `Versioned->stageTable()`

src/GraphQL/Resolvers/VersionedResolver.php
```diff
@@ -0,0 +1,255 @@
...
+            }
+            if (!$object->canViewStage(Versioned::DRAFT, $context[QueryHandler::CURRENT_USER])) {
+                throw new Exception(sprintf(
+                    'Cannot view versions on %s',
+                    $this->getDataObjectClass()
+                ));
+            }
+
+            // Get all versions
+            return $object->VersionsList();
```
chillu, line 89: Is this paginated? Feels like we'd run into performance issues otherwise, objects can have a *lot* of versions. Can be a follow up enhancement ticket, not a blocker for the PR.

chillu (changes requested):

tests/php/GraphQL/Plugins/VersionedReadTest.php
```diff
@@ -0,0 +1,52 @@
...
+use SilverStripe\Security\Member;
+use SilverStripe\Security\Security;
+use SilverStripe\Versioned\GraphQL\Plugins\VersionedDataObject;
+use SilverStripe\Versioned\GraphQL\Plugins\VersionedRead;
+use SilverStripe\Versioned\GraphQL\Resolvers\VersionedResolver;
+use SilverStripe\Versioned\GraphQL\Types\VersionedInputType;
+use SilverStripe\Versioned\Tests\GraphQL\Fake\Fake;
+use SilverStripe\Versioned\Versioned;
+
+class VersionedReadTest extends SapphireTest
```
chillu, line 27: There's some autoloading warnings in the build: https://travis-ci.com/github/silverstripe/silverstripe-versioned/jobs/429504664#L649

src/GraphQL/Resolvers/VersionFilters.php
chillu: Since this is pre-existing functionality, I've moved the discussion to https://github.com/silverstripe/silverstripe-versioned/issues/310

src/GraphQL/Resolvers/VersionFilters.php
chillu: https://github.com/silverstripe/silverstripe-versioned/issues/310

src/GraphQL/Resolvers/VersionFilters.php
chillu: https://github.com/silverstripe/silverstripe-versioned/issues/310

tests/php/GraphQL/Plugins/VersionedReadTest.php
chillu: Wider discussion at https://github.com/silverstripe/silverstripe-graphql/issues/282#issuecomment-725777083

_graphql/enums.yml
```diff
@@ -0,0 +1,47 @@
...
+    DRAFT:
+      value: Stage
+      description: The draft stage
+    LIVE:
+      value: Live
+      description: The live stage
+
+VersionedQueryMode:
+  description: The versioned mode to use
+  values:
```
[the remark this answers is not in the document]
unclecheese: This may be outdated. I see ALL_VERSIONS below. (It wasn't before, and was causing versioned-admin tests to fail)

src/GraphQL/Resolvers/VersionedResolver.php
unclecheese: Yeah it's paginated here: https://github.com/silverstripe/silverstripe-versioned/blob/pulls/1/schemageddon/src/GraphQL/Plugins/VersionedDataObject.php#L102

chillu: Pre-existing bug. Even if we implemented pagination here, we'd also need to find a way to use it in the UI - scope creep. See https://github.com/silverstripe/silverstripe-versioned-admin/issues/185

_config/graphql_schema.yml
unclecheese: Tracking: https://github.com/silverstripe/silverstripe-graphql/issues/320

PR #312, opened 2020-11-16
Fixed implicit GraphQL dependency

chillu:
versioned has require-dev with "silverstripe/graphql": "3.x-dev || 4.x-dev".
framework requires versioned.
require-dev only kicks in for root requirements.
versioned GraphQL implementation and tests have some if(!class_exists(<graphql-class>)) return false which prevents issues when installing versioned without GraphQL
versioned also uses this to distinguish between implementations and tests relying on GraphQL v3 and v4 classes
This isn’t consistently implemented in versioned

So if you do a composer create-project silverstripe/framework . && vendor/bin/phpunit, it’ll fail tests because it’s missing classes. My assumption is that we don’t want to make make GraphQL a hard requirement of versioned, so I’m implementing the class_exists() checks consistently.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,7 +43,7 @@ before_script:
   # Fix for running phpunit 5 on php 7.4+
   - composer require --no-update sminnee/phpunit-mock-objects:^3
 
-  - 'if [[ ! $BACKWARD_COMPAT ]]; then composer require silverstripe/graphql:4.x-dev --no-update; fi'
+  - 'if [[ $BACKWARD_COMPAT ]]; then composer require silverstripe/graphql:3.x-dev --no-update; fi'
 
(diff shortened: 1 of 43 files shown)


unclecheese (approved):

chillu:
I've hopefully fixed the tests on the next run, and also flipped around the `BACKWARD_COMPAT` usage. 

Current state:
 * `BACKWARD_COMPAT=` builds with graphql:dev-master: https://travis-ci.com/github/silverstripe/silverstripe-asset-admin/jobs/440490062#L339
 * `BACKWARD_COMPAT=1` builds with graphql:dev-master: https://travis-ci.com/github/silverstripe/silverstripe-asset-admin/jobs/440490061#L340

Issue #321, opened 2021-02-05
Update to symfony-cache 4


sb-relaxt-at:
As the core Silverstripe components have moved to Symfony 4 it would be great if this module could be updated as well :-)

### PR
- [x] https://github.com/silverstripe/silverstripe-versioned/pull/322

emteknetnz:
Linked PR has been merged

PR #322, opened 2021-02-12
MNT Remove explicit Symfony cache dependency

dnsl48:
Rely on the framework requirements instead.

Resolves https://github.com/silverstripe/silverstripe-versioned/issues/321

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -20,8 +20,8 @@
     ],
     "require": {
         "silverstripe/framework": "^4.7",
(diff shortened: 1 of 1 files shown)


emteknetnz:
@dnsl48 Instead of removing this I've updated it to `^3.4 || ^4` - I noticed the framework has `^3.3 || ^4` - I don't know the significant (if any) of 3.4 vs 3.3 specifically, so it's probably easier to just append ` || ^4`

PR #325, opened 2021-02-23
Unset sort for subquery if we're using MS SQL database

### LINKED ISSUE #324: Incorrect SQL query generated for MS SQL database
Found a bug while trying to set up a site using MS SQL database. 
I'm using MS SQL 2019 database, with `SS_DATABASE_CLASS="MSSQLDatabase"`

**Test case:**
1. Install:
- SS 4.7.2
- silverstripe/mssql: 2.0.3
- tractorcow/silverstripe-fluent: 5.0.0
- dnadesign/silverstripe-elemental: 4.0.0
2. Add ElementalPageExtension to Page:
```yml
Page:
  extensions:
    - DNADesign\Elemental\Extensions\ElementalPageExtension
```
3. Add one (Elemental) block to page; Content is default - it's ok
4. Publish page

**Throws error:**
> [Emergency] Uncaught SilverStripe\ORM\Connect\DatabaseException: Couldn't run query: UPDATE "Element_Live" SET "ParentID" = ? WHERE ("Element_Live"."ParentID" = ?) AND ("Element_Live"."ID" NOT IN SELECT "Source"."ID" FROM SELECT DISTINCT "Element"."ClassName", "Element"."LastEdited", "Element"."Created", "Element"."Version", "Element"."Title", "Element"."ShowTitle", "Element"."Sort", "Element"."ExtraClass", "Element"."Style", "Element"."ParentID", "Element"."ID", CASE WHEN "Element"."ClassName" IS NOT NULL THEN "Element"."ClassName" ELSE N'DNADesign\Elemental\Models\BaseElement' END AS "RecordClassName" FROM "Element" WHERE ("Element"."ParentID" = ?) ORDER BY "Element"."Sort" ASC) AS "Source")) 42000, 1033, [Microsoft][ODBC Driver 17 for SQL Server][SQL Server]
The ORDER BY clause is invalid in views, inline functions, derived tables, subqueries, and common table expressions, unless TOP, OFFSET or FOR XML is also specified.

Looks like when `RecursivePublishable::unlinkDisownedRelationship` method is called it generates otherwise normal query but MS SQL doesn't like it. 

It works if I try to comment out $default_sort setting on BaseElement class.

So after a bit of digging I came up to solution to unset sort order in subquery if we're using MS SQL database. I'm submitting PR also (#325 ).

t3hn0:
Fixes #324

### DIFF
--- a/src/RecursivePublishable.php
+++ b/src/RecursivePublishable.php
@@ -412,9 +412,9 @@ class RecursivePublishable extends DataExtension
         // Query the source for the list of items to NOT remove
         $ownedSQL = $owner->getComponents($relationship);
         // unset sort for subquery if we're using MSSQL
-        if( class_exists("SilverStripe\\MSSQL\\MSSQLSchemaManager") ) {
+        if (class_exists("SilverStripe\\MSSQL\\MSSQLSchemaManager")) {
             $manager = DB::get_schema();
-            if( get_class($manager) == "SilverStripe\MSSQL\MSSQLSchemaManager" ) {
+            if (get_class($manager) == "SilverStripe\MSSQL\MSSQLSchemaManager") {
                 $ownedSQL = $ownedSQL->dataQuery()->sort(null, null, true);
             }
         }
--- a/src/RecursivePublishable.php
+++ b/src/RecursivePublishable.php
@@ -410,9 +410,9 @@ class RecursivePublishable extends DataExtension
         }
 
         // Query the source for the list of items to NOT remove
-        $ownedSQL = $owner->getComponents($relationship);
+        $ownedQuery = $owner->getComponents($relationship);
         // remove sort for sub query because some DB drivers don't let you sort sub queries
-        $ownedSQL = $ownedSQL->dataQuery()->sort(null, null, true)->sql($ownedParams);
+        $ownedSQL = $ownedQuery->dataQuery()->sort(null, null, true)->sql($ownedParams);
         $disowned->addWhere([
             "\"{$targetTable}\".\"ID\" NOT IN (SELECT \"Source\".\"ID\" FROM ({$ownedSQL}) AS \"Source\")" => $ownedParams
         ]);
--- a/src/RecursivePublishable.php
+++ b/src/RecursivePublishable.php
@@ -411,14 +411,8 @@ class RecursivePublishable extends DataExtension
 
         // Query the source for the list of items to NOT remove
         $ownedSQL = $owner->getComponents($relationship);
-        // unset sort for subquery if we're using MSSQL
-        if (class_exists("SilverStripe\\MSSQL\\MSSQLSchemaManager")) {
-            $manager = DB::get_schema();
-            if (get_class($manager) == "SilverStripe\MSSQL\MSSQLSchemaManager") {
-                $ownedSQL = $ownedSQL->dataQuery()->sort(null, null, true);
-            }
-        }
-        $ownedSQL = $ownedSQL->sql($ownedParams);
+        // remove sort for sub query because some DB drivers don't let you sort sub queries
+        $ownedSQL = $ownedSQL->dataQuery()->sort(null, null, true)->sql($ownedParams);
         $disowned->addWhere([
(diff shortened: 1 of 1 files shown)


t3hn0:
I'm not sure how to fix error with "codecov/patch" check.

src/RecursivePublishable.php
```diff
@@ -409,7 +410,15 @@ public function unlinkDisownedRelationship($source, $targetStage, $relationship)
         }
 
         // Query the source for the list of items to NOT remove
-        $ownedSQL = $owner->getComponents($relationship)->sql($ownedParams);
+        $ownedSQL = $owner->getComponents($relationship);
+        // unset sort for subquery if we're using MSSQL
+        if (class_exists("SilverStripe\\MSSQL\\MSSQLSchemaManager")) {
```
maxime-rainville, line 415: I'd rather avoid having references to `SilverStripe\MSSQL\MSSQLSchemaManager` in a core module because the MSSql module is not supported.

Looking at the original query you put on the parent issue, I'm 95% that none of the other DB drivers will care about whether this sub query is sorted or not. If anything, removing the sorting on the sub query might speed things up a tiny little bit.

My suggestion is just remove all the conditional statement and disable sorting for all the DB Drivers. Maybe add a quick comment saying that we're removing the sorting because some DB drivers don't let you sort sub queries.


If the test are green than we can merge this one.

src/RecursivePublishable.php
```diff
@@ -409,7 +410,15 @@ public function unlinkDisownedRelationship($source, $targetStage, $relationship)
         }
 
         // Query the source for the list of items to NOT remove
-        $ownedSQL = $owner->getComponents($relationship)->sql($ownedParams);
+        $ownedSQL = $owner->getComponents($relationship);
```
maxime-rainville, line 413: Maybe rename this variable until you actually get it converted to an SQL string. Otherwise, you'll have a variable that start as a `DataList` and then gets assigned a string value later on.

```suggestion
        $ownedQuery = $owner->getComponents($relationship);
```

src/RecursivePublishable.php
```diff
@@ -409,7 +410,15 @@ public function unlinkDisownedRelationship($source, $targetStage, $relationship)
...
+        // unset sort for subquery if we're using MSSQL
+        if (class_exists("SilverStripe\\MSSQL\\MSSQLSchemaManager")) {
+            $manager = DB::get_schema();
+            if (get_class($manager) == "SilverStripe\MSSQL\MSSQLSchemaManager") {
+                $ownedSQL = $ownedSQL->dataQuery()->sort(null, null, true);
```
maxime-rainville, line 418:
```suggestion
                $ownedQuery = $ownedQuery->dataQuery()->sort(null, null, true);
```

src/RecursivePublishable.php
```diff
@@ -409,7 +410,15 @@ public function unlinkDisownedRelationship($source, $targetStage, $relationship)
...
+            }
+        }
+        $ownedSQL = $ownedSQL->sql($ownedParams);
```
maxime-rainville, line 421:
```suggestion
        $ownedSQL = $ownedQuery->sql($ownedParams);
```

maxime-rainville (changes requested):
Thanks for reporting this issue and providing a fix. That's awesome.

If you implement the suggestions below, we can most likely get this merged.

`codecov/patch` is complaining because the the conditional logic you added never gets run since we don't test for MS SQL. Disabling sorting all the time should squash that warning.

src/RecursivePublishable.php
t3hn0: I implemented your suggestion.

src/RecursivePublishable.php
t3hn0: I implemented your suggestion.

src/RecursivePublishable.php
t3hn0: removed both conditionals and added comment about removing sort in sub queries.

t3hn0 (commented):
Requested changes were implemented

maxime-rainville (approved):
Good stuff

PR #326, opened 2021-02-23
BUGFIX: replace versions pagination, sort to fix versioned-admin

unclecheese:
versioned-admin is broken because of this. See: https://github.com/silverstripe/silverstripe-graphql/issues/309

### DIFF
--- a/src/GraphQL/Plugins/VersionedDataObject.php
+++ b/src/GraphQL/Plugins/VersionedDataObject.php
@@ -8,8 +8,11 @@ use SilverStripe\GraphQL\Schema\Exception\SchemaBuilderException;
 use SilverStripe\GraphQL\Schema\Field\Field;
 use SilverStripe\GraphQL\Schema\Interfaces\ModelTypePlugin;
 use SilverStripe\GraphQL\Schema\Interfaces\SchemaUpdater;
+use SilverStripe\GraphQL\Schema\Plugin\AbstractQuerySortPlugin;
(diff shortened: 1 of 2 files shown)


emteknetnz:
@unclecheese what exactly is broken?  I'm not sure what really sure how I'm supposed to go about evaluating this pull-request?

unclecheese:
Versioned admin when running graphql 4.x

emteknetnz:
Yeah but what's broken? :-)  Is it what's in the description of the linked issue?  Or is it something else that broke as part of the recent graphql 4 refactor?

unclecheese:
Yeah there was a [refactor](https://github.com/silverstripe/silverstripe-graphql/commit/35fa149a996357a8fdf11628fae5f83d14015316) recently that changed a bunch of stuff, including when and how plugins are applied. Versioned-admin relied on a plugin adding a plugin, which somehow worked in earlier versions, but doesn't anymore. A plugin adding a plugin shouldn't be a thing, so it's been refactored to use a very imperative approach that will eventually [get refactored](https://github.com/silverstripe/silverstripe-graphql/issues) as well. :-)

Bit of a temporary fix to get the builds green and CMS fulling working again in GraphQL 4.

emteknetnz (approved):
Cheers for the explanation.  If you can think of a cleaner and more proper way to do this then that may be worth pursuing, though given the context of just trying to the graphql4 bit green then this seems OK.

PR #328, opened 2021-03-11
NEW: Stages differ recursive

mfendeksilverstripe:
# Stages differ recursive

* adds capability to check if an object which owns nested objects needs to be published or not (stages differ recursive)
* this covers common cases where publishing cascades down to nested objects
* examples: block page with blocks, complex block such as carousel block with carousel items
* this feature queries the nested objects based on existing relationship setup so it doesn't add any overhead for the developers to implement any functionality
* there are two modes available:
  * strong ownership - traverses via `owns` relation
  * weak ownership - traverses via `cascade_duplicates` relation

## Parent issue
- #420


src/RecursiveStagesService.php
```diff
@@ -0,0 +1,171 @@
...
+use SilverStripe\ORM\SS_List;
+
+/**
+ * Class RecursiveStagesService
+ *
+ * Functionality for detecting the need of publishing nested objects owned by common parent / ancestor object
+ *
+ * @package SilverStripe\Versioned
+ */
+class RecursiveStagesService
```
maxime-rainville, line 17: That might just be a pet peeve, but I think there's value in creating interfaces for services, so people who want to override a class/services know what they are coding to.

src/RecursiveStagesService.php
```diff
@@ -0,0 +1,171 @@
...
+use SilverStripe\ORM\DataList;
+use SilverStripe\ORM\DataObject;
+use SilverStripe\ORM\SS_List;
+
+/**
+ * Class RecursiveStagesService
+ *
+ * Functionality for detecting the need of publishing nested objects owned by common parent / ancestor object
+ *
+ * @package SilverStripe\Versioned
```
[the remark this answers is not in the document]
maxime-rainville: I might be tempted to mark this class as `@internal` to avoid promising full support right away

maxime-rainville (commented):
Just double checking my understanding of the purpose here. Basically, you want a way to check if recursively publishing a DataObject will lead to any other owned/related Dataobjects to be published as well?

Presumably, this would allow us to have a smarter "dirty" state. e.g.: You have a UserForm page. The UserForm page has no outstanding changes directly applied to it, but one of its form field has been modified. So logically, the UserForm page should be in a dirty state.

src/RecursiveStagesService.php
mfendeksilverstripe: Sure, can do.

mfendeksilverstripe:
@maxime-rainville That's correct, the goal is to have smarter "dirty" state and also allow some level of configuration.

maxime-rainville:
From memory, there's a `ChangeSet` object which is used by `publishRecursive`. My main concern at this stage is that this PR might be duplicating the ChangeSet logic and/or diverging from it in subtle ways that will create ambiguity down the road. It's probably worth asking ourselves if there's a way we could have some sort of unified logic for both.

@clarkepaul Currently, if you have an object that owns a bunch of child object with unpublished changes, the CMS doesn't show that object in a "dirty" state. If this PR got merge, we would be very close to being able to highlight such a scenario. How much value would you put on this kind of improvement?

e.g.: You have a UserForm page. The UserForm page has no outstanding changes directly applied to it, but one of its form field has been modified. So logically, the UserForm page should be in a dirty state as well.

clarkepaul:
@maxime-rainville it does trip CMS users up, they think their page is published and nothing needs to be done further when in-fact deeper content isn't reflected on the published page, hence I would put a lot of value on this. We don't have the ability to train users about this stuff so rely on UI experiences to inform people.

maxime-rainville:
@brynwhyman I had a preliminary review. Looks like this as the potential to provide substantial benefit beyond Mojmir's use case.

My main concern would be making sure the chosen approach aligns with the existing logic in ChangeSet.

mfendeksilverstripe:
@maxime-rainville Added an interface for the service as requested.

brynwhyman:
Just want to cc @chillu into this pull request too. This sounds like it might overlap some of the work that was done with versioned-snapshots to better understand the state of child objects - admittedly that had more of a focus on the version records, but still...

As an aside, would this PR also resolve this issue? https://github.com/silverstripe/silverstripe-elemental/issues/790

mfendeksilverstripe:
@brynwhyman Just to provide some context:

Version snapshots it essentially a logger of events, the main drive for this module was to bypass issues with Fluent / Versioned (which are now resolved in Fluent 6), it doesn't provide you with information about state of objects which is the goal of this PR.

brynwhyman:
Alright, thanks @mfendeksilverstripe that's helpful 😄

halles:
Does this need more reviewing? I can help out. In the meantime I've patched this with an extension, but this seems much more robust.

FWIW This has been flagged as a bug by clients of ours

GuySartorelli:
As this is adding new functionality, please target this to the `2` branch, rebase against that branch, any resolve and conflicts that might arise.

_config/versionedownership.yml
```diff
@@ -12,3 +12,9 @@ OnlyIf:
 SilverStripe\Admin\LeftAndMain:
   extensions:
     RecursivePublishableHandler: SilverStripe\Versioned\RecursivePublishableHandler
+---
+Name: versionedrecursivestages
+---
+SilverStripe\Core\Injector\Injector:
+  SilverStripe\Versioned\RecursiveStagesInterface:
+    class: SilverStripe\Versioned\RecursiveStagesService
```
ChloeSartorelli: Why is this in its own block? It should probably either go in the `versionedownership` block above, or in its own file.

src/RecursiveStagesInterface.php
```diff
@@ -0,0 +1,24 @@
+<?php
+
+namespace SilverStripe\Versioned;
+
+use SilverStripe\ORM\DataObject;
+
+/**
+ * Interface RecursiveStagesInterface
+ *
+ * Interface for @see RecursiveStagesService to provide the capability to for "smart" durty model state
```
ChloeSartorelli:
```suggestion
 * Interface for @see RecursiveStagesService to provide the capability to for "smart" dirty model state
```
What is "smart" dirty model state?

src/RecursiveStagesInterface.php
```diff
@@ -0,0 +1,24 @@
...
+ */
+interface RecursiveStagesInterface
+{
+
+    /**
+     * Determine if content differs on stages including nested objects
+     *
+     * @param DataObject $object
+     * @param string $mode
+     * @return bool
```
ChloeSartorelli:
```suggestion
```
No need to duplicate what's already in the explicit type hints

Please apply this to any other new API as well - we only need PHPDoc references to params or return statements if the PHPDoc adds extra context about what that param/return statement is for or how it's used.

src/RecursiveStagesInterface.php
ChloeSartorelli, line 1: Why is this an entire interface? It seems to me like this would just belong on the `Versioned` extension, no?

src/RecursiveStagesService.php
ChloeSartorelli, line 1: What is the intention of having this as its own separate class? These sorts of methods normally live on `Versioned` I think

src/RecursiveStagesService.php
```diff
@@ -0,0 +1,166 @@
...
+
+    /**
+     * Strong ownership uses 'owns' configuration to determine relationships
+     */
+    public const OWNERSHIP_STRONG = 'strong';
+
+    /**
+     * Weak ownership uses 'cascade_duplicates' configuration to determine relationships
+     */
+    public const OWNERSHIP_WEAK = 'weak';
```
ChloeSartorelli: This feels like we're just working around a limitation in the current ownership model... we probably need to go back to the drawing board and rethink what `owns` does if we think we need two different types of "ownership"

src/RecursiveStagesService.php
```diff
@@ -0,0 +1,166 @@
...
+
+    /**
+     * Determine if content differs on stages including nested objects
+     *
+     * @param DataObject $object
+     * @param string $mode
+     * @return bool
+     */
```
ChloeSartorelli:
```suggestion
```
This PHPDoc doesn't differ from the interface, so don't override it to avoid getting out of sync.

src/RecursiveStagesService.php
```diff
@@ -0,0 +1,166 @@
...
+
+    /**
+     * Find all identifiers for owned objects
+     *
+     * @param DataObject $object
+     * @param string $mode
+     * @param string $stage
+     * @return array
+     */
+    protected function findOwnedIdentifiers(DataObject $object, string $mode, string $stage): array
```
ChloeSartorelli: What is an "identifier"?

src/RecursiveStagesService.php
```diff
@@ -0,0 +1,166 @@
...
+     *
+     * "Weakly owned" objects could be looked up via "owns" setting
+     * Such objects can exist even without the owner objects as they are often used as shared objects
+     * managed independently of their owners
+     *
+     * @param DataObject $object
+     * @param string $mode
+     * @return array
+     */
+    protected function findOwnedObjects(DataObject $object, string $mode): array
```
ChloeSartorelli: `RecusrivePublishable` defined a `findOwned` method already - does this do the same as that method? If not, is it trying to work around some limitation in that method? And if not... what is the intention of this method which the existing one doesn't cover?

src/RecursiveStagesService.php
```diff
@@ -0,0 +1,166 @@
...
+     */
+    protected function checkNeedPublishingItem(DataObject $item): bool
+    {
+        if ($item->hasExtension(Versioned::class)) {
+            /** @var $item Versioned */
+            return !$item->isPublished() || $item->stagesDiffer();
+        }
+
+        return false;
+    }
```
ChloeSartorelli: This definitely feels like it belongs on `RecursivePublishable` or `Versioned`

GuySartorelli (changes requested):
I haven't done a full review - but at a glance I have at the very least some questions to clarify the approach and intent.

GuySartorelli (commented):
Why was `cascade_duplicates` chosen as the marker for the "weak" ownership, and not `cascade_deletes`? Or a combination of those?

src/RecursiveStagesService.php
ChloeSartorelli: Why was `cascade_duplicates` chosen as the marker for the "weak" ownership, and not `cascade_deletes`? Or a combination of those?

_config/versionedownership.yml
mfendeksilverstripe: Fair enough, can update this config.

src/RecursiveStagesInterface.php
mfendeksilverstripe: Yup, I believe these doc-blocks were still relevant in 2021. Will remove them.

src/RecursiveStagesInterface.php
mfendeksilverstripe: It's a dirty model state that traverses down to owned models.

src/RecursiveStagesInterface.php
mfendeksilverstripe: I believe this was a change that @maxime-rainville requested. My impression is that it's a good practice to provide an interface along with the service to allow better extensibility.

src/RecursiveStagesService.php
mfendeksilverstripe: I think the intention is to structure the code in a better way to avoid dumping all the code into a single massive file like the `Versioned.php` is.

src/RecursiveStagesInterface.php
ChloeSartorelli: I think my confusion here is more.... why is there a separate service in the first place? Duplicate of https://github.com/silverstripe/silverstripe-versioned/pull/328#discussion_r1349831737 though so I'll mark this as resolved

src/RecursiveStagesService.php
mfendeksilverstripe: I agree with this assessment. I would prefer to have a new relation configuration specific to only this need so we don't have to rely on other cascade_deletes/duplicates. We should still aim to provide some sort of fix until a better long term solution can be delivered.

src/RecursiveStagesService.php
ChloeSartorelli: While I agree in principle, I don't think this is the way to go about it. I think Versioned needs to be refactored at a more holistic level - if we take this precedent, every new feature we add will add a new interface and a new service which only exist to avoid dumping stuff in `Versioned`.

I think for this PR we should put it directly into `Versioned` or `RecursivePublishable` (whichever is appropriate for this functionality), and we can look at refactoring those classes more holistically at a later date.

src/RecursiveStagesService.php
mfendeksilverstripe: I think this is a unique identifier for a DataObject. Looking at the code it; currently using `implode('_', [$object->baseClass(), $object->ID]);` but we should probably be able to use `getUniqueKey()` instead.

src/Versioned.php
```diff
@@ -1997,6 +1998,23 @@ public function stagesDiffer()
...
+     * @param string $mode "strong" or "weak"
+     * @return bool
+     */
+    public function stagesDifferRecursive(string $mode = RecursiveStagesService::OWNERSHIP_STRONG): bool
+    {
+        /** @var RecursiveStagesInterface $service */
+        $service = Injector::inst()->get(RecursiveStagesInterface::class);
+
+        return $service->stagesDifferRecursive($this->owner, $mode);
+    }
```
ChloeSartorelli: Currently this PR doesn't introduce actually calling this at any point... what is the intention as far as using this new API? Are there places in core that should be using this? Or is this purely for bespoke projects to leverage, and if so, how do you anticipate them using it?

src/RecursiveStagesService.php
```diff
@@ -0,0 +1,166 @@
...
+     * @param string $mode
+     * @param string $stage
+     * @return array
+     */
+    protected function findOwnedIdentifiers(DataObject $object, string $mode, string $stage): array
+    {
+        $ids = Versioned::withVersionedMode(function () use ($object, $mode, $stage): array {
+            Versioned::set_stage($stage);
+
+            $object = DataObject::get_by_id($object->ClassName, $object->ID);
```
ChloeSartorelli: It feels a little odd that you're overriding the object with itself. I gather that the intention is that the assignment of the second object reference is that it's the object for the given stage. With that in mind, please make this change which makes it a little easier to see what's happening at a glance

```suggestion
        $ids = Versioned::withVersionedMode(function () use ($object, $mode, $stage): array {
            Versioned::set_stage($stage);

            $stagedObject = DataObject::get_by_id($object->ClassName, $object->ID);
```

src/RecursiveStagesService.php
mfendeksilverstripe: Good point - I think this can be changed to use the `findOwned` instead.

src/RecursiveStagesService.php
```diff
@@ -0,0 +1,166 @@
...
+            if ($object === null) {
+                return [];
+            }
+
+            $items = [$object];
+            $ids = [];
+
+            while ($object = array_shift($items)) {
+                $ids[] = implode('_', [$object->baseClass(), $object->ID]);
+                $relatedObjects = $this->findOwnedObjects($object, $mode);
```
ChloeSartorelli: We will have already called `findOwnedObjects` for this object in `stagesDifferRecursive()` won't we? It feels a bit weird to be doing it _again_ here. Is there a way we can pass in the existing list of owned objects, or avoid finding them in `stagesDifferRecursive()` altogether?

src/RecursiveStagesService.php
mfendeksilverstripe: Weak ownership is intended for model which can exist without the owner like assets (but we still want to publish them when owner is present and published). Strong ownership is used for things like carousel items, that can't exist without its parent (carousel block).

src/RecursiveStagesService.php
```diff
@@ -0,0 +1,166 @@
...
+
+        // compare deleted content
+        $draftIdentifiers = $this->findOwnedIdentifiers($object, $mode, Versioned::DRAFT);
+        $liveIdentifiers = $this->findOwnedIdentifiers($object, $mode, Versioned::LIVE);
+
+        return $draftIdentifiers !== $liveIdentifiers;
+    }
+
+    /**
+     * Find all identifiers for owned objects
```
ChloeSartorelli:
```suggestion
     * Get unique identifiers for all owned objects
```

src/RecursiveStagesService.php
ChloeSartorelli: Ahh okay. I see after reading the code that it's just unique identifiers for the records to do a comparison against. Yeah probably better to use `getUniqueKey()` here now that that's an option.

mfendeksilverstripe:
Hey @maxime-rainville I would like to get some clarity on what to do with this PR. @GuySartorelli provided me with very good feedback which I'm happy to action. However, from my past experience these PRs even after sinking a lot of effort into them and addressing all the feedback might not get merged because of lack of alignment on how we actually want to approach this problem.

This is my understanding of the situation:

* we agree that we have a systemic issue with how page changes are detected in the CMS as the button states don't really work with nested owned models (such as content blocks)
* this feature is really important to content authors as it prompts them to publish changes
* there are also other components that are impacted by this, for example https://github.com/silverstripe/silverstripe-cms/issues/2892

From the feedback I see two distinct groups of changes:

* long term solution which redefines how ownership is managed - this is CMS 6 stuff the earliest
* workaround solution for CMS 4 / 5 - something that we can deliver now

Please let me know which of these (if any) should this PR focus on.

src/RecursiveStagesService.php
ChloeSartorelli: In that case, shouldn't we use `cascade_deletes` (i.e. the child record literally cannot exist without its parent - it gets deleted (or archived) if the parent is deleted (or archived)) for the strong ownership? As per [documentation](https://docs.silverstripe.org/en/5/developer_guides/model/versioning/#defining-ownership-between-related-versioned-dataobjects), `owns` triggers a cascading publish, but _not_ a cascading delete. If what we care about here is whether the child can exist without the parent, `owns` doesn't seem to be relevant at all.

If I'm correct about that, then we shouldn't be calling this "ownership" at all. I'm not sure what we _should_ be calling it, though.

emteknetnz (commented):
The overall approach seems logical, recursively traverse down the $owns (or other ownership config) tree and looks for unpublished changes

So the intention with this PR is to set the initial state of the Page Publish button to the results of `Versioned::stagesDifferRecursive()`?

Just to confirm, is the main issue you're having with the Page Publish button being out of sync related its initial state on page load, or is it it after changes are made i.e. the javascript changed detection doesn't work on nested things?

mfendeksilverstripe:
Hey @emteknetnz This change is aiming to address the issues related to initial page state only. Yes, we would need to update several places in the CMS UI where `stagesDiffer()` is used and replace it with `stagesDifferRecursive()` to achieve that. Most notable places would be the publish button state but also other components such as Cancel draft changes action.

I recognise that there are other issues with the publish button state and cancel draft changes action which are caused by user interaction. Most notable case is the usage of block level edit in the elemental inline editor. These issues are out of scope of this PR though.

emteknetnz:
> Cancel draft changes action

What's a cancel draft changes action?

mfendeksilverstripe:
> What's a cancel draft changes action?

@emteknetnz Please check on https://github.com/silverstripe/silverstripe-cms/issues/2892 , I left more details there.

emteknetnz:
OK cool thanks, so it's intended use case is limited to a couple of things that are both calculated in SiteTree::getCMSActions().  Hopefully should be easy to rollout then

> From the feedback I see two distinct groups of changes:
>   long term solution which redefines how ownership is managed - this is CMS 6 stuff the earliest
>   workaround solution for CMS 4 / 5 - something that we can deliver now

I think treating this PR as a "workaround" PR is fine, IMO it's not really a workaround it's almost a bugfix to get the existing architecture (which was probably designed pre-elemental) to work as it should.  My only real concern is that there will be a performance hit on this, particularly if you have a large number of blocks. I'm not sure how impactful this would be in practice though, it may be negligible.

I'd agree with Guy that this should just go straight into Versioned rather than having a separate service class.  While I get that the intention is to split things off in to easily digestable pieces, it just make it things inconsistent because all the existing stuff in Versioned just stays where it is. I'd much rather just keep adding to Versioned for now.

mfendeksilverstripe:
Agreed to remove the `cascade_duplicates` based traversal - we will only use `owns` which is consistent with `publishRecursive()`

* add test for versioned model > non-versioned model > versioned model

mfendeksilverstripe:
@GuySartorelli I've pushed up new changes as per our discussion, please review.

cc: @maxime-rainville

src/RecursiveStagesExtension.php
ChloeSartorelli, line 1: Please just add this directly to `RecursivePublishable` - we don't need a whole separate extension for this.

src/RecursiveStagesService.php
```diff
@@ -0,0 +1,187 @@
...
+        $service->flushCachedData();
+    }
+
+    /**
+     * Determine if content differs on stages including nested objects
+     * This method also supports non-versioned models to allow traversal of hierarchy
+     * which includes both versioned and non-versioned models
+     *
+     * @param DataObject $object
+     * @return bool
```
ChloeSartorelli:
```suggestion
```
No need to include the params and return type since they're already part of the method signature, and the PHPDoc here isn't providing any additional context.

Please also remove these from all other PHPDoc you've added that aren't providing more context than the method signature does.

src/RecursiveStagesService.php
```diff
@@ -0,0 +1,187 @@
...
+     * @throws Exception
```
ChloeSartorelli: This doesn't provide enough information - it should say what circumstances result in an exception being thrown. Also, if there's a more specific extension that gets thrown, use that as the hint instead of just `Extension`. If different exception classes are used based on the scenario, add each with their own context.

Please also add additional context to all other method PHPDocs that you've added a `@throws` annotation to.

ChloeSartorelli: Alternatively, if this is a scenario that can be handled with a try/catch block in this method, please handle it that way.

src/RecursiveStagesService.php
ChloeSartorelli: Please just put this code directly into `Versioned`. It doesn't seem like any of this code is being called from anything that isn't `Versioned` with exception of the `reset()` method, which is static anyway.

src/RecursiveStagesService.php
```diff
@@ -0,0 +1,187 @@
...
+
+    /**
+     * This lookup will attempt to find "owned" objects
+     * This method uses the 'owns' relation, same as @see RecursivePublishable::publishRecursive()
+     *
+     * @param DataObject|RecursivePublishable $object
+     * @return array
+     * @throws Exception
+     */
+    protected function getOwnedObjects(DataObject $object): array
```
ChloeSartorelli, line 155:
```suggestion
    private function getOwnedObjects(DataObject $object): array
```
There's no need for this to be exposed as a protected method at this stage. If someone wants to extend this functionality in the future we can consider their use case at that time, and decide if we want to open this up to be part of our [public API](https://docs.silverstripe.org/en/5/contributing/code/#definition-of-public-api).

Please do this for any other protected methods you've added

src/RecursiveStagesService.php
```diff
@@ -0,0 +1,187 @@
...
+    {
+        if (!$object->hasExtension(RecursivePublishable::class)) {
+            return [];
+        }
+
+        // Add versioned stage to cache key to cover the case where non-versioned model owns versioned models
+        // In this situation the versioned models can have different cached state which we need to cover
+        $cacheKey = sprintf('%s-%s', $object->getUniqueKey(), Versioned::get_stage());
```
ChloeSartorelli:
```suggestion
        $cacheKey = $object->getUniqueKey() . '-' . Versioned::get_stage();
```
This is easier to scan-read at a glance

src/RecursiveStagesService.php
```diff
@@ -0,0 +1,187 @@
...
+
+        if (!array_key_exists($cacheKey, $this->ownedObjectsCache)) {
+            $this->ownedObjectsCache[$cacheKey] = $object
+                // We intentionally avoid recursive traversal here as it's not memory efficient,
+                // stack based approach is used instead for better performance
```
ChloeSartorelli: There's no stack being created here is there? Is this comment meant to be in one (or both) of the methods that call this one?

src/RecursiveStagesService.php
```diff
@@ -0,0 +1,187 @@
...
+     * Non-versioned models are supported
+     *
+     * @param DataObject $object
+     * @return bool
+     */
+    protected function isModelDirty(DataObject $object): bool
+    {
+        if ($object->hasExtension(Versioned::class)) {
+            /** @var $object Versioned */
+            return !$object->isPublished() || $object->stagesDiffer();
```
ChloeSartorelli: Why specifically `!isPublished()`? For example take a look at `Versioned::isModifiedOnDraft()`, which according to its PHPDoc "Compares current draft with live version, and returns true if these versions differ, meaning there have been unpublished changes to the draft site."

Seems like we should just be doing this instead?
```suggestion
            return $object->isModifiedOnDraft();
```

ChloeSartorelli: For that matter, do we need this method at all? The condition for the one place it's called could be `if ($object->hasExtension(Versioned::class) && $object->isModifiedOnDraft())`

src/RecursiveStagesService.php
```diff
@@ -0,0 +1,187 @@
...
+     * Determine if content differs on stages including nested objects
+     * This method also supports non-versioned models to allow traversal of hierarchy
+     * which includes both versioned and non-versioned models
+     *
+     * @param DataObject $object
+     * @return bool
+     * @throws Exception
+     */
+    public function stagesDifferRecursive(DataObject $object): bool
+    {
```
ChloeSartorelli, line 50:
```suggestion
    public function stagesDifferRecursive(?DataObject $object = null): bool
    {
        if (!$object) {
            $object = $this->owner;
        }
```
You'll need to make this adjustment when moving this into `Versioned` so we don't have to call it like `$obj->stagesDifferRecursive($obj)`

src/RecursiveStagesService.php
```diff
@@ -0,0 +1,187 @@
...
+        $cacheKey = $object->getUniqueKey();
+
+        if (!array_key_exists($cacheKey, $this->stagesDifferCache)) {
+            $this->stagesDifferCache[$cacheKey] = $this->determineStagesDifferRecursive($object);
```
ChloeSartorelli, line 54: Instead of declaring a whole new `determineStagesDifferRecursive` method, please just move that logic into this method and either call this method recursively on children, or if you have a clear reason why we shouldn't use recursion (which should be explained clearly in a comment in this method), at least check the cache of each child.
That will have the following benefits:
- Don't have the extra method, which doesn't need to be there in this case
- If we called `stagesDifferRecursive()` on one of the children already, it will already be cached. So when we call `stagesDifferRecursive()` on the parent, it doesn't have to recalculate the child data as well.

src/RecursiveStagesService.php
```diff
@@ -0,0 +1,187 @@
...
+     * @param DataObject $object
+     * @return bool
+     * @throws Exception
+     */
+    protected function determineStagesDifferRecursive(DataObject $object): bool
+    {
+        if (!$object->exists()) {
+            // Model hasn't been saved to DB, so we can just bail out as there is nothing to inspect
+            return false;
+        }
```
ChloeSartorelli, line 70: Is there ever a scenario where the owned child of `$object` in an `UnsavedRelationList` would return true? If so, we should probably account for it

src/RecursiveStagesService.php
```diff
@@ -0,0 +1,187 @@
...
+
+        $models = [$object];
+
+        // Compare existing content (perform full ownership traversal)
+        while ($model = array_shift($models)) {
```
ChloeSartorelli: Why are we doing this in a loop instead of as a recursive call?

src/RecursiveStagesService.php
```diff
@@ -0,0 +1,187 @@
...
+
+    /**
+     * Get unique identifiers for all owned objects, so we can easily compare them
+     *
+     * @param DataObject $object
+     * @param string $stage
+     * @return array
+     * @throws Exception
+     */
+    protected function getOwnedIdentifiers(DataObject $object, string $stage): array
```
ChloeSartorelli, line 114: We have an in-memory cache for owned objects, should we have one for owned identifiers (per stage) as well?

src/RecursiveStagesService.php
```diff
@@ -0,0 +1,187 @@
...
+            if ($stagedObject === null) {
+                return [];
+            }
+
+            $models = [$stagedObject];
+            $identifiers = [];
+
+            while ($model = array_shift($models)) {
+                // Compose a unique identifier, so we can easily compare models
+                // Note that we intentionally use base class here, so we can cover the situation where model class changes
```
ChloeSartorelli: > Note that we intentionally use base class here, so we can cover the situation where model class changes

Can you please explain that a bit further for me? How would we get into a situation where the model class has changed, and how does using the base class here help us avoid problems related to that situation?

src/RecursiveStagesService.php
```diff
@@ -0,0 +1,187 @@
...
+                $identifiers[] = implode('_', [$model->baseClass(), $model->ID]);
```
ChloeSartorelli:
```suggestion
                $identifiers[] = implode('_', [$model->baseClass(), $model->getUniqueKey()]);
```
Unless there's a reason why we wouldn't want the extra identifying data from `getUniqueKeyComponents()`, we should probably use `getUniqueKey()`.

If there is a good reason not to do that, please add a comment in the code, so we don't swap it out in the future.

src/RecursiveStagesService.php
```diff
@@ -0,0 +1,187 @@
...
+                $relatedObjects = $this->getOwnedObjects($model);
+                $models = array_merge($models, $relatedObjects);
```
ChloeSartorelli, line 140: If we have a `many_many` relationship, where each side has `$owns` config to their side of the relation, will this result in an infinite loop?

GuySartorelli (changes requested):
Regarding `@throws` in phpdocs - I notice you have added that in many methods which themselves don't actually throw the exception. That doesn't tend to be the way `@throws` is used in Silverstripe CMS.

Can you please provide some information about why those are being added in this way? With citations preferably if it's a common practice.

I have added some notes below on the assumption that you're right that they should be here - but I'd like to understand _why_ they should be there if they indeed should.

For `NotFoundExceptionInterface` specifically, I'm assuming that's coming from the injector. Please don't add a throws statement for that, especially when we're explicitly defining the service that we're requesting. Otherwise we'd have to add this throws statement to basically every method across all the supported modules, since injector is used for basically everything.

src/RecursiveStagesService.php
mfendeksilverstripe: Happy to remove these. These docblocks are auto generated by my IDE.

src/RecursiveStagesService.php
mfendeksilverstripe: Good point, will make an update.

src/RecursiveStagesService.php
mfendeksilverstripe: There is, will update the comment.

src/RecursiveStagesService.php
mfendeksilverstripe: This isn't needed, will replace this with `ClassName`

src/RecursiveStagesService.php
mfendeksilverstripe: Dumping the new code into the `Versioned` is something I definitely view as a strict downgrade with the two main reasons:

### Code readability

Dumping more code into a huge file like `Versioned` will make the code less maintainable and it goes agains the general practice of "modular code" where we have several smaller pieces of code that have different responsibilities.

### Code extensibility

One of the aims of providing a separate service that has the responsibility over the `stagesDifferRecursive()` is to allow easier extensibility. Developers can either sublclass the existing service (hence the protected methods) or create their own service with the provided interface (hence the interface). Having a separate service makes it easier to subclass as the subclass scope is smaller.

Additionally, this provides greater flexibility when subclassing as the subclass scope doesn't impact `Versioned`. This minimises the cases where modules may want to provide their own subclasses for some core extension such as Versioned snapshots resulting in subclass conflicts which are painful to deal with resulting in a complex subclass hierarchy.

This aligns with the "composition over in inheritance" principle. On top of that we already have precedence for the modular service driven approach. Some examples listed below:

* [DataObject::getUniqueKey()](https://github.com/silverstripe/silverstripe-framework/blob/5/src/ORM/DataObject.php#L4448)
* [CopyLocaleAction::handleAction](https://github.com/tractorcow-farm/silverstripe-fluent/blob/master/src/Forms/CopyLocaleAction.php#L90)

If this change-set breaks the current semver constraints I suggest to review the release target for this feature as opposed to the code downgrade.

ChloeSartorelli: I understand that, and in a future major release we will likely be revisiting the `Versioned` class and versioning functionality holistically - but at the moment adding a separate service for this piece of functionality is a departure from the way the versioning codebase is set up. It sets a precedent that we should add a new interface and a new service class for each new piece of versioning functionality.

This service is unlikely to be used anywhere other than inside the versioned class directly.
It's also frankly very unlikely to be customised in projects.

Both Steve (see https://github.com/silverstripe/silverstripe-versioned/pull/328#issuecomment-1762431609) and I agree that this code should go directly into `Versioned` for now as this is consistent with the way this module is set up and will be easier to maintain for the time being. We can then look at `Versioned` holistically for a future major release to identify what functionality should be separated out, and _how_ it should be separated out.

src/RecursiveStagesService.php
mfendeksilverstripe: See my other comment about protected methods.

src/RecursiveStagesService.php
mfendeksilverstripe: @maxime-rainville What are your thoughts on this? I'm still reluctant to downgrade the code.

Also, I would like to challenge the narrative about specific code being unlikely to be customised. In general, customisation ability is one of the notable shortcomings of the core modules when looking at the projects I worked on for last 5 years. Customisation can be functional (application code) but also non-functional (tests).

src/RecursiveStagesService.php
mfendeksilverstripe: Good point, will review this scenario

src/RecursiveStagesService.php
mfendeksilverstripe: This is pending the overall discussion about whether we keep the service or not

src/RecursiveStagesService.php
mfendeksilverstripe: I tend to avoid "true" recursion for performance reasons, instead using "stack based" recursion. Functionally the same, but the latter doesn't cause memory spikes and has overall faster execution time due to smaller stack depth. Will update code comments related to this.

src/RecursiveStagesService.php
mfendeksilverstripe: Nice catch - will add a case for this.

mfendeksilverstripe: After some investigation it looks like this case is properly handled in `mergeRelatedObjects()` so no additional change should be needed.

src/RecursiveStagesService.php
mfendeksilverstripe: This is to avoid" true" recursion for performance reasons. Using stack based recursion instead.

src/RecursiveStagesService.php
mfendeksilverstripe: As far as I can see the benefit of this cache would be minimal as other cache instances should cover all the cases.

src/RecursiveStagesService.php
mfendeksilverstripe: Will update the comment to make this more clear.

src/RecursiveStagesService.php
mfendeksilverstripe: Have updated code comments to reflect this.

src/RecursiveStagesService.php
mfendeksilverstripe: Added checks that prevent infinite loops

src/RecursiveStagesService.php
maxime-rainville: I like the "composition over in inheritance" principle. In the long run, we should aim to implement it this way. 

If we were starting from scratch, we probably would want a "RecursiveOwnership" service that handles things like recursive publishing, recursive rollbacks and recursive stage differ check. So it feels a bit weird having a service just for the stage check.

I'm keen to add a card to the CMS 6 milestone to re-factor Versioned to rely more on composition.

My gut instinct would be to put `stagesDifferRecursive` logic in `RecursivePublishable` because that's where I would expect the most coupling. But I don't have any beef with the way @mfendeksilverstripe has chosen to implement it. 

This seems like a pretty trivial concern that's not worth arguing over all that much. Either way, we are adding a bit of short term technical debt.

mfendeksilverstripe: Thanks for your feedback @maxime-rainville I'll leave the implementation as is.

@GuySartorelli can you please do another round of review and let me know if there are any more concerns that I should address?

ChloeSartorelli: It makes no sense to me whatsoever to have a service for this one method - surely we want the `stagesDiffer` and the `stagesDifferRecursive` code to be in the same place since they're effectively the same concern. 

But sure. We can have this weird architecture for now and fix it in a major release later. I'll continue the review on Tuesday.

src/RecursiveStagesService.php
ChloeSartorelli: Can you please respond to this query?

src/RecursiveStagesService.php
ChloeSartorelli: I'll leave these as they are since we're keeping the service.

src/RecursiveStagesService.php
ChloeSartorelli: Is there a unit test for this?

GuySartorelli (commented):
Overall looks good. There's just a couple of clarifying questions here.

We also won't fully know if this is fit for purpose until we start on https://github.com/silverstripe/silverstripe-admin/issues/1593 but whoever picks that up can make whatever adjustments are necessary if there's anything we missed here, so I'm okay merging as-is once the questions below are answered.

src/RecursiveStagesService.php
mfendeksilverstripe: hey, sorry I missed this comment somehow. The values are passed via in-memory cache now so we don't execute the same lookup more than once.

ChloeSartorelli: Fair point. That'll work fine for the expected lifetime of this service

src/RecursiveStagesService.php
mfendeksilverstripe: We don't have a unit test for this one but I'll add it.

mfendeksilverstripe: Oh, actually there is - see `testStageDiffersRecursiveWithInvalidObject()`

mfendeksilverstripe:
Hey @GuySartorelli I believe I addressed all the feedback from you. Please hit me up if there are more things to follow up on 😉

tests/php/RecursiveStagesServiceTest.php
```diff
@@ -0,0 +1,130 @@
...
+        ],
+    ];
+
+    public function testStageDiffersRecursiveWithInvalidObject(): void
+    {
+        Versioned::withVersionedMode(function (): void {
+            Versioned::set_stage(Versioned::DRAFT);
+
+            /** @var PrimaryObject|Versioned $primaryItem */
+            $primaryItem = PrimaryObject::create();
```
ChloeSartorelli, line 47:
```suggestion
            $primaryItem = PrimaryObject::create();
            $primaryItem->Title = 'This registers as a change in DataObject::isChanged()';
```
Just to make sure - otherwise we could theoretically be seeing no changes just because there _were_ no changes, rather than because we're intentionally not checking it.

GuySartorelli (changes requested):

GuySartorelli:
Just the one requested change - then I'll be happy to merge.

tests/php/RecursiveStagesServiceTest.php
mfendeksilverstripe: All good - I've added your change. Feel free to merge once tests pass.

GuySartorelli (approved):
Looks good, thanks!

PR #336, opened 2021-07-07
FIX Avoid inferring model type

chillu:
Since Versioned->versions() results in an ArrayList, it triggers database queries.
The database isn't always available when the schema is built (e.g. on deployment and CI environments).

Relies on: https://github.com/silverstripe/silverstripe-graphql/pull/395
Context: https://github.com/silverstripe/silverstripe-graphql/issues/388

## Tasks

 * [ ] Fix setting of resolvedModelClass


GuySartorelli:
Since this is an old draft, and I haven't seen instances of this being a problem in my time maintaining Silverstripe CMS, I'm going to close this.

If someone can show me a recent case where this has been a problem, we can look into opening an issue and taking it from there.

PR #352, opened 2022-02-15
Don't run queries when trying to retrieve versions for unsaved records

maxime-rainville:
Right now if you call `allVersions` on a singleton DataObject, it will try to run a query ... which is nonsensical. This is causing grievances to our graphql `--no-database` build steps.

# Parent issue
- ~~https://github.com/silverstripe/silverstripe-graphql/issues/445~~
- #353

### DIFF
--- a/tests/php/VersionedTest.php
+++ b/tests/php/VersionedTest.php
@@ -893,6 +893,15 @@ class VersionedTest extends SapphireTest
         $this->assertEquals($extraFields, ['2005', '2007', '2009'], 'Additional version fields returned');
     }
 
+    public function testAllVersionsOnUnsavedRecord()
+    {
+        $newPage = new VersionedTest\Subclass();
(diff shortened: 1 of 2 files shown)


tests/php/VersionedTest.php
```diff
@@ -893,6 +893,15 @@ public function testAllVersions()
         $this->assertEquals($extraFields, ['2005', '2007', '2009'], 'Additional version fields returned');
     }
 
+    public function testAllVersionsOnUnsavedRecord()
+    {
+        $newPage = new VersionedTest\Subclass();
```
maxime-rainville, line 898: Those tests passed with the old logic.

maxime-rainville (commented):
I'm going to call this a performance improvement that's worth doing ... irrespective of the NullDatabase work.

emteknetnz (approved):

PR #354, opened 2022-03-15
FIX Return correct changed fields when fromRecord is null

GuySartorelli:
Ran into `__clone method called on non-object` when using the assets admin history view. Elsewhere in this class there are checks against this variable being falsey, so this is a consistent way to resolve it and allows the history view to render without issues.

### DIFF
--- a/src/DataDifferencer.php
+++ b/src/DataDifferencer.php
@@ -258,7 +258,11 @@ class DataDifferencer extends ViewableData
      */
     public function changedFieldNames()
     {
-        $diffed = clone $this->fromRecord;
+        if ($this->fromRecord) {
+            $diffed = clone $this->fromRecord;
+        } else {
+            $diffed = clone $this->toRecord;
+        }
         $fields = array_keys($diffed->toMap());
 
         $changedFields = [];
@@ -267,7 +271,7 @@ class DataDifferencer extends ViewableData
             if (in_array($field, $this->ignoredFields)) {
                 continue;
             }
-            if ($this->fromRecord->$field != $this->toRecord->$field) {
+            if (!$this->fromRecord || $this->fromRecord->$field != $this->toRecord->$field) {
                 $changedFields[] = $field;
             }
         }


michalkleiner:
Would branch from and target 1.10 since this is a bugfix.

GuySartorelli:
@michalkleiner 
Fair point! I was just kinda rushing it through as I worked on other things so I didn't really think about it, sorry.

Branched from and targetting 1.10 now.

michalkleiner:
@GuySartorelli is it the first record when this happens, do you know? There are calls to clone the toRecord in the class as well, I wonder whether we need to protect that as well or the fromRecord is the only one that can _legitimately_ be null.

GuySartorelli:
> is it the first record when this happens, do you know?

I haven't checked to confirm, but my assumption is that's the case - the PHPdoc comment in the constructor even points out that a numm from record is expected for a new record
https://github.com/silverstripe/silverstripe-versioned/blob/7cedcf6cb26afc5e0babe80deec83159f379196d/src/DataDifferencer.php#L56

> There are calls to clone the toRecord in the class as well, I wonder whether we need to protect that as well or the fromRecord is the only one that can _legitimately_ be null.

I think that although the toRecord has a null default argument, the expectation is for it to never be null during execution. I didn't encounter any situations where it was null when I was playing around, and there aren't any protections in this class against null toRecords.

GuySartorelli:
I have just taken a closer look at what's happening throughout this class though and it looks like there's a bit more work to be done though - elsewhere if there's no fromRecord then all the fields in the toRecord are considered changed, so we should probably do that here to be consistent.

GuySartorelli:
There we go. @michalkleiner I think this is a more complete fix.

michalkleiner:
Since we're poking this already, @GuySartorelli, I think we should refactor the `changedFieldNames` and `ChangedFields` methods so that they both use a private or protected method of getting the diffed fields (seems like doesn't even need the cloned object tbh) and then in both those methods just iterate over it and do the work needed — [push the field name into an array](https://github.com/silverstripe/silverstripe-versioned/blob/28e40806679f7edd1f512f79b4a108170b58a832/src/DataDifferencer.php#L275) or, respectively, [push the diff data to an array list](https://github.com/silverstripe/silverstripe-versioned/blob/28e40806679f7edd1f512f79b4a108170b58a832/src/DataDifferencer.php#L228).

That seems like the most robust approach where we remove duplicate code.

GuySartorelli:
Great minds think alike! I was thinking of doing some refactoring for this as a separate PR, but if you're happy for it to be in the same PR (but a separate commit obviously) I'll try get some time to refactor this class tomorrow. I thought though that since refactoring is an `ENH` commit that would need to be targeting `1` so it gets a minor release?

GuySartorelli:
I'm not comfortable mixing a refactor commit with the fix commit. If someone wonders why changes were made in the future and look back at the commit history I'd rather it be clear that the refactoring was done for ease of maintenance and the fix was done to fix an actual issue. They're two separate concerns which should be committed separately.

michalkleiner:
It's not a refactor where you'd completely change things, so I'm ok with it. It's abstracting one method's mechanism to be reused in another to fix a problem that the first method doesn't suffer with.
If you still want to do the fix separately, please replicate the approach from `ChangedFields` in `changedFieldNames` i.e. no object cloning and same convention/naming etc.

GuySartorelli:
I just don't see that being part of the fix. I've already committed a fix, and it is consistent with the `diffedData` method in this class which also uses cloning.
In my opinion any change to _how_ the method does what it does should be in a separate commit (it's not part of the _fix_, which is what my current commit is for) and at this stage I'd rather do that as a completely separate PR like I had originally planned to do. If you'd rather have additional changes in the same commit I have no problem with you raising a PR to replace this one.

michalkleiner:
The difference in the `diffedData` method is that it combines fields from both from and to records if from record exists, so I'd say the fix should be closer to how `ChangedFields` method works. I guess the cloning isn't a dealbreaker, but is probably not needed since nothing is changing on the object. I'll merge this as is to let you do the refactor in a follow up PR, which won't be able to be from the minor version, but we can live with that.

michalkleiner:
Tagged `1.10.1` on `1.10` branch and merged it up to `1` branch.

PR #355, opened 2022-03-21
ENH: Refactor DataDifferencer to reduce duplicate code.

GuySartorelli:
Also remove cloning where it isn't necessary.

@michalkleiner per our discussion starting in [this comment](https://github.com/silverstripe/silverstripe-versioned/pull/354#issuecomment-1068788809) on #354.

I didn't make a separate method in the end because the list of diffed fields is exactly what is returned by `changedFieldNames`.

### DIFF
--- a/src/DataDifferencer.php
+++ b/src/DataDifferencer.php
@@ -19,7 +19,7 @@ use SilverStripe\View\ViewableData;
  * <code>
  * $fromRecord = Versioned::get_version('SiteTree', $pageID, $fromVersion);
  * $toRecord = Versioned::get_version('SiteTree, $pageID, $toVersion);
- * $diff = new DataDifferencer($fromRecord, $toRecord);
+ * $diff = DataDifferencer::create($fromRecord, $toRecord);
  * </code>
  *
  * And then it can be used in a number of ways.  You can use the ChangedFields() method in a template:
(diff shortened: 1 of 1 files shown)


src/DataDifferencer.php
```diff
@@ -208,45 +208,32 @@ protected function getObjectDisplay($object = null)
      */
     public function ChangedFields()
     {
+        $base = $this->fromRecord ?: $this->toRecord;
         $changedFields = new ArrayList();
```
michalkleiner, line 212:
```suggestion
        $changedFields = ArrayList::create();
```

src/DataDifferencer.php
```diff
@@ -208,45 +208,32 @@ protected function getObjectDisplay($object = null)
...
                 }
-                $changedFields->push(new ArrayData([
-                    'Name' => $field,
-                    'Title' => $base->fieldLabel($field),
-                    'Diff' => $fieldDiff,
-                    'From' => $this->fromRecord ? $this->fromRecord->$field : null,
-                    'To' => $this->toRecord ? $this->toRecord->$field : null,
-                ]));
             }
+            $changedFields->push(new ArrayData([
```
michalkleiner, line 230:
```suggestion
            $changedFields->push(ArrayData::create([
```

michalkleiner (commented):
Nice, that looks tidier.
Could you also please update to use the `::create` syntax since we're touching this.

GuySartorelli:
Yeah can do. I noticed that and only left it alone because one of the things I aim on doing is a global overhaul of replacing `new` with `create` in all the repositories. But I have no problem doing this one as part of this PR :)

GuySartorelli:
I've replaced instances of the `new` keyword in this class with dependency injection for `Injectable` classes.

michalkleiner (approved):

PR #358, opened 2022-04-11
ENH Various fixes for PHP 8.1 compatibility

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10250

### DIFF
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -655,6 +655,9 @@ class Versioned extends DataExtension implements TemplateGlobalProvider, Resetta
         // We can ONLY promote a filter on the primary key of the base table. Any other conditions will make the
         // version returned incorrect, as we are filtering out version that may be the latest (and correct) version
         foreach ($baseQuery->getWhere() as $condition) {
+            if (is_object($condition)) {
+                continue;
+            }
             $conditionClause = key($condition);
             // Pull out the table and field for this condition. We'll skip anything we can't parse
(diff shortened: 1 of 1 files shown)


GuySartorelli:
Gonna need a bit of context before I can approve this one - what class were the objects that were coming through? [`key()`](https://www.php.net/manual/en/function.key.php) and [`reset()`](https://www.php.net/manual/en/function.reset.php) take an array or an object - what was the deprecation warning for this one?

emteknetnz:
In php8.1, passing in an object was deprecated
https://www.php.net/manual/en/function.key.php#refsect1-function.key-changelog

From memory it was mostly an array coming in that was the key was the 'WHERE' portion of an sql clause

When it was an object coming in, it was a `DataQuery_SubGroup`.  However in this context we can ignore it because it's a subquery being passed in, and based on the inline comments above this block

```
        // Parse "where" conditions to find those appropriate to be "promoted" into an inner join
        // We can ONLY promote a filter on the primary key of the base table. Any other conditions will make the
        // version returned incorrect, as we are filtering out version that may be the latest (and correct) version
```
it doesn't seem like a the contents of a subquery should be used to 'promote' something to an inner join

it also would fail this test pre 8.1:

if (preg_match('/^"([^"]+)"\."([^"]+)"/', $conditionClause, $matches) !== 1) {

As `$conditionClass` is supposed to be someting like `"SiteTree"."Title"`

GuySartorelli (approved):
> it also would fail this test pre 8.1
That's what won me over :p
LGTM.

PR #361, opened 2022-05-12
Optionally disable checking canViewStage when using History tab

elliot-sawyer:
I'm having a strange issue with the History tab where on some records, it fails to show and displays a never-ending CMS logo. When I commented out these lines, the problem resolved itself. I'm unsure of the cause, but when logged in as an admin it seems to be a case of an Admin not being able to view a draft. That shouldn't be possible, but in any event, a good fix would be to wrap it in a config setting and allow a developer to override it if needed.

I considered subclassing it and replacing with Injector, but ReadVersions is invoked with `new ReadVersions` instead of `ReadVersions::create`, so that option isn't available here.

### DIFF
--- a/_legacy/GraphQL/Operations/ReadVersions.php
+++ b/_legacy/GraphQL/Operations/ReadVersions.php
@@ -4,6 +4,7 @@ namespace SilverStripe\Versioned\GraphQL\Operations;
 
 use Exception;
 use GraphQL\Type\Definition\ResolveInfo;
+use SilverStripe\Core\Config\Configurable;
 use SilverStripe\GraphQL\OperationResolver;
 use SilverStripe\GraphQL\Scaffolding\Scaffolders\ListQueryScaffolder;
 use SilverStripe\ORM\DataObject;
@@ -22,6 +23,10 @@ if (!class_exists(ListQueryScaffolder::class)) {
  */
 class ReadVersions extends ListQueryScaffolder implements OperationResolver
 {
+    use Configurable;
+
+    //disable checking of drafts on object->canViewStage
+    private static $disable_draft_visibility_check = false;
     /**
      * ReadOperationScaffolder constructor.
      *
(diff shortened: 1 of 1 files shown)


kinglozzer:
I think we should try to fix this rather than hide broken code behind a config flag. First thing that springs to mind - is `$context['currentUser']` empty?

elliot-sawyer:
When I dumped it out, it was Default Admin, member id #1

GuySartorelli:
Sounds like it's this issue: silverstripe/silverstripe-cms#2738

elliot-sawyer:
I actually raised that issue thinking the move to GraphQL 4 might resolve it, but seems unrelated. That issue occurs on a default install with a freshly created database.

elliot-sawyer:
Does it need fixing? ReadVersions is marked as deprecated, and so is the DataObjectScaffolderExtension invoking it. I would love to move to 4.11 and use GQL 4, but can't due to the issue @GuySartorelli pointed out.

Would replacing `new ReadVersions` with `ReadVersions::create` in the extensions and the tests make an acceptable PR? That way I can subclass ReadVersions with the Injector and replace it with my own logic.  It's a no-op for everybody else.

GuySartorelli:
I can't reproduce the issue.

>  It's a no-op for everybody else.
Do you mean to say you're the only person affected? I wouldn't feel right merging in a fix that only affects a single person. Do you have some custom code that is resulting in this problem? If so, it would definitely be better if you amend that custom code.

elliot-sawyer:
It actually benefits everybody else, because it makes the framework and its modules more extensible. I can't see a reason why this particular part of the modules is still using `new` when practically everything else has switched to `::create`. How is that any different from adding an extension hook to any other method?

GuySartorelli:
@elliot-sawyer Is this resolved by flushing, like with the linked issue?

elliot-sawyer:
Unfortunately not, this one happens locally and on a UAT server, reported by the client

GuySartorelli:
> I can't see a reason why this particular part of the modules is still using `new` when practically everything else has switched to `::create`

There's a whole discussion to be had about that (check out silverstripe/silverstripe-framework#10269)
The short version is: If you create a PR to replace the use of the `new` keyword with using injection to instantiate a class that uses the `Injectable` trait, it will probably get merged.

I think we should probably close _this_ PR though for reasons already mentioned in other comments above. Feel free to reopen this PR if you have objections to that (and can respond to the above reasoning).

elliot-sawyer:
Long story short: I ran into issues with this portion of the query because my HomePage was extending SiteTree instead of Page

elliot-sawyer:
The docs say it should work on all SiteTree objects automatically, but it seems it only works on Page automatically. Unsure if the docs are incorrect, or if this is a bug, so raised an issue about this here: https://github.com/silverstripe/silverstripe-versioned/issues/363.

PR #364, opened 2022-06-16
FIX comparing versions when class changes (#363)

### LINKED ISSUE #363: HistoryViewer doesn't work when pagetype extends SiteTree
Context: https://github.com/silverstripe/silverstripe-versioned/pull/361

According to the docs, [history viewer should be enabled for SiteTree objects and DataObjects automatically](https://docs.silverstripe.org/en/4/developer_guides/model/versioning/#using-the-history-viewer). However, the graphql query utilised by ReadVersions is calling on Pages, and its subclasses, to render the view. This means if you have a pagetype that extends SiteTree directly, the HistoryViewer will not render with the following error returned by the /graphql endpoint

```
message: "Cannot view versions on Page", code: 0, file: "/path/to/ss411/vendor/silverstripe/versioned/src/GraphQL/Resolvers/VersionedResolver.php", … }
message	"Cannot view versions on Page"
code	0
file	"/home/elliotsawyer/Sites/ss411/vendor/silverstripe/versioned/src/GraphQL/Resolvers/VersionedResolver.php"
line	79
```

Working with @unclecheese on Slack, we determined the issue was at DataQuery line 313. For some reason, the query is getting all subclasses of Page, rather than subclasses of SiteTree. :
```
                $classNames = ClassInfo::subclassesFor($this->dataClass);
                $classNamesPlaceholders = DB::placeholders($classNames);
```

If the documentation is correct, I'd expect this to be a bug? Otherwise, should the docs be updated to say "History Viewer should be enabled for Page objects ..."?

Reproduction steps on 4.11

1. `composer create-project silverstripe/installer 4.11.0-beta1`
2. create a pagetype `class HomePage extends SiteTree` with one $db field `$db = ['Test' => 'Varchar(32)'];` and `$table_name = 'HomePage';`
3. dev/build
4. By default, Home is a "Page" type
5. View History tab, observe that it's working
6. Change Page type to "Home Page' in Settings and publish
7. View History tab, observe that it's not working

wilr:
Fixes #363

### DIFF
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -1688,7 +1688,7 @@ SQL
 
         // If we weren't definitely loaded from live, and we can't view non-live content, we need to
         // check to make sure this version is the live version and so can be viewed
-        $latestVersion = Versioned::get_versionnumber_by_stage(get_class($owner), static::LIVE, $owner->ID);
+        $latestVersion = Versioned::get_versionnumber_by_stage($this->owner->baseClass(), static::LIVE, $owner->ID);
         if ($latestVersion == $owner->Version) {
             // Even if this is loaded from a non-live stage, this is the live version
             return true;
@@ -1730,7 +1730,8 @@ SQL
             Versioned::set_stage($stage);
 
             $owner = $this->owner;
-            $versionFromStage = DataObject::get(get_class($owner))->byID($owner->ID);
+            $baseClass = DataObject::getSchema()->baseDataClass($owner);
+            $versionFromStage = DataObject::get($baseClass)->byID($owner->ID);
 
             return $versionFromStage ? $versionFromStage->canView($member) : false;
         });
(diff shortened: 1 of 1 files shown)


michalkleiner:
Does this need any migration script to update the db records from record classes to their base classes?

wilr:
@michalkleiner not that I would imagine. Every `HomePage` should already have a `SiteTree` record. Travis isn't happy but not sure if that's a pre-existing issue.

GuySartorelli:
Travis isn't happy because it's trying to install `silverstripe-installer` 4.11 - this is a known problem with the travis setup, but we're aiming to swap to github actions in the next week or so so if you're okay with waiting for that it's probably easier than fixing travis for this PR and then reverting the fix on merge-up.

kinglozzer (approved):
Tested this locally and confirmed it fixes https://github.com/silverstripe/silverstripe-versioned/issues/363 in our projects

GuySartorelli:
Force pushed to get the new CI running against this PR.

kinglozzer:
@GuySartorelli is this mergeable with the Travis error as long as the Github Actions CI builds are passing?

GuySartorelli:
Yup. The Travis error is just because it can't find the correct version of installer, which isn't something we're going to resolve for Travis. We're only keeping Travis around for now as a point of comparison, which only applies on the default (and _maybe_ 4.11) branch.

If GitHub actions is passing then that's the main thing.

GuySartorelli (approved):

GuySartorelli:
@wilr released as [1.10.2](https://github.com/silverstripe/silverstripe-versioned/releases/tag/1.10.2)

PR #368, opened 2022-08-06
API Migrate SilverStripeNagivator

GuySartorelli:
These `SilverStripeNavigatorItem` subclasses are useful for `Versioned` objects without requiring `silverstripe/cms`

See also https://github.com/silverstripe/silverstripe-cms/pull/2761

## Parent issue
- https://github.com/silverstripe/silverstripe-admin/issues/1339


GuySartorelli:
These actually belong in versioned-admin, not here. See https://github.com/silverstripe/silverstripe-versioned-admin/pull/253

PR #372, opened 2022-08-26
Rescue Master Branch Commit ENH Prefer `::class` syntax over class strings

GuySartorelli:
Rescues and extends https://github.com/silverstripe/silverstripe-versioned/commit/43be1a09235507eb8472f1dc70d5586240ef145c

There's no change to API here so it makes sense to target `1` so both major lines can benefit.

## Parent issue
- https://github.com/silverstripe/silverstripe-framework/issues/10350

### DIFF
--- a/src/GridFieldRestoreAction.php
+++ b/src/GridFieldRestoreAction.php
@@ -126,11 +126,11 @@ class GridFieldRestoreAction implements GridField_ColumnProvider, GridField_Acti
             $restoreToRoot = RestoreAction::shouldRestoreToRoot($record);
 
             $title = $restoreToRoot
-                ? _t('SilverStripe\\Versioned\\RestoreAction.RESTORE_TO_ROOT', 'Restore to draft at top level')
-                : _t('SilverStripe\\Versioned\\RestoreAction.RESTORE', 'Restore to draft');
+                ? _t(RestoreAction::class . '.RESTORE_TO_ROOT', 'Restore to draft at top level')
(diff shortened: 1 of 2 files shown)


emteknetnz (changes requested):
This looks quite different from the rescued commit?

GuySartorelli:
This PR rescues "and expands upon" the original PR.
I cherry-picked in the commit from the original PR and then looked for instances of what the original commit was doing which had been either missed or added since the original commit was created. The PR would be incomplete if these additional items were left unchanged, since the purpose is to update the old hardcoded string syntax.

lang/en.yml
```diff
@@ -72,4 +72,5 @@ en:
     ONLIVEONLYSHORT: 'On live only'
     ONLIVEONLYSHORTHELP: 'Item is published, but has been deleted from draft'
   SilverStripe\Versioned\VersionedHTTPMiddleware:
+    Unpublished: 'Unpublished {name} {title}'
```
emteknetnz: We can remove this?

emteknetnz (changes requested):

lang/en.yml
ChloeSartorelli: Yeah. Not sure why that was being added in that commit - I had assumed it was in the wrong place originally and that this was the new place but now that I'm doing a search for it, it was definitely already in the right place. I'll remove this.

ChloeSartorelli: Done

emteknetnz (approved):

PR #375, opened 2022-10-18
API Update deprecations

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10531

### DIFF
--- a/_legacy/GraphQL/Extensions/DataObjectScaffolderExtension.php
+++ b/_legacy/GraphQL/Extensions/DataObjectScaffolderExtension.php
@@ -2,6 +2,7 @@
 
 namespace SilverStripe\Versioned\GraphQL\Extensions;
 
+use SilverStripe\Dev\Deprecation;
 use GraphQL\Type\Definition\ObjectType;
 use GraphQL\Type\Definition\Type;
 use SilverStripe\Core\Extension;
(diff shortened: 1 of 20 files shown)


GuySartorelli (changes requested):
Some of the version numbers are incorrect - use `1` major versioning instead. Also apply to any other deprecation notices in this module.

_legacy/GraphQL/Extensions/DataObjectScaffolderExtension.php
```diff
@@ -19,14 +20,19 @@
 }
 
 /**
- * @deprecated 4.8..5.0 Use silverstripe/graphql:^4 functionality.
+ * @deprecated 1.8.0 Use _graphql directory functionality instead
```
ChloeSartorelli: The message should say to use the latest version of graphql - the directory isn't the important factor here.
The message will need to be updated in some other classes below as well.

GuySartorelli (changes requested):

_legacy/GraphQL/Extensions/DataObjectScaffolderExtension.php
emteknetnz: Updated

GuySartorelli (approved):
MOG

PR #377, opened 2022-11-02
API Stop using deprecated API

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10542

### DIFF
--- a/src/Dev/VersionedTestSessionExtension.php
+++ b/src/Dev/VersionedTestSessionExtension.php
@@ -12,7 +12,6 @@ use SilverStripe\Versioned\VersionedStateExtension;
  * Decorates TestSession object to update get / post requests with versioned querystring arguments.
  * Session vars assigned by FunctionalTest::useDraftSite are respected here.
  *
- * @deprecated 2.2.0 Use ?stage=Stage|Live in your request's querystring instead
  * @property TestSession $owner
  */
 class VersionedTestSessionExtension extends VersionedStateExtension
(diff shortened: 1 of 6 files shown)


src/GraphQL/Resolvers/VersionedResolver.php
```diff
@@ -83,10 +83,17 @@ public static function resolveVersionList(array $resolverContext): Closure
             }
 
             // Get all versions
-            return $object->VersionsList();
+            return self::getVersionsList($object);
         };
     }
 
+    private static function getVersionsList(DataObject $object)
```
emteknetnz, line 90: This was copied from the deprecated Versioned::VersionsList()

This was the only non unit-test that called that method, so I kept it deprecated and just copied it to here as a private

src/Dev/VersionedTestSessionExtension.php
```diff
@@ -12,7 +12,6 @@
  * Decorates TestSession object to update get / post requests with versioned querystring arguments.
  * Session vars assigned by FunctionalTest::useDraftSite are respected here.
  *
- * @deprecated 2.2.0 Use ?stage=Stage|Live in your request's querystring instead
```
emteknetnz, line 15: I've raised an [issue](https://github.com/silverstripe/silverstripe-versioned/issues/378) to look at re-deprecating this separately from the rest of this issue - I don't want to leave this deprecated at this stage because I expect it to create too much noise

src/Versioned.php
```diff
@@ -2074,22 +2074,22 @@ public function stagesDiffer()
      * @param string $join Deprecated, use leftJoin($table, $joinClause) instead
      * @param string $having
      * @return ArrayList
+     * @deprecated 1.12.0 Use allVersions() instead'
```
emteknetnz: Deprecating this and undeprecating allVersions() because all this method does is all allVersions(). I have no idea why they were deprecated the other way around before.

sabina-talipova (approved):

PR #379, opened 2022-11-18
ENHANCEMENT: adding the ability to save; and publish if published already

sunnysideup:
…already

### DIFF
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -2684,14 +2684,15 @@ SQL
     
     /**
      * write to draft and if a published version already exists and it is currently the same as draft then 
-     * publish it as well.
+     * publish the change as well.
      * 
      * This allows you to update pages and keep them in the same state (published / modified / draft) as before
      */
     public function writeAndPublishIfAppropriate(?bool $recursive = true, ?bool $forceInsert = false)
     {
         // check reading mode
-        ReadingMode::validateStage(Versioned::DRAFT);
+        $myStage = Versioned::get_stage();
+        Versioned::set_stage(Versioned::DRAFT);
         // is it on live and is live the same as draft
         $canBePublished = $this->owner->isPublished() && ! $this->owner->isModifiedOnDraft();
         $this->owner->writeToStage(Versioned::DRAFT, $forceInsert);
@@ -2702,6 +2703,7 @@ SQL
                 $this->owner->publishRecursive();
             }
         }    
+        Versioned::set_stage($myStage);
     }
 
     /**
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -2698,9 +2698,10 @@ SQL
         $this->owner->writeToStage(Versioned::DRAFT, $forceInsert);
         if ($canBePublished) {
             if ($recursive) {
-                $this->owner->publishSingle();
-            } else {
                 $this->owner->publishRecursive();
+            } else {
+                $this->owner->publishSingle();
+                
             }
         }    
         Versioned::set_stage($myStage);
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -2681,6 +2681,28 @@ SQL
             }
         });
     }
+    
+    /**
+     * write to draft and if a published version already exists and it is currently the same as draft then 
+     * publish it as well.
+     * 
+     * This allows you to update pages and keep them in the same state (published / modified / draft) as before
+     */
+    public function writeAndPublishIfAppropriate(?bool $recursive = true, ?bool $forceInsert = false)
+    {
+        // check reading mode
+        ReadingMode::validateStage(Versioned::DRAFT);
+        // is it on live and is live the same as draft
+        $canBePublished = $this->owner->isPublished() && ! $this->owner->isModifiedOnDraft();
+        $this->owner->writeToStage(Versioned::DRAFT, $forceInsert);
+        if ($canBePublished) {
+            if ($recursive) {
+                $this->owner->publishSingle();
+            } else {
+                $this->owner->publishRecursive();
+            }
+        }    
+    }
 
     /**
      * Roll the draft version of this record to match the published record.


michalkleiner:
Where would this be used from?

src/Versioned.php
```diff
@@ -2681,6 +2681,30 @@ public function writeToStage($stage, $forceInsert = false)
...
+        $canBePublished = $this->owner->isPublished() && ! $this->owner->isModifiedOnDraft();
+        $this->owner->writeToStage(Versioned::DRAFT, $forceInsert);
+        if ($canBePublished) {
+            if ($recursive) {
+                $this->owner->publishSingle();
+            } else {
+                $this->owner->publishRecursive();
+            }
+        }    
+        Versioned::set_stage($myStage);
```
michalkleiner: You don't need to save and restore the old stage, you can use `withVersionedMode` and after the callback is processed the current stage is restored automatically.

Also the `publishSingle()` and `publishRecursive()` calls seemed to be the other way round in relation to the `$recursive` condition.

```suggestion
        static::withVersionedMode(function () use ($recursive, $forceInsert) {
            Versioned::set_stage(Versioned::DRAFT);
	        // is it on live and is live the same as draft
	        $canBePublished = $this->owner->isPublished() && ! $this->owner->isModifiedOnDraft();
	        $this->owner->writeToStage(Versioned::DRAFT, $forceInsert);
	        if ($canBePublished) {
	            if ($recursive) {
	                $this->owner->publishRecursive();
	            } else {
	                $this->owner->publishSingle();
	            }
	        }    
        }
```

sunnysideup:
>Where would this be used from?

This is useful if you would like to make an update to a page without changing its status.

src/Versioned.php
sunnysideup: > withVersionedMode

I prefer not to use callbacks, as this feels easier to debug / read.  Would you be ok with that? 

> recursive swap

oooops!  sorry, fixed

michalkleiner: > > withVersionedMode
> 
> I prefer not to use callbacks, as this feels easier to debug / read. Would you be ok with that?

Taking a quick look the two methods right above this suggested addition (and likely a lot of others in the class) are using the `static::withVersionedMode` callback approach, so I would prefer to keep that approach for consistency.

michalkleiner:
Since this is expanding the public API surface of the class, we would also need to add a test for this.

I can see how this can be handy in some use cases, though I'm on the fence about whether this should be a part of the class in this form and shape. It'd need the blessing of another core committer. Until resolved, the enhancement can be added via an extension or a trait in project code.

michalkleiner:
Failing linting is related.

sunnysideup:
We will set up our own module to provide this feature as this will take less time.  Thank you for looking at it.

michalkleiner:
We definitely appreciate your contributions and effort to make the CMS and its modules better, @sunnysideup. 

I think Guy or someone else already mentioned on some past PRs of yours that if something requires additional time from the core team to polish up and add tests etc. it's less likely to get the attention and fast action you would expect/prefer.
If the PR was adhering to the coding standards without linting issues, had tests added/updated to cover the new/changed behaviour and provided an explanation of why the addition is needed, handy or what problem it solves, then it's way more likely it will get reviewed and merged in faster.

sunnysideup:
>I think Guy or someone else already mentioned on some past PRs of yours that if something requires additional time from the core team to polish up and add tests etc. it's less likely to get the attention and fast action you would expect/prefer.
If the PR was adhering to the coding standards without linting issues, had tests added/updated to cover the new/changed behaviour and provided an explanation of why the addition is needed, handy or what problem it solves, then it's way more likely it will get reviewed and merged in faster.

Totally agree and understand that. But that is not a commercial reality for me. As you said, you are not sure if it is a useful contribution then it is not good use of my time and money to write a test (I would find this tricky).  I did answer this questions:

>provided an explanation of why the addition is needed, handy or what problem it solves

The reason I put it forward as an idea is that I have code like this in three projects, where we use cron jobs to update pages (e.g. expire events when they are done). We would like to do so without affecting the "published" status of the page. 


>If the PR was adhering to the coding standards without linting issues

As I have said before: you should install a linter to fix up these issues.  

```
----------------------------------------------------------------------
FOUND 4 ERRORS AFFECTING 4 LINES
----------------------------------------------------------------------
 2686 | ERROR | [x] Whitespace found at end of line
 2688 | ERROR | [x] Whitespace found at end of line
 2704 | ERROR | [x] Blank line found at end of control structure
 2706 | ERROR | [x] Whitespace found at end of line
----------------------------------------------------------------------
PHPCBF CAN FIX THE 4 MARKED SNIFF VIOLATIONS AUTOMATICALLY
----------------------------------------------------------------------
```

I guess from my perspective I just load a rough PR and if it gets a lot of traction then I am happy to put in more time. In this case it seems that is not worth it here. However, it is a good resource for others who may have the same issue etc... so I am not worried or upset if it gets closed and never acted on.

src/Versioned.php
ChloeSartorelli: I agree that `withVersionedMode` should be used

GuySartorelli:
> Since this is expanding the public API surface of the class, we would also need to add a test for this.

Absolutely agreed

> I'm on the fence about whether this should be a part of the class in this form and shape. 

I am too... though I think most of my hesitance comes from the name, which feels a little bit clunky... though I can't think of a better one right now.

I think I'm leaning more towards the "this is useful enough to have" side - but I would definitely not merge it without some pretty good tests.

GuySartorelli:
> you should install a linter to fix up these issues

It's a good idea, but it's just not happening any time soon. We have a lot of work to do, and little quality of life things like this which can be done manually with little effort are low priority right now.

> I just load a rough PR and if it gets a lot of traction then I am happy to put in more time.

The unfortunate reality is that at least I (and I suspect others) are unlikely to want to spend time on a rough PR. Something that looks nearly ready to merge is a lot more likely to get attention in the first place - rough PRs are a big glowing sign that says "This will need a lot of your time and effort to get it across the line" and unfortunately we just don't have that right now.

A better approach for features that you are unsure of would be to create an issue first - then we can say "yes that sounds good" or "no we don't want that" before you ever do any work, and it means your first PR for the issue can be basically ready to merge out of the gate because you don't need a "rough PR" to start.

sunnysideup:
>It's a good idea, but it's just not happening any time soon. We have a lot of work to do, and little quality of life things like this which can be done manually with little effort are low priority right now.

I would disagree with this, because it would make "doing a pull request" a bit easier - and this would be a really good thing.

dhensby:
> As I have said before: you should install a linter to fix up these issues.

@sunnysideup This has been mentioned before and I provided a pretty thorough response about why that's a bad idea (messing with git history/forcing commits to contributor branches/littering commit history with lintfix commits).

Linting is very low effort as the linter is already shipped with the repos: `composer run lint-clean` and you're done. If you have time to create the code, commit the work and open a PR, you have time to run the linting script too.

michalkleiner:
Linting related discussion with links to other issues and PRs — https://github.com/silverstripe/silverstripe-framework/issues/9616

michalkleiner:
I'll close this PR for now as this is unlikely to get in without further adjustments. Happy to reopen if there's new activity here. For the lining etc. we can move to the other linked issue or to the forum.

PR #381, opened 2022-11-28
API Remove deprecated code

sabina-talipova:
### Parent issue
- https://github.com/silverstripe/silverstripe-framework/issues/10501


src/Versioned.php
```diff
@@ -1959,11 +1941,6 @@ public function doRevertToLive()
      */
     public function copyVersionToStage($fromStage, $toStage, $createNewVersion = true)
     {
-        // Disallow $createNewVersion = false
```
emteknetnz, line 1962: Remove the $createNewVersion param

emteknetnz (changes requested):

src/Versioned.php
sabina-talipova: Done

emteknetnz (approved):

PR #384, opened 2022-11-30
Add canPublish check for ChangeSet

### LINKED ISSUE #383: Page `$owns` does not respect asset `canPublish`
Error scenario:
- An asset has an Expiry date (we do not own the image and we are only allowed to use the image for a certain timeframe.)
- We need to prevent the image to be published if an `ExpiryDate` field has not been added by the Content Editor.
- We can still publish the parent page as it is only the asset that has restriction

Expected:
- Page can be published while its image should remain in Draft mode.

Project setup:
- Followed the steps here : https://docs.silverstripe.org/en/4/getting_started/
- Add a file extension   -> (simple canPublish to false) 
- Add upload field to Page   
- Dev build as usual

How to re-create:
- Add a page and insert a `Photo`
- Upload the photo (as you can see no publish button will show up as I removed it via `canPublish`)
- Go back to the page
- Hit Publish
- Attached photo is also published.
  
----------
FileExtension:
```php
<?php

namespace App\Extensions;

use SilverStripe\ORM\DataExtension;

class FileExtension extends DataExtension
{
    /**
     * Don't allow a file to be published if there is no Expiry details
     */
    public function canPublish($member = null)
    {
        return false;
    }

    /**
     * Check if this file can be published
     *
     * @param Member $member
     * @return boolean
     */
    public function extendCanPublish($member = null)
    {
        return false;
    }
}
```
----------
Page:
```php
<?php

namespace {

    use SilverStripe\AssetAdmin\Forms\UploadField;
    use SilverStripe\Assets\Image;
    use SilverStripe\CMS\Model\SiteTree;

    class Page extends SiteTree
    {
        private static $db = [];

        private static $has_one = [
            'Photo' => Image::class,
        ];

        private static $owns = [
            'Photo',
        ];


        public function getCMSFields()
        {
            $fields = parent::getCMSFields();

            $fields->removeFieldFromTab('Root.Main', 'ParentURL');

            $fields->addFieldToTab('Root.Attachments', UploadField::create('Photo'));

            return $fields;
        }
    }
}
```
----------
Extension:
```yml
---
Name: file-assets
After:
  - "asset-admin/*"
  - "versioned/*"
  - assetadmin
---
SilverStripe\Assets\File:
  extensions:
    - App\Extensions\FileExtension
```

## PRs
- https://github.com/silverstripe/silverstripe-versioned/pull/384

mark-a-j-adriano:
Maybe the solution for https://github.com/silverstripe/silverstripe-versioned/issues/383

### DIFF
--- a/src/ChangeSetItem.php
+++ b/src/ChangeSetItem.php
@@ -304,16 +304,21 @@ class ChangeSetItem extends DataObject implements Thumbnail
             case static::CHANGE_CREATED: {
                 // Non-recursive publish
                 $object = $this->getObjectInStage(Versioned::DRAFT);
-                $object->publishSingle();
+                $member = Security::getCurrentUser();
 
-                // Point after version to the published version actually created, not the
-                // version copied from draft.
-                $this->VersionAfter = Versioned::get_versionnumber_by_stage(
-                    $this->ObjectClass,
-                    Versioned::LIVE,
-                    $this->ObjectID,
-                    false
-                );
+                // Check if object has canPublish set to true
+                if ($object->canPublish($member)) {
+                    $object->publishSingle();
+
+                    // Point after version to the published version actually created, not the
+                    // version copied from draft.
+                    $this->VersionAfter = Versioned::get_versionnumber_by_stage(
+                        $this->ObjectClass,
+                        Versioned::LIVE,
+                        $this->ObjectID,
+                        false
+                    );
+                }
                 break;
             }
             default:
@@ -524,8 +529,8 @@ class ChangeSetItem extends DataObject implements Thumbnail
             $live = $this->getObjectInStage(Versioned::LIVE);
             if ($live instanceof CMSPreviewable && $live->canView() && ($link = $live->PreviewLink())) {
                 $links[Versioned::LIVE] = [
-                'href' => Controller::join_links($link, '?stage=' . Versioned::LIVE),
-                'type' => $live->getMimeType(),
+                    'href' => Controller::join_links($link, '?stage=' . Versioned::LIVE),
+                    'type' => $live->getMimeType(),
                 ];
             }
         }


src/ChangeSetItem.php
```diff
@@ -304,16 +304,21 @@ public function publish()
...
-                $this->VersionAfter = Versioned::get_versionnumber_by_stage(
-                    $this->ObjectClass,
-                    Versioned::LIVE,
-                    $this->ObjectID,
-                    false
-                );
+                $member = Security::getCurrentUser();
+
+                // Check if object has canPublish set to true
+                if ($object->canPublish($member)) {
```
kinglozzer, line 310: We shouldn’t add this kind of logic here, as it’ll potentially block programatic publishing (e.g. a CLI task where no user is logged in)

lozcalver, line 310: We shouldn’t add this kind of logic here, as it’ll potentially block programatic publishing (e.g. a CLI task where no user is logged in)

lozcalver (changes requested):

michalkleiner:
`ChangeSetItem` already has its own [`canPublish` method ](https://github.com/silverstripe/silverstripe-versioned/blob/1/src/ChangeSetItem.php#L403) that checks the record object's canPublish, so it must be something else going on (if something is actually going on).

src/ChangeSetItem.php
mark-a-j-adriano: > `ChangeSetItem` already has its own [`canPublish` method ](https://github.com/silverstripe/silverstripe-versioned/blob/1/src/ChangeSetItem.php#L403) that checks the record object's canPublish, so it must be something else going on (if something is actually going on).

Nope, `ChangeSetItem` canPublish method is **not** called.  https://github.com/silverstripe/silverstripe-versioned/blob/4baa7b48298c18617c2efc1ea5cf912af62de414/src/ChangeSetItem.php#L307
Already acts on the asset

mark-a-j-adriano: @kinglozzer -> not sure where to put the logic if it is not the correct place. (In our use case this might suffice)

lozcalver: I think for your use case I’d probably remove the image from `$owns` and add some logic in `onAfterPublish()` to check the date and publish if applicable

michalkleiner: That would be a workaround for Mark, but I think this is a legit bug where it is reasonable to anticipate the permissions are respected, or we openly document that ownership cascade publishing trumps individual object permissions.

ChloeSartorelli: > I think this is a legit bug where it is reasonable to anticipate the permissions are respected

I agree

lozcalver: Yep sorry I meant as a workaround! Will post my thoughts in the issue #383

GuySartorelli:
I'm going to close this PR for now. As mentioned in the issue, this can't be resolved until we're ready to start working on a new major release.
The closed PR will still be associated with and linked from the issue, which is already in the CMS 6 milestone, so when we start working on that we'll have access to this PR to reference as a potential solution.

PR #385, opened 2022-12-05
FIX Ensure sort is not empty

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10449

Requires https://github.com/silverstripe/silverstripe-framework/pull/10600 to be merged first - but this was run though an installer CI run: https://github.com/emteknetnz/silverstripe-installer/actions/runs/3617074332

### DIFF
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -2874,11 +2874,14 @@ SQL
      */
     public static function get_including_deleted($class, $filter = "", $sort = "")
     {
-        $list = DataList::create($class)
-            ->where($filter)
-            ->sort($sort)
-            ->setDataQueryParam("Versioned.mode", "latest_versions");
-
+        $list = DataList::create($class);
+        if (!empty($filter)) {
+            $list = $list->where($filter);
+        }
+        if (!empty($sort)) {
+            $list = $list->orderBy($sort);
+        }
+        $list = $list->setDataQueryParam("Versioned.mode", "latest_versions");
         return $list;
     }
 
--- a/tests/php/VersionedGridFieldTest/TestController.php
+++ b/tests/php/VersionedGridFieldTest/TestController.php
@@ -42,7 +42,7 @@ class TestController extends Controller implements TestOnly
     public function Form()
     {
         $objects = TestObject::get()
-            ->sort('"VersionedTest_DataObject"."ID" ASC');
+            ->orderBy('"VersionedTest_DataObject"."ID" ASC');
         $field = new GridField('testfield', 'testfield', $objects, GridFieldConfig_RelationEditor::create());
         return new Form($this, 'Form', new FieldList($field), new FieldList());
     }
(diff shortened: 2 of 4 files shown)


src/Versioned.php
```diff
@@ -2874,11 +2874,14 @@ public function isModifiedOnDraft()
...
     public static function get_including_deleted($class, $filter = "", $sort = "")
     {
-        $list = DataList::create($class)
-            ->where($filter)
-            ->sort($sort)
-            ->setDataQueryParam("Versioned.mode", "latest_versions");
-
+        $list = DataList::create($class);
+        if (!empty($filter)) {
+            $list = $list->where($filter);
```
michalkleiner, line 2879: Can we use `filter` in line with the other work related to replacing where with filter?

```suggestion
            $list = $list->filter($filter);
```

src/Versioned.php
```diff
@@ -2874,11 +2874,14 @@ public function isModifiedOnDraft()
...
+        }
+        if (!empty($sort)) {
+            $list = $list->orderBy($sort);
```
ChloeSartorelli, line 2882: Is `orderBy()` correct here? What passes raw SQL into this method?

If it's just something passing in `"Table"."Column"` syntax then we should probably either a) consider allowing that syntax in `sort()` (it's fairly common and isn't vulnerable to sql injection attacks) or b) reformat the value, for example by excluding the table name, depending on what exactly is being passed in from where.

GuySartorelli (changes requested):

src/Versioned.php
emteknetnz: The "migration" is to use orderBy() so that existing raw sql still works, though change it to sort() if it's easy enough.  To me, this is a hard migration so I think orderBy() is ok here, though obviously I'd prefer sort()

a) would is basically new functionality on DataList::sort() and would requires a new card IMO since there's a whole lot of validating that tables exist etc. I have no idea what happens for instance when joining tables where the there is a $table_name that doesn't match the Model class name for instance.

b) Is just a) - handling table names

In terms of where the $sort param is actually used, outside of unit tests, it's only Hierarchy.php

```php
        $baseTable = $owner->baseTable();
        $parentIDColumn = $owner->getSchema()->sqlColumnForField($owner, 'ParentID');
        return Versioned::get_including_deleted(
            $owner->baseClass(),
            [ $parentIDColumn => $owner->ID ],
            "\"{$baseTable}\".\"ID\" ASC"
        );
```

I'm not sure how I refactor this to sort() the results by `"{$baseTable}"."ID" ASC"` - where $baseTable is the mysql table name

src/Versioned.php
emteknetnz: That should be done as part of this issue - https://github.com/silverstripe/silverstripe-framework/issues/10601

src/Versioned.php
emteknetnz: Based on [notation clarification](https://github.com/silverstripe/silverstripe-framework/pull/10600#discussion_r1040323455) - we need orderBy() here

GuySartorelli (approved):

GuySartorelli:
Looks good.
Will merge when silverstripe/silverstripe-framework#10600 is ready to merge

PR #386, opened 2022-12-14
FIX Handle a bool return type from DateTime::getLastErrors()

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/10403

Fixes `1) SilverStripe\Versioned\GraphQL\Resolvers\VersionedFiltersTest::testItSetsReadingStateByArchiveDate
TypeError: array_sum(): Argument #1 ($array) must be of type array, bool given`
https://github.com/emteknetnz/silverstripe-installer/actions/runs/3699031675/jobs/6265942969

[DateTime::getLastErrors()](https://www.php.net/manual/en/datetime.getlasterrors.php) will return false when there are no errors with the date parsed and `array_sum(false)` will throw the error above - I'm honestly not sure how this used to previously work.

### DIFF
--- a/src/GraphQL/Resolvers/VersionFilters.php
+++ b/src/GraphQL/Resolvers/VersionFilters.php
@@ -221,7 +221,19 @@ class VersionFilters
     protected function isValidDate($date)
     {
         $dt = DateTime::createFromFormat('Y-m-d', $date);
-
-        return ($dt !== false && !array_sum($dt->getLastErrors() ?? []));
+        if ($dt === false) {
+            return false;
+        }
+        // DateTime::getLastErrors() has an undocumented difference pre PHP 8.2 for what's returned
+        // if there are no errors
+        // https://www.php.net/manual/en/datetime.getlasterrors.php
+        $errors = $dt->getLastErrors();
+        // PHP 8.2 - will return false if no errors
+        if ($errors === false) {
+            return true;
+        }
+        // PHP 8.2+ will only return an array containing a count of errors only if there are errors
+        // PHP < 8.2 will always return an array containing a count of errors even if there are no errors
+        return array_sum($errors) === 0;
     }
 }


src/GraphQL/Resolvers/VersionFilters.php
```diff
@@ -221,7 +221,19 @@ public function validateArgs(array $versioningArgs)
     protected function isValidDate($date)
     {
         $dt = DateTime::createFromFormat('Y-m-d', $date);
-
-        return ($dt !== false && !array_sum($dt->getLastErrors() ?? []));
+        if ($dt === false) {
+            return false;
+        }
+        // DateTime::getLastErrors() has an undocumented difference pre PHP 8.2 for what's returned
```
emteknetnz, line 227: This was a surprise - old behaviour was buggy - fixed in php8.2 - mentioned in [php 8 changelog](https://www.php.net/ChangeLog-8.php)

Fixed bug [GH-9431](https://github.com/php/php-src/issues/9431)

Fix happened [here](https://github.com/php/php-src/issues/9431)

```
PHP 7.4 and PHP 8.1
php -r 'var_dump(DateTime::createFromFormat("Y-m-d", "2018-01-01")->getLastErrors());'
array(4) {
  ["warning_count"]=>
  int(0)
  ["warnings"]=>
  array(0) {
  }
  ["error_count"]=>
  int(0)
  ["errors"]=>
  array(0) {
  }
}

PHP 8.2
php -r 'var_dump(DateTime::createFromFormat("Y-m-d", "2018-01-01")->getLastErrors());'
bool(false)
```

src/GraphQL/Resolvers/VersionFilters.php
```diff
@@ -221,7 +221,19 @@ public function validateArgs(array $versioningArgs)
...
+        // if there are no errors
+        // https://www.php.net/manual/en/datetime.getlasterrors.php
+        $errors = $dt->getLastErrors();
+        // PHP 8.2 - will return false if no errors
+        if ($errors === false) {
+            return true;
+        }
+        // PHP 8.2+ will only return an array containing a count of errors only if there are errors
+        // PHP < 8.2, will return always return array containing a count of errors even if there are no errors
+        return array_sum($errors) === 0;
```
michalkleiner, line 237:
```suggestion
        return $errors['error_count'] === 0;
```
or
```suggestion
        return count($errors['errors']) === 0;
```
?

src/GraphQL/Resolvers/VersionFilters.php
```diff
@@ -221,7 +221,19 @@ public function validateArgs(array $versioningArgs)
...
+        // DateTime::getLastErrors() has an undocumented difference pre PHP 8.2 for what's returned
+        // if there are no errors
+        // https://www.php.net/manual/en/datetime.getlasterrors.php
+        $errors = $dt->getLastErrors();
+        // PHP 8.2 - will return false if no errors
+        if ($errors === false) {
+            return true;
+        }
+        // PHP 8.2+ will only return an array containing a count of errors only if there are errors
+        // PHP < 8.2, will return always return array containing a count of errors even if there are no errors
```
michalkleiner:
```suggestion
        // PHP < 8.2 will always return an array containing a count of errors even if there are no errors
```

emteknetnz: Updated

src/GraphQL/Resolvers/VersionFilters.php
emteknetnz: I think this should be kept as array_sum so that it matches the old logic.  If we change it to what's recommended here then we're ignoring the 'warning_count' key.

michalkleiner: Right, ok, even though we're doubling up if there are any errors it doesn't matter for our use case.

michalkleiner (approved):

PR #400, opened 2023-03-26
MNT Add config key for Versioned extension applied to File

michalkleiner:
Add named key to the config where Versioned is applied to File so that the extension can be removed through custom config if needed.

### DIFF
--- a/_config/versionedfiles.yml
+++ b/_config/versionedfiles.yml
@@ -3,4 +3,4 @@ Name: versionedfiles
 ---
 SilverStripe\Assets\File:
   extensions:
-    - SilverStripe\Versioned\Versioned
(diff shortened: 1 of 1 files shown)


michalkleiner:
Not sure if this should target `2` or `2.0`, can retarget if needed.

GuySartorelli:
Retargetted to `2.0` - this doesn't break (or even really materially change) anything so happy for it to be part of the initial release.

@michalkleiner looks like retargetting has dragged along some unrelated commits

michalkleiner:
I'll rebase it, just a sec

GuySartorelli (approved):
Nice, thanks!
Ignoring CI failures from PUSH because the ones related to the PR itself are passing.

PR #402, opened 2023-04-05
FIX Use WasPublished to match GraphQL 3 behaviour

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-versioned-admin/issues/255

This matches the graphql3 resolver https://github.com/silverstripe/silverstripe-versioned/blob/1.12/_legacy/GraphQL/Extensions/DataObjectScaffolderExtension.php#L74

### DIFF
--- a/src/GraphQL/Resolvers/VersionedResolver.php
+++ b/src/GraphQL/Resolvers/VersionedResolver.php
@@ -43,7 +43,7 @@ class VersionedResolver
             case 'publisher':
                 return $obj->Publisher();
             case 'published':
-                return $obj->isPublished();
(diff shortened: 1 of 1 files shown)


michalkleiner (approved):
MOG after the versioned-admin module's PR is merged and tests rerun

emteknetnz:
@michalkleiner We actually need THIS pr merged first, then re-run the versioned-admin PR so that it goes green - this other one failed (expected), it should pass on re-run once this is merged

michalkleiner:
Yeah of course. Merged and triggered a re-run of the versioned-admin test.

Issue #422, opened 2023-10-24
Item that has been added implicitly to ChangeSet, cannot be removed


sabina-talipova:
## Description
Item that has been added implicitly to the ChangeSet as ChangeSetItem,  cannot be remove if explicitly added parent item was removed and ChangeSet has State = 'published'.

### Steps to reproduce
- Create page
- Add an Image to the Page
- Save, but don't publish
- Add the Page to Campaign
- Go to the Campaign Admin
- Open Created Campaign
- You should see your Page in PAGES section and Image in FILES section
- Click on the Page in PAGES section
- Click Remove in "Action Menu"
- Page and appended to the page image were removed form PAGES and FILES sections
- Repeat previous steps before "Click Remove in "Action Menu""
- Instead Click Publish Campaign button
- Click on the Page in PAGES section
- Click Remove in "Action Menu"
- Page was removed form PAGES and Image still exist in FILES sections
- Click on the Image in FILES section
- Click "Action Menu"
- **You should see "Required by {number} item(s), and cannot be removed directly."**

## Related issue
- https://github.com/silverstripe/silverstripe-versioned/issues/415

## PRs
- https://github.com/silverstripe/silverstripe-versioned/pull/423
- https://github.com/silverstripe/silverstripe-campaign-admin/pull/286

maxime-rainville:
I've reviewed #423.

I'm wondering if the correct behaviour shouldn't be to "freeze" a campaign once it has been published. It sounds kind of nonsensical to be able to remove items from Campaign Admin post publication. If the we decide the correct behaviour here is to block people from removing item from published campaign, than the behat PR will need to be updated.

maxime-rainville:
As discussed, we don't think there's a valid usecase for removing item form campaign post publication.

So the second Behat PR is not needed.

I've created this follow up card: https://github.com/silverstripe/silverstripe-campaign-admin/issues/287

PR #423, opened 2023-10-24
FIX Remove implicitly added item

sabina-talipova:
## Description
- Implicitly added item to the ChangeSet will be removed together with parent explicitly added item if ChangeSet has State = 'open'.
- Implicitly added item to the ChangeSet cannot be remove if explicitly added parent item was removed and ChangeSet has State = 'published'. 
We should remove implicitly added item together with parent item even if ChangeSet was already published.   

## Parent issue
- https://github.com/silverstripe/silverstripe-versioned/issues/422

### DIFF
--- a/src/ChangeSet.php
+++ b/src/ChangeSet.php
@@ -216,6 +216,22 @@ class ChangeSet extends DataObject
             // TODO: Handle case of implicit added item being removed.
 
             $item->delete();
+
+            // Get the implicitly included items for this ChangeSet
+            $implicit = $this->calculateImplicit();
+
+            foreach ($this->Changes()->filter(['Added' => ChangeSetItem::IMPLICITLY]) as $changeSetItem) {
+                $objectKey = $this->implicitKey($changeSetItem);
+
+                // If a ChangeSetItem exists, but isn't in $implicit, it's no longer required, so delete it
+                if (!array_key_exists($objectKey, $implicit)) {
+                    $changeSetItem->delete();
+                } else {
+                    // Otherwise it is required, so update ReferencedBy and remove from $implicit
+                    $changeSetItem->ReferencedBy()->setByIDList($implicit[$objectKey]['ReferencedBy']);
+                    unset($implicit[$objectKey]);
+                }
+            }
         }
 
         $this->sync();
--- a/tests/php/ChangeSetTest.php
+++ b/tests/php/ChangeSetTest.php
@@ -160,6 +160,7 @@ class ChangeSetTest extends SapphireTest
                 ChangeSetTest\BaseObject::class . '.base' => ChangeSetItem::EXPLICITLY,
                 ChangeSetTest\MidObject::class . '.mid1' => ChangeSetItem::IMPLICITLY,
                 ChangeSetTest\MidObject::class . '.mid2' => ChangeSetItem::IMPLICITLY,
+                ChangeSetTest\MidObject::class . '.mid4' => ChangeSetItem::IMPLICITLY,
                 ChangeSetTest\EndObject::class . '.end1' => ChangeSetItem::IMPLICITLY,
                 ChangeSetTest\EndObject::class . '.end2' => ChangeSetItem::IMPLICITLY,
             ]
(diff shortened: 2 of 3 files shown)


tests/php/ChangeSetTest.php
```diff
@@ -735,4 +735,32 @@ public function testIsSyncedCanBeSkipped()
...
+
+        $mid1 = $this->objFromFixture(ChangeSetTest\MidObject::class, 'mid1');
+        $mid2 = $this->objFromFixture(ChangeSetTest\MidObject::class, 'mid2');
+
+        $changeset = new ChangeSet();
+        $changeset->write();
+        $changeset->addObject($mid1);
+        $changeset->addObject($mid2);
+        $changeset->publish();
+
```
maxime-rainville, line 757: Can we add a `assertChangeSetLooksLike` assertion here that `mid1` and `end1` are alse part of the changeset before we remove them?

tests/php/ChangeSetTest.php
```diff
@@ -735,4 +735,32 @@ public function testIsSyncedCanBeSkipped()
...
+            [
+                ChangeSetTest\MidObject::class . '.mid2' => ChangeSetItem::EXPLICITLY,
+                ChangeSetTest\EndObject::class . '.end2' => ChangeSetItem::IMPLICITLY,
+            ]
+        );
+
+        $changeset->removeObject($mid2);
+
+        $this->assertChangeSetLooksLike($changeset, []);
+    }
```
maxime-rainville, line 845: Can we add two extra test cases:
- Item A and Item B both reference Item C. Item A is removed from changeset. Item C should still be in the changeset.
- Item D references Item E. Item D and Item E are both explicitly added to changeset. Removing Item D should not remove Item E.

From the UI testing I've done, both those test case seem to behave as expected. It's just a question of making sure they are covered by Unit test.

maxime-rainville (changes requested):
That's an elegant solution. I'm just thinking we should add a few extra scenarios in the test coverage.

tests/php/ChangeSetTest.php
sabina-talipova: Updated

tests/php/ChangeSetTest.php
sabina-talipova: Updated

src/ChangeSet.php
```diff
@@ -216,6 +216,22 @@ public function removeObject(DataObject $object)
...
             $item->delete();
+
+            // Get the implicitly included items for this ChangeSet
+            $implicit = $this->calculateImplicit();
+
+            foreach ($this->Changes()->filter(['Added' => ChangeSetItem::IMPLICITLY]) as $changeSetItem) {
+                $objectKey = $this->implicitKey($changeSetItem);
+
+                // If a ChangeSetItem exists, but isn't in $implicit, it's no longer required, so delete it
+                if (!array_key_exists($objectKey, $implicit ?? [])) {
```
michalkleiner: Null coalescing fallback not needed here as `calculateImplicit` should return an array already. If not, it should be added there rather than here.

```suggestion
                if (!array_key_exists($objectKey, $implicit)) {
```

sabina-talipova: Updated

tests/php/ChangeSetTest.php
```diff
@@ -735,4 +743,69 @@ public function testIsSyncedCanBeSkipped()
...
+        $mid3 = $this->objFromFixture(ChangeSetTest\MidObject::class, 'mid4'); // Item mid4 reference Item end2
+
+        $end1 = $this->objFromFixture(ChangeSetTest\EndObject::class, 'end1'); // Item end1
+
+        $changeset = new ChangeSet();
+        $changeset->write();
+        $changeset->addObject($mid1);
+        $changeset->addObject($mid2);
+        $changeset->addObject($mid3);
+        $changeset->addObject($end1); // Item end1 explisitly added to ChangeSet
```
maxime-rainville:
```suggestion
        $changeset->addObject($end1); // Item end1 explicitly added to ChangeSet
```

tests/php/ChangeSetTest.php
```diff
@@ -735,4 +743,69 @@ public function testIsSyncedCanBeSkipped()
 
         $this->assertFalse($changeset->isSyncCalled, 'isSynced is skipped when providing truthy argument to publish');
     }
+
+    public function testRemoveObject()
```
maxime-rainville, line 747: Personally, I would have broken up this method into 3 different tests just so we can better see what is being tested. But I'm not fuss about it.

maxime-rainville (changes requested):
Consider adding an optional `$message` parameter to the `assertChangeSetLooksLike` helper method. That could help make the test more readable. 

You could provide assumption like:
- Removing an explicitly defined ChangeSetItem removes all its implicit ChangeSetItems.
- Removing an explicitly defined ChangeSetItem does not remove its implicit ChangeSetItem if that item is also referenced by another a third one.
- Removing an explicitly defined ChangeSetItem does not remove its implicit ChangeSetItem if that item is also explicitly added to the changeset

This is more in the nice to have category.

tests/php/ChangeSetTest.php
sabina-talipova: Updated

tests/php/ChangeSetTest.php
sabina-talipova: Updated

maxime-rainville (approved):
LGTM

Issue #417, opened 2023-10-09
We're calling a method (`having`) that doesn't exist on `DataList`


sabina-talipova:
## Description
Method `having()` doesn't exist in `DataList` class, but we're trying to call it If `Versioned::Versions()` receives `$having` parameter. 

We need to either add that method to `DataList` or apply the `$having` argument via `alterDataQuery()`

See https://github.com/silverstripe/silverstripe-versioned/blob/53594900206c2b1909ce11cfd22286f9c7aaa46f/src/Versioned.php#L2008-L2026

## Acceptance criteria
- The `$having` argument in `Versioned::Versions()` can be called without throwing an exception and can be meaningfully use.

## Note
We are assuming that this can be done relatively easily. But if it turns out to be very difficult, bring it back for re-refinement.

## PR
- https://github.com/silverstripe/silverstripe-versioned/pull/425
- https://github.com/silverstripe/developer-docs/pull/396

GuySartorelli:
Marking as impact/high.
This would be a critical in other circumstances, but the fact that this hasn't caused actual problems until now mitigates its impact somewhat.

GuySartorelli:
PR merged. Reassigning to Steve to add a note to the PR about the deprecated parameter.

emteknetnz:
> Reassigning to Steve to add a note to the PR about the deprecated parameter.

Not sure what I'm supposed to do here?

GuySartorelli:
Changelog, sorry.

emteknetnz:
PR added

GuySartorelli:
PRs merged

PR #425, opened 2023-11-03
API Deprecate Versions() having parameter

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-versioned/issues/417

Currently this parameter is used and if called could cause a hard error since the having() method doesn't exist on DataList

I'm not even sure why this paramater ever existed since I'm not sure how it could even be used in this context since you're calling DataList::get() and then passing through the $filter, $sort, $limit and $join paramaters ... though there doesn't seem like there's anyway to even use having since it's not creating a select SQL statement with an aggregate function like `count()` in it

Seems best to just prevent it from even attempting to call the non-existant DataList::having() and deprecate it

### DIFF
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -12,6 +12,7 @@ use SilverStripe\Core\ClassInfo;
 use SilverStripe\Core\Config\Config;
 use SilverStripe\Core\Extension;
 use SilverStripe\Core\Resettable;
+use SilverStripe\Dev\Deprecation;
 use SilverStripe\Forms\FieldList;
 use SilverStripe\ORM\ArrayList;
 use SilverStripe\ORM\DataExtension;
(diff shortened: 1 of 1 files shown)


GuySartorelli:
This wasn't one of the options discussed in refinement. Why not implement having on datalist, or use `alterDataQuery` to call having on the underlying `DataQuery`?

emteknetnz:
Having a better look at this now seems like this param on Versions() that shouldn't have been there in the first place, so it doesn't make sense IMO to do work to support it

GuySartorelli (changes requested):
Need to target `2`. We don't deprecate API in patch releases.

Not super happy with this approach but it's an improvement on what we had, since it's not throwing an error any time someone tries to use the parameter now - so I'll be okay with it when it's targetting the `2` branch.

emteknetnz:
@GuySartorelli done

GuySartorelli (approved):
LGTM

PR #426, opened 2023-11-03
ENH add CAN_DEV_BUILD to non_live_permissions

andrewandante:
Companion PR to https://github.com/silverstripe/silverstripe-framework/pull/10979

## Issue

https://github.com/silverstripe/silverstripe-framework/issues/10852

### DIFF
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -282,7 +282,7 @@ class Versioned extends DataExtension implements TemplateGlobalProvider, Resetta
      * @config
      * @var array
      */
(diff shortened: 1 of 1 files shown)


composer.json
```diff
@@ -20,7 +20,7 @@
     ],
     "require": {
         "php": "^8.1",
-        "silverstripe/framework": "^5",
+        "silverstripe/framework": "^5.2",
```
ChloeSartorelli: No need to bump this, since the change won't break anything even if the permission doesn't exist in a given project.

GuySartorelli (changes requested):

GuySartorelli (approved):
LGTM, works as expected. Will merge when the framework PR has been merged.

PR #431, opened 2024-01-11
ENH Add generic types

GuySartorelli:
Requires https://github.com/silverstripe/silverstripe-framework/pull/11108 for these generics to take effect.

## Issue
- https://github.com/silverstripe/silverstripe-framework/issues/11066

### DIFF
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -36,8 +36,9 @@ use SilverStripe\View\TemplateGlobalProvider;
  * Note: This extension relies on the object also having the {@see Ownership} extension applied.
  *
  * @property int $Version
- * @property DataObject|RecursivePublishable|Versioned $owner
  * @mixin RecursivePublishable
+ *
+ * @extends DataExtension<DataObject&RecursivePublishable&static>
  */
 class Versioned extends DataExtension implements TemplateGlobalProvider, Resettable
 {
@@ -1854,7 +1855,6 @@ SQL
      */
     public function doUnpublish()
     {
-        /** @var DataObject|Versioned $owner */
         $owner = $this->owner;
         // Skip if this record isn't saved
         if (!$owner->isInDB()) {
@@ -1875,7 +1875,6 @@ SQL
             // Re-fetch the current DataObject to ensure we have data from the LIVE stage
             // This is particularly relevant for DataObject's in a modified state so that
             // any delete extensions have the correct database record values
-            /** @var DataObject|Versioned $obj */
             $obj = $owner::get()->byID($owner->ID);
             if (!$obj) {
                 return;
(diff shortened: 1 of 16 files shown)


src/Versioned.php
```diff
@@ -36,8 +36,9 @@
  * Note: This extension relies on the object also having the {@see Ownership} extension applied.
  *
  * @property int $Version
- * @property DataObject|RecursivePublishable|Versioned $owner
  * @mixin RecursivePublishable
+ *
+ * @extends DataExtension<DataObject&RecursivePublishable&Versioned>
```
emteknetnz:
```suggestion
 * @extends DataExtension<DataObject&RecursivePublishable&static>
```

src/Versioned.php
```diff
@@ -2363,13 +2360,13 @@ public static function reading_archived_date($date, $stage = self::DRAFT)
...
-     * @param string $class The name of the class.
+     * @template T of DataObject
+     * @param class-string<T> $class The name of the class.
      * @param string $stage The name of the stage.
      * @param string $filter A filter to be inserted into the WHERE clause.
      * @param boolean $cache Use caching.
      * @param string $sort A sort expression to be inserted into the ORDER BY clause.
-     *
-     * @return DataObject
+     * @return T&Versioned
```
emteknetnz:
```suggestion
     * @return T&static
```

src/Versioned.php
```diff
@@ -2657,9 +2654,10 @@ public function rollbackSingle($version)
     /**
      * Return the latest version of the given record.
      *
-     * @param string $class
+     * @template T of DataObject
+     * @param class-string<T> $class
      * @param int $id
-     * @return DataObject
+     * @return T&Versioned
```
emteknetnz:
```suggestion
     * @return T&static
```

src/VersionedStateExtension.php
```diff
@@ -11,7 +11,8 @@
 /**
  * Persists versioned state between requests via querystring arguments
  *
- * @property RequestHandler|DataObject $owner
+ * @template T of RequestHandler|DataObject
+ * @extends Extension<T>
```
emteknetnz, line 15:
```suggestion
 * @extends Extension<RequestHandler|DataObject>
```
Wouldn't this do the same thing?

emteknetnz (changes requested):

src/VersionedStateExtension.php
ChloeSartorelli: Nope. Having the `@template T` there means subclasses can say "Use my type instead" using `@extends`.
`VersionedTestSessionExtension` is a subclass of `VersionedStateExtension` and has its own type, so `VersionedStateExtension` has to use `@template T` to allow that to happen.

src/Versioned.php
ChloeSartorelli: Done

src/Versioned.php
ChloeSartorelli: Done

src/Versioned.php
ChloeSartorelli: Done

emteknetnz (approved):

Issue #447, opened 2024-05-27
Remove `canArchive()` method for CMS 6


GuySartorelli:
When calling `delete()` on a record (without changing what versioned state you're operating in), the record actually ends up being archived. While it is more correct to call `doArchive()` instead of `delete()` if you intend to archive the record, it's not uncommon to just call `delete()` directly.

Because of this, it's easy to call `canDelete()` as the permission check before archiving a record, forgetting that there's a specific `canArchive()` method that should be called instead.

This has lead to [at least one issue](https://github.com/silverstripe/silverstripe-linkfield/issues/291) in the past.

We should consider whether to remove the `canArchive()` permission method, and just assume that `canDelete()` is always `canArchive()`.

### Note
- DataObject::delete() and Versioned::doArchive() methods are not being changed

### Acceptance criteria
- [x] canArchive() method is deprecated in CMS 5
- [x] canArchive() method is deleted in CMS 6
- [x] Code in updated to use canDelete() in CMS 6

## CMS 5 PRs
- https://github.com/silverstripe/silverstripe-versioned/pull/461
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1483
- https://github.com/silverstripe/silverstripe-cms/pull/2985
- https://github.com/silverstripe/silverstripe-linkfield/pull/319
- https://github.com/silverstripe/silverstripe-elemental/pull/1233

## CMS 6 PRs

> [!IMPORTANT]
> Don't merge these until the CMS 5 PRs have been merged and merged up.
> If there are merge conflicts (which there almost 100% will be) reassign to Guy to resolve.

- https://github.com/silverstripe/silverstripe-versioned/pull/460
- https://github.com/silverstripe/silverstripe-asset-admin/pull/1482
- https://github.com/silverstripe/silverstripe-cms/pull/2984
- https://github.com/silverstripe/silverstripe-linkfield/pull/318
- https://github.com/silverstripe/silverstripe-elemental/pull/1232
- https://github.com/silverstripe/developer-docs/pull/551

PR #460, opened 2024-08-13
API Delete deprecated canArchive method

GuySartorelli:
For versioned records, there is no concept of "deleting" the record - it's either being archived, unpublished, or in rare cases removed from draft but left published.

Previously the `canDelete()` method has been used to check if a record can be removed from draft specifically for versioned records - which is such a rare edge case it almost doesn't bare thinking about, especially given there's no way to do that purely by interacting with the UI, nor with the web API endpoints available in core or supported modules.

Instead, `canDelete()` is now to be used to check if the record can be archived. This reduces cognitive load for developers and is a step towards simplifying the overly-complex versioning system. This is also in keeping with other mechanisms, such as `cascade_deletes` which cascades archiving for versioned records.

Note that if the record has stages and is published, we don't allow users to archive if they can't unpublish.

## Issue
- https://github.com/silverstripe/silverstripe-versioned/issues/447

### DIFF
--- a/src/GridFieldArchiveAction.php
+++ b/src/GridFieldArchiveAction.php
@@ -3,7 +3,6 @@
 namespace SilverStripe\Versioned;
 
 use SilverStripe\Control\Controller;
-use SilverStripe\Dev\Deprecation;
 use SilverStripe\Forms\GridField\GridField;
 use SilverStripe\Forms\GridField\GridField_ActionMenuItem;
 use SilverStripe\Forms\GridField\GridField_ActionProvider;
@@ -12,6 +11,7 @@ use SilverStripe\Forms\GridField\GridField_FormAction;
 use SilverStripe\Forms\GridField\GridFieldDeleteAction;
 use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\ValidationException;
+use SilverStripe\View\ViewableData;
 
 /**
  * This class is a {@link GridField} component that replaces the delete action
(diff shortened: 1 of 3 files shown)


src/GridFieldArchiveAction.php
```diff
@@ -190,13 +191,13 @@ public function handleAction(GridField $gridField, $actionName, $arguments, $dat
...
-     * @param DataObject $record
+     * @param ViewableData $record
      * @return GridField_FormAction|null
      */
     public function getArchiveAction($gridField, $record)
     {
-        /* @var DataObject|Versioned $record */
-        if (!$record->hasMethod('canArchive') || !$record->canArchive()) {
+        /** @var ViewableData|Versioned $record */
+        if (!$record->has_extension(Versioned::class) || !$record->canDelete()) {
```
ChloeSartorelli: Note that the method check used to be there to distinguish between versioned and unversioned records. `has_extension()` is the correct way to do that, so we're doing it that way instead.

src/GridFieldArchiveAction.php
```diff
@@ -190,13 +191,13 @@ public function handleAction(GridField $gridField, $actionName, $arguments, $dat
      * Returns the GridField_FormAction if archive can be performed
      *
      * @param GridField $gridField
-     * @param DataObject $record
+     * @param ViewableData $record
```
ChloeSartorelli, line 194: semi-unrelated PHPDoc fix - gridfield can accept any viewable data. This class is used to replace the delete button when versioned is installed, regardless of whether the gridfield in use has dataobjects or not.

emteknetnz (approved):

PR #461, opened 2024-08-13
API Deprecate Versioned::canArchive()

GuySartorelli:
For versioned records, there is no concept of "deleting" the record - it's either being archived, unpublished, or in rare cases removed from draft but left published.

Previously the `canDelete()` method has been used to check if a record can be removed from draft specifically for versioned records - which is such a rare edge case it almost doesn't bare thinking about, especially given there's no way to do that purely by interacting with the UI, nor with the web API endpoints available in core or supported modules.

Instead, `canDelete()` should be used to check if the record can be archived. This reduces cognitive load for developers and is a step towards simplifying the overly-complex versioning system. This is also in keeping with other mechanisms, such as `cascade_deletes` which cascades archiving for versioned records.

## Issue
- https://github.com/silverstripe/silverstripe-versioned/issues/447

### DIFF
--- a/src/GridFieldArchiveAction.php
+++ b/src/GridFieldArchiveAction.php
@@ -3,6 +3,7 @@
 namespace SilverStripe\Versioned;
 
 use SilverStripe\Control\Controller;
+use SilverStripe\Dev\Deprecation;
 use SilverStripe\Forms\GridField\GridField;
 use SilverStripe\Forms\GridField\GridField_ActionMenuItem;
 use SilverStripe\Forms\GridField\GridField_ActionProvider;
@@ -176,7 +177,8 @@ class GridFieldArchiveAction implements GridField_ColumnProvider, GridField_Acti
                 return;
             }
 
-            if (!$item->canArchive()) {
+            $canArchive = Deprecation::withNoReplacement(fn() => $item->canArchive());
+            if (!$canArchive) {
                 throw new ValidationException(
                     _t(__CLASS__ . '.ArchivePermissionsFailure', "No archive permissions")
                 );
@@ -196,7 +198,11 @@ class GridFieldArchiveAction implements GridField_ColumnProvider, GridField_Acti
     public function getArchiveAction($gridField, $record)
     {
         /* @var DataObject|Versioned $record */
-        if (!$record->hasMethod('canArchive') || !$record->canArchive()) {
+        if (!$record->hasMethod('canArchive')) {
+            return null;
+        }
+        $canArchive = Deprecation::withNoReplacement(fn() => $record->canArchive());
+        if (!$canArchive) {
             return null;
         }
 
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -1498,16 +1498,18 @@ SQL
      *
      * @param Member $member
      * @return bool
+     * @deprecated 5.3.0 Use canDelete() instead.
      */
     public function canArchive($member = null)
     {
+        Deprecation::notice('5.3.0', 'Use canDelete() instead.');
         if (!$member) {
             $member = Security::getCurrentUser();
         }
 
         // Standard mechanism for accepting permission changes from extensions
         $owner = $this->owner;
-        $extended = $owner->extendedCan('canArchive', $member);
+        $extended = Deprecation::withNoReplacement(fn() => $owner->extendedCan('canArchive', $member));
         if ($extended !== null) {
             return $extended;
         }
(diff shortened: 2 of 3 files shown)


src/Versioned.php
```diff
@@ -1498,16 +1498,18 @@ protected function extendCanUnpublish()
...
     {
+        Deprecation::notice('5.3.0', 'Use canDelete() instead.');
         if (!$member) {
             $member = Security::getCurrentUser();
         }
 
         // Standard mechanism for accepting permission changes from extensions
         $owner = $this->owner;
-        $extended = $owner->extendedCan('canArchive', $member);
+        $extended = Deprecation::withNoReplacement(fn() => $owner->extendedCan('canArchive', $member));
```
ChloeSartorelli, line 1512: `extendedCan()` will result in calling `extendCanArchive()` which is also deprecated, so we need to wrap this call too.

src/Versioned.php
```diff
@@ -1530,8 +1532,12 @@ public function canArchive($member = null)
         return true;
     }
 
+    /**
+     * @deprecated 5.3.0 Will be removed without equivalent functionality.
+     */
     protected function extendCanArchive()
     {
+        Deprecation::notice('5.3.0', 'Will be removed without equivalent functionality.');
```
ChloeSartorelli, line 1540: `protected` methods are part of our [definition of public API](https://docs.silverstripe.org/en/5/project_governance/public_api/) so we need to deprecate this even though nobody is likely to be calling it directly and it's effectively a no-op.

src/GridFieldArchiveAction.php
```diff
@@ -176,7 +177,7 @@ public function handleAction(GridField $gridField, $actionName, $arguments, $dat
                 return;
             }
 
-            if (!$item->canArchive()) {
+            if (!Deprecation::withNoReplacement(fn() => $item->canArchive())) {
```
emteknetnz:
```suggestion
            $canArchive = Deprecation::withNoReplacement(fn() => $item->canArchive());
            if (!$canArchive) {
```
Makes this a lot easier to read

Also update all other instances in this PR

emteknetnz (changes requested):

src/GridFieldArchiveAction.php
ChloeSartorelli: That's very subjective and I personally don't agree, but I'll make the change to avoid ping pong.

ChloeSartorelli: Done, fingers cross I caught them all

src/GridFieldArchiveAction.php
```diff
@@ -196,7 +198,11 @@ public function handleAction(GridField $gridField, $actionName, $arguments, $dat
     public function getArchiveAction($gridField, $record)
     {
         /* @var DataObject|Versioned $record */
-        if (!$record->hasMethod('canArchive') || !$record->canArchive()) {
+        if (!$record->hasMethod('canArchive')) {
+            return null;
+        }
+        $canArchive = !Deprecation::withNoReplacement(fn() => $record->canArchive());
+        if ($canArchive) {
```
ChloeSartorelli: Had to separate these conditions to comply with https://github.com/silverstripe/silverstripe-versioned/pull/461#discussion_r1716249601

src/GridFieldArchiveAction.php
```diff
@@ -176,7 +177,8 @@ public function handleAction(GridField $gridField, $actionName, $arguments, $dat
                 return;
             }
 
-            if (!$item->canArchive()) {
+            $canArchive = !Deprecation::withNoReplacement(fn() => $item->canArchive());
+            if ($canArchive) {
```
emteknetnz:
```suggestion
            $canArchive = Deprecation::withNoReplacement(fn() => $item->canArchive());
            if (!$canArchive) {
```

src/GridFieldArchiveAction.php
emteknetnz:
```suggestion
        $canArchive = Deprecation::withNoReplacement(fn() => $record->canArchive());
        if (!$canArchive) {
```

emteknetnz (changes requested):

src/GridFieldArchiveAction.php
ChloeSartorelli: Oops! Done

src/GridFieldArchiveAction.php
ChloeSartorelli: Oops! Done

emteknetnz (approved):

PR #462, opened 2024-08-15
API Deprecate GraphQL

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-admin/issues/1799

### DIFF
--- a/src/GraphQL/Operations/PublishCreator.php
+++ b/src/GraphQL/Operations/PublishCreator.php
@@ -3,6 +3,7 @@
 namespace SilverStripe\Versioned\GraphQL\Operations;
 
 use SilverStripe\GraphQL\Schema\Interfaces\OperationCreator;
+use SilverStripe\Dev\Deprecation;
 
 // GraphQL dependency is optional in versioned,
 // and the following implementation relies on existence of this class (in GraphQL v4)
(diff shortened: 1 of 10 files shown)


src/GraphQL/Operations/PublishCreator.php
```diff
@@ -12,10 +13,11 @@
 
 /**
  * Scaffolds a generic update operation for DataObjects.
+ *
+ * @deprecated 5.3.0 Will be removed without equivalent functionality to replace it
  */
 class PublishCreator extends AbstractPublishOperationCreator
 {
-
```
emteknetnz, line 18: Not calling deprecation method in __construct() because that's already being done in parent class AbstractPublishOperationCreator

src/GraphQL/Operations/UnpublishCreator.php
```diff
@@ -12,6 +12,8 @@
 
 /**
  * Scaffolds a generic update operation for DataObjects.
+ *
+ * @deprecated 5.3.0 Will be removed without equivalent functionality to replace it
  */
 class UnpublishCreator extends AbstractPublishOperationCreator
```
emteknetnz, line 18: Not calling deprecate in __construct() because that's already being done in AbstractPublishOperationCreator

src/GraphQL/Operations/PublishCreator.php
ChloeSartorelli: Should do it here as well, but don't call `parent::__construct()` - the one in the parent class is _for_ the parent class. We want warnings for both.
Please do that for any other classes you've avoided adding the deprecation notice for.

emteknetnz: Done

src/GraphQL/Operations/UnpublishCreator.php
emteknetnz: Have updated to include call to deprecate

GuySartorelli (approved):
LGTM

PR #481, opened 2024-11-14
FIX Ensure no admin routed links are affected by versioning

GuySartorelli:
This code was originally introduced in https://github.com/silverstripe/silverstripe-versioned/pull/173 and is intended to avoid adding stage params to any `admin/*` routed URL (which used to be theoretically restricted to LeftAndMain subclasses).

## Issue
- https://github.com/silverstripe/silverstripe-admin/issues/1762

### DIFF
--- a/tests/php/VersionedStateExtensionTest.php
+++ b/tests/php/VersionedStateExtensionTest.php
@@ -2,7 +2,11 @@
 
 namespace SilverStripe\Versioned\Tests;
 
+use SilverStripe\Admin\AdminController;
 use SilverStripe\Admin\LeftAndMain;
+use SilverStripe\Control\HTTPRequest;
+use SilverStripe\Control\Session;
+use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Dev\SapphireTest;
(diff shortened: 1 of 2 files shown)


tests/php/VersionedStateExtensionTest.php
```diff
@@ -63,18 +67,21 @@ public function testUpdateLinkRespectsQueryArgs()
...
             return;
         }
 
+        $request = new HTTPRequest('', '');
+        $request->setSession(new Session([]));
+        Injector::inst()->registerService($request, HTTPRequest::class);
         $controller = new LeftAndMain();
 
-        $liveClientConfig = $controller->getClientConfig();
+        $liveClientConfig = $controller->getCombinedClientConfig();
```
ChloeSartorelli, line 82: Using `getCombinedClientConfig` instead of `getClientConfig` to expand the breadth of the test. This should have been picked up back in https://github.com/silverstripe/silverstripe-admin/issues/1761 but wasn't because the test wasn't looking wide enough.

Note this expansion is also why we now need a request with a session above.

emteknetnz (approved):

PR #489, opened 2025-01-30
NEW Add new methods for querying data in various version modes

GuySartorelli:
Needed to convert existing CMSSiteTreeFilter implementations from using filterByCallback() to instead perform all filtering inside the database.

## Issue
- https://github.com/silverstripe/silverstripe-cms/issues/2949

### DIFF
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -474,6 +474,12 @@ class Versioned extends Extension implements TemplateGlobalProvider, Resettable
             case 'archive':
                 $this->augmentSQLVersionedArchive($query, $dataQuery);
                 break;
+            case 'archive_only':
+                $this->augmentSQLVersionedArchiveOnly($query, $dataQuery);
+                break;
+            case 'removed_from_draft':
+                $this->augmentSQLVersionedRemovedFromDraft($query, $dataQuery);
+                break;
             case 'latest_version_single':
                 $this->augmentSQLVersionedLatestSingle($query, $dataQuery);
                 break;
@@ -742,6 +748,42 @@ SQL
             20,
             $params
         );
+    }
+
+    /**
+     * Only include records which are archived - that is they have been removed from both the draft and live stages.
+     */
+    protected function augmentSQLVersionedArchiveOnly(SQLSelect $query, DataQuery $dataQuery): void
+    {
+        $baseTable = $this->baseTable();
+        $liveTable = $this->stageTable($baseTable, Versioned::LIVE);
+
+        $query->addLeftJoin(
+            $liveTable,
+            "\"{$baseTable}\".\"ID\" = \"{$liveTable}\".\"ID\""
+        );
+        $query->addWhere("\"{$liveTable}\".\"ID\" IS NULL");
+
+        $this->augmentSQLVersionedRemovedFromDraft($query, $dataQuery);
+    }
+
+    /**
+     * Only include records which are removed from draft - these might be archived but might also still be in the live stage.
+     */
+    protected function augmentSQLVersionedRemovedFromDraft(SQLSelect $query, DataQuery $dataQuery): void
+    {
+        $baseTable = $this->baseTable();
+
+        // Join a temporary alias BaseTable_Draft, renaming this on execution to BaseTable
+        // See Versioned::augmentSQLVersioned() For reference on this alias
+        $query->addLeftJoin(
+            "{$baseTable}_Draft",
+            "\"{$baseTable}\".\"ID\" = \"{$baseTable}_Draft\".\"ID\""
+        );
+
+        $query->addWhere("\"{$baseTable}_Draft\".\"ID\" IS NULL");
+
+        $this->augmentSQLVersionedLatest($query, $dataQuery);
     }
 
     /**
@@ -2479,24 +2521,29 @@ SQL
      * @template T of DataObject
      * @param class-string<T> $class The name of the class.
      * @param string $stage The name of the stage.
-     * @param string $filter A filter to be inserted into the WHERE clause.
-     * @param string $sort A sort expression to be inserted into the ORDER BY clause.
-     * @param int $limit A limit on the number of records returned from the database.
      * @param string $containerClass The container class for the result set (default is DataList)
      *
      * @return DataList<T> A modified DataList designated to the specified stage
      */
     public static function get_by_stage(
-        $class,
-        $stage,
-        $filter = '',
-        $sort = '',
-        $limit = null,
-        $containerClass = DataList::class
-    ) {
-        ReadingMode::validateStage($stage);
+        string $class,
+        string $stage,
+        string|array $filter = '',
+        string|array|null $sort = '',
+        string|array|null $limit = null,
+        string $containerClass = DataList::class
+    ): DataList {
         $result = DataObject::get($class, $filter, $sort, $limit, $containerClass);
-        return $result->setDataQueryParam([
+        return static::updateListToAlsoIncludeStage($result, $stage);
+    }
+
+    /**
+     * Update an existing DataList to include records for a given stage.
+     */
+    public static function updateListToAlsoIncludeStage(DataList $list, string $stage): DataList
+    {
+        ReadingMode::validateStage($stage);
+        return $list->setDataQueryParam([
             'Versioned.mode' => 'stage',
             'Versioned.stage' => $stage
         ]);
@@ -2784,11 +2831,11 @@ SQL
      *
      * @template T of DataObject
      * @param class-string<T> $class
-     * @param string $filter
-     * @param string $sort
+     * @param string $filter Explicitly used in where clause as raw SQL - use with caution!
+     * @param string $sort Explicitly used in orderBy clause as raw SQL - use with caution!
      * @return DataList<T>
      */
-    public static function get_including_deleted($class, $filter = "", $sort = "")
+    public static function get_including_deleted(string $class, string|array $filter = '', string $sort = ''): DataList
     {
         $list = DataList::create($class);
         if (!empty($filter)) {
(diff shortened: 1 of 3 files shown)


src/Versioned.php
```diff
@@ -2479,24 +2521,29 @@ public static function prepopulate_versionnumber_cache($class, $stage, $idList =
...
+    ): DataList {
         $result = DataObject::get($class, $filter, $sort, $limit, $containerClass);
-        return $result->setDataQueryParam([
+        return static::updateListForStage($result, $stage);
+    }
+
+    /**
+     * Update an existing DataList to only include records for a given stage.
+     */
+    public static function updateListForStage(DataList $list, string $stage): DataList
```
ChloeSartorelli: These `updateList*` methods will be used by the final set of PRs - e.g. see https://github.com/silverstripe/silverstripe-cms/pull/3052/files#diff-388e9de8adbc550a50277642109b4902208c4d028d3fdcc5a95fe81758f7bb2a

src/Versioned.php
```diff
@@ -2479,24 +2521,29 @@ public static function prepopulate_versionnumber_cache($class, $stage, $idList =
...
-        $containerClass = DataList::class
-    ) {
-        ReadingMode::validateStage($stage);
+        string $class,
+        string $stage,
+        string|array $filter = '',
+        string|array|null $sort = '',
+        string|array|null $limit = null,
+        string $containerClass = DataList::class
+    ): DataList {
```
ChloeSartorelli, line 2535: Some unrelated refactoring to add strict typing - I've done that with a few methods in this PR

emteknetnz (changes requested):
I do not think we should be adding all these new methods as as `updateListToIncludeArchivedOnly()`, `getRemovedOnDraft()` etc.  They seems like needless bloat to our API surface

They seem like they're simply wrappers around `DataList::setDataQueryParam()` to give I guess more clarity about the arguments that are later passed to `Versioned::augmentSQL()`, which I guess are not particularly obvious?

This is pretty unnecessary as you can simply call `setDataQueryParam()` from where ever you are calling things like `updateListToIncludeArchivedOnly()`.  It also creates inconsistency as we do not also have methods for the other things in `Versioned::augmentSQL()` such as `stage_unique`

GuySartorelli:
I want to keep all of the new methods I have added.

Yes, for many of these we could use call `setDataQueryParam()` and set some seemingly arbitrary params. But who will ever think to do that, and how will they know what the correct params are?

I want to update my list to only include archived records... how will I do that? Well, I know `Versioned` deals with that sort of thing so I type `Versioned::archived` in my IDE and it will show me the relevant methods. 

I want to get a new list which includes only records removed from draft. How will I do that? I type `Versioned::draft` in my IDE and it shows me the relevant methods.

Getting these lists or updating existing lists to perform these queries is hard to know how to do without a clear method to call.

src/Versioned.php
```diff
@@ -2797,8 +2844,78 @@ public static function get_including_deleted($class, $filter = "", $sort = "")
...
+            $list = $list->orderBy($orderBy);
+        }
+        return static::updateListToIncludeRemovedOnDraft($list);
+    }
+
+    /**
+     * Update a DataList to query only records which have been removed from draft.
+     * This includes archived records and records which were published but have no draft record.
+     */
+    public static function updateListToIncludeRemovedOnDraft(DataList $list): DataList
```
emteknetnz:
```suggestion
    public static function updateDataListToOnlyIncludeRemovedFromDraft(DataList $list): DataList
```
And update references to this

Just because `removed_on_draft` is a poor name shouldn't mean we need to repeat it, "removed from draft" is far more intuitive

src/Versioned.php
```diff
@@ -2797,8 +2844,78 @@ public static function get_including_deleted($class, $filter = "", $sort = "")
...
+     * Return the equivalent of a DataList::create() call, querying only records which have been removed
+     * from draft. This includes archived records and records which were published but have no draft record.
+     *
+     * @template T of DataObject
+     * @param class-string<T> $class
+     * @param string $where Explicitly used in where clause as raw SQL - use with caution!
+     * @param string $orderBy Explicitly used in orderBy clause as raw SQL - use with caution!
+     * @return DataList<T>
+     */
+    public static function getRemovedOnDraft(string $class, string|array $where = '', string $orderBy = ''): DataList
```
emteknetnz:
```suggestion
    public static function getRemovedFromDraft(string $class, string|array $where = '', string $orderBy = ''): DataList
```
And update references

src/Versioned.php
emteknetnz:
```suggestion
    public static function updateDataListToOnlyIncludeStage(DataList $list, string $stage): DataList
```
And update references to it

src/Versioned.php
```diff
@@ -2797,8 +2844,78 @@ public static function get_including_deleted($class, $filter = "", $sort = "")
...
-        return $list;
+        return static::updateListToIncludeDeleted($list);
+    }
+
+    /**
+     * Update a DataList to query the latest version of each record stored in the (class)_Versions tables.
+     *
+     * In particular, this will query deleted records as well as active ones.
+     */
+    public static function updateListToIncludeDeleted(DataList $list): DataList
```
emteknetnz:
```suggestion
    public static function updateDataListToAlsoIncludeDeleted(DataList $list): DataList
```
And update references to it

src/Versioned.php
```diff
@@ -2797,8 +2844,78 @@ public static function get_including_deleted($class, $filter = "", $sort = "")
...
+        if (!empty($orderBy)) {
+            $list = $list->orderBy($orderBy);
+        }
+        return static::updateListToIncludeArchivedOnly($list);
+    }
+
+    /**
+     * Update a DataList to query only records which are archived (i.e. removed in both draft and live)
+     */
+    public static function updateListToIncludeArchivedOnly(DataList $list): DataList
```
emteknetnz:
```suggestion
    public static function updateDataListToOnlyIncludeArchived(DataList $list): DataList
```

tests/php/VersionedTest.php
```diff
@@ -217,8 +218,10 @@ public function testForceChangeUpdatesVersion()
 
     /**
      * Test Versioned::get_including_deleted()
+     * This test seems to be testing edge cases and some rippling effects.
+     * See testGetIncludingDeleted for a pure test of results from get_including_deleted()
      */
-    public function testGetIncludingDeleted()
+    public function testGetIncludingDeletedWithOtherStuff()
```
emteknetnz, line 224: Ensure all test methods names are updated match the updated method names

emteknetnz (changes requested):
Yeah I guess they do more or less match existing methods such as Versioned::get_by_stage()

src/Versioned.php
ChloeSartorelli: Done - though `updateDataList` feels overly verbose, so kept as `updateList...`.

src/Versioned.php
ChloeSartorelli: Done

src/Versioned.php
ChloeSartorelli: Done - though `updateDataList` feels overly verbose, so kept as `updateList...`.

src/Versioned.php
ChloeSartorelli: Done - though `updateDataList` feels overly verbose, so kept as `updateList...`.

The "also" feels a bit redundant but I've kept it, I guess you intend it to constrast with the "only" from the other methods?

src/Versioned.php
ChloeSartorelli: Done - though `updateDataList` feels overly verbose, so kept as `updateList...`.

GuySartorelli: Also updated `getArchivedOnly` to `getOnlyArchived` so they're matching

tests/php/VersionedTest.php
GuySartorelli: This suggestion doesn't match any updated method name... I'm not sure what you intend with that change?

ChloeSartorelli: Done, though this one didn't change.

src/Versioned.php
ChloeSartorelli: Actually, not done. This isn't accurate. `stage_unique` (which doesn't have a corresponding method) would suit this name, but that's not how this one works (I see I had the PHPDoc wrong, so I've updated that). See the unit tests.

emteknetnz: It's not clear what this does

Depending on what it does, use one of the following:
- updateListToOnlyIncludeStage (i.e. filters existing records)
- updateListToAlsoIncludeStage (i.e. adds new records)

ChloeSartorelli: Done

src/Versioned.php
```diff
@@ -474,6 +474,12 @@ protected function augmentSQL(SQLSelect $query, ?DataQuery $dataQuery = null)
             case 'archive':
                 $this->augmentSQLVersionedArchive($query, $dataQuery);
                 break;
+            case 'archive_only':
+                $this->augmentSQLVersionedArchiveOnly($query, $dataQuery);
+                break;
+            case 'removed_on_draft':
+                $this->augmentSQLVersionedRemovedOnDraft($query, $dataQuery);
```
emteknetnz:
```suggestion
            case 'removed_on_draft':
                $this->augmentSQLVersionedRemovedFromDraft($query, $dataQuery);
```
Sorry just realised these are newly added.  Need to change the 'on' to 'from'

src/Versioned.php
```diff
@@ -744,6 +750,42 @@ protected function augmentSQLVersionedArchive(SQLSelect $query, DataQuery $dataQ
...
+        $baseTable = $this->baseTable();
+        $liveTable = $this->stageTable($baseTable, Versioned::LIVE);
+
+        $query->addLeftJoin(
+            $liveTable,
+            "\"{$baseTable}\".\"ID\" = \"{$liveTable}\".\"ID\""
+        );
+        $query->addWhere("\"{$liveTable}\".\"ID\" IS NULL");
+
+        $this->augmentSQLVersionedRemovedOnDraft($query, $dataQuery);
```
emteknetnz:
```suggestion
        $this->augmentSQLVersionedRemovedFromDraft($query, $dataQuery);
```

src/Versioned.php
```diff
@@ -744,6 +750,42 @@ protected function augmentSQLVersionedArchive(SQLSelect $query, DataQuery $dataQ
...
+    }
+
+    /**
+     * Only include records which are removed from draft - these might be archived but might also still be in the live stage.
+     */
+    protected function augmentSQLVersionedRemovedOnDraft(SQLSelect $query, DataQuery $dataQuery): void
```
emteknetnz:
```suggestion
    protected function augmentSQLVersionedRemovedFromDraft(SQLSelect $query, DataQuery $dataQuery): void
```

src/Versioned.php
```diff
@@ -2797,8 +2844,78 @@ public static function get_including_deleted($class, $filter = "", $sort = "")
...
+        return static::updateListToOnlyIncludeRemovedFromDraft($list);
+    }
+
+    /**
+     * Update a DataList to query only records which have been removed from draft.
+     * This includes archived records and records which were published but have no draft record.
+     */
+    public static function updateListToOnlyIncludeRemovedFromDraft(DataList $list): DataList
+    {
+        return $list->setDataQueryParam('Versioned.mode', 'removed_on_draft');
```
emteknetnz:
```suggestion
        return $list->setDataQueryParam('Versioned.mode', 'removed_from_draft');
```

emteknetnz (changes requested):

src/Versioned.php
ChloeSartorelli: Oh. I thought you have omitted that last time so that these protected methods could match the string they're using.... will update

ChloeSartorelli: Done. Please tell me this is the last cycle of name changes.

src/Versioned.php
ChloeSartorelli: Done

src/Versioned.php
ChloeSartorelli: Done - please don't flag each instance of the same named thing in the future, it means I have to check each one and go "is this just him saying the same thing or is this a different method?"

src/Versioned.php
ChloeSartorelli: Done

GuySartorelli:
Done - didn't bother rerunning sink for this as the things renamed this time around are only ever referenced in this module directly.

emteknetnz (approved):

PR #500, opened 2025-04-24
ENH Better version number caching

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-framework/issues/11703

Needed for https://github.com/silverstripe/silverstripe-asset-admin/pull/1560 and https://github.com/silverstripe/silverstripe-versioned-admin/pull/395

Versioned is a huge file, so I've moved the x2 types of caching to a pair of new traits

### DIFF
--- a/_config/versionedcache.yml
+++ b/_config/versionedcache.yml
@@ -9,3 +9,6 @@ SilverStripe\Core\Injector\Injector:
     constructor:
       args:
         container: 'SilverStripe\Versioned\Caching\VersionedCacheAdapter'
+SilverStripe\ORM\DataList:
+  extensions:
+    - SilverStripe\Versioned\Extensions\DataListExtension
--- a/src/Extensions/DataListExtension.php
+++ b/src/Extensions/DataListExtension.php
@@ -1,0 +1,24 @@
+<?php
+
+namespace SilverStripe\Versioned\Extensions;
+
+use SilverStripe\Core\Extension;
+use SilverStripe\ORM\DataList;
+use SilverStripe\Versioned\Versioned;
+
+/**
+ * @extends Extension<DataList>
+ */
+class DataListExtension extends Extension
+{
+    protected function onPrepopulateCaches(array $ids): void
+    {
+        $dataClass = $this->getOwner()->dataClass();
+        if (!$dataClass::has_extension(Versioned::class)) {
+            return;
+        }
+        // Version number cache
+        Versioned::prepopulateVersionNumberCache($dataClass, $ids);
+        Versioned::prepopulateVersionsCache($dataClass, $ids);
+    }
+}
--- a/src/Traits/VersionNumberCacheTrait.php
+++ b/src/Traits/VersionNumberCacheTrait.php
@@ -1,0 +1,119 @@
+<?php
+
+namespace SilverStripe\Versioned\Traits;
+
+use InvalidArgumentException;
+use SilverStripe\Core\Config\Config;
+use SilverStripe\ORM\DataList;
+use SilverStripe\ORM\DataObject;
+use SilverStripe\ORM\DB;
+use SilverStripe\Versioned\ReadingMode;
+use SilverStripe\Versioned\Versioned;
+
+/**
+ * This trait exists to help encapsulate specific functionality for the Versioned class.
+ * Its use with any other class is not supported.
+ * @internal
+ */
+trait VersionNumberCacheTrait
+{
+    /**
+     * A cache used by get_versionnumber_by_stage().
+     * Clear through {@link flushCache()}.
+     * A value of 0 in [$baseClass][$stage][$id] means a record with that ID is not in that stage.
+     *
+     * @var array
+     */
+    protected static $cache_versionnumber;
+
+    /**
+     * Used to enable or disable the prepopulation of the version number cache.
+     * Set this configuration on the extension, not on the class being extended. For example:
+     * <code>
+     * SilverStripe\Versioned\Versioned:
+     *    prepopulate_versionnumber_cache: false
+     * </code>
+     * Defaults to true.
+     *
+     * @var boolean
+     */
+    private static $prepopulate_versionnumber_cache = true;
+
+    /**
+     * Shorthand method to call prepopulateVersionNumberCacheForStage() for all applicable stages
+     * This caches the version numbers of a particular DataObject class
+     * Not to be confused with prepopulateVersionsCache()
+     */
+    public static function prepopulateVersionNumberCache(string $dataClass, ?array $ids = null): void
+    {
+        $stages = [Versioned::DRAFT];
+        if ($dataClass::has_extension(Versioned::class)) {
+            $stages[] = Versioned::LIVE;
+        }
+        foreach ($stages as $stage) {
+            self::class::prepopulateVersionNumberCacheForStage($dataClass, $stage, $ids);
+        }
+    }
+
+    /**
+     * Pre-populate the cache for Versioned::get_versionnumber_by_stage() for
+     * a list of record IDs, for more efficient database querying.
+     * If the $ids param is null, then every record will be cached.
+     */
+    public static function prepopulateVersionNumberCacheForStage(
+        string $dataClass,
+        string $stage,
+        ?array $ids = null,
+    ): void {
+        $ids ??= [];
+        ReadingMode::validateStage($stage);
+        if (!Config::inst()->get(static::class, 'prepopulate_versionnumber_cache')) {
+            return;
+        }
+        $singleton = DataObject::singleton($dataClass);
+        $baseClass = $singleton->baseClass();
+        $baseTable = $singleton->baseTable();
+        $stageTable = $singleton->stageTable($baseTable, $stage);
+        $usePlaceholders = DataList::config()->get('use_placeholders_for_integer_ids');
+        $parameters = null;
+        $filter = '';
+        if (count($ids)) {
+            // Check for non int IDs to protected against SQL injection
+            $notInts = array_filter($ids, fn($id) => !ctype_digit((string) $id) || $id != (int) $id);
+            if ($usePlaceholders || count($notInts)) {
+                $filter = 'WHERE "ID" IN (' . DB::placeholders($ids) . ')';
+                $parameters = $ids;
+            } else {
+                $filter = 'WHERE "ID" IN (' . implode(', ', $ids) . ')';
+            }
+        } else {
+            // If we are caching IDs for _all_ records then we can mark this cache as "complete" and in the case of a cache-miss
+            // no subsequent call is necessary
+            self::class::$cache_versionnumber[$baseClass][$stage] = [ '_complete' => true ];
+        }
+        $sql = "SELECT \"ID\", \"Version\" FROM \"$stageTable\" $filter";
+        if ($usePlaceholders) {
+            $versions = DB::prepared_query($sql, $parameters)->map();
+        } else {
+            $versions = DB::query($sql)->map();
+        }
+        if ($ids) {
+            foreach ($ids as $id) {
+                $version = 0;
+                foreach ($versions as $vid => $val) {
+                    if ($id === $vid) {
+                        $version = $val;
+                        break;
+                    }
+                }
+                self::class::$cache_versionnumber[$baseClass][$stage][$id] = $version;
+            }
+        } else {
+            foreach ($versions as $id => $version) {
+                self::class::$cache_versionnumber[$baseClass][$stage][$id] = $version;
+            }
+        }
+        $object = DataObject::singleton($dataClass);
+        $object->invokeWithExtensions('updatePrePopulateVersionNumberCache', $versions, $dataClass, $stage, $ids);
+    }
+}
--- a/src/Traits/VersionsCacheTrait.php
+++ b/src/Traits/VersionsCacheTrait.php
@@ -1,0 +1,155 @@
+<?php
+
+namespace SilverStripe\Versioned\Traits;
+
+use SilverStripe\Core\Config\Config;
+use SilverStripe\ORM\DataObject;
+use SilverStripe\ORM\DB;
+use SilverStripe\Model\List\ArrayList;
+use SilverStripe\Versioned\Versioned;
+use SilverStripe\Versioned\Versioned_Version;
+use SilverStripe\ORM\DataList;
+
+/**
+ * Functionality to cache Versioned::Versions(), which is used for the <DataObject>_Versions tables
+ * This trait exists to help encapsulate specific functionality for the Versioned class.
+ * Its use with any other class is not supported.
+ * @internal
+ */
+trait VersionsCacheTrait
+{
+    /**
+     * In-memory cache of Versions i.e. the <DataObject>_Versions table
+     * @internal
+     */
+    private static array $versions_cache = [];
+
+    /**
+     * Prepopulate the in-memory Versions cache, which are used for <DataObject>_Versions tables
+     * The values will be retrieved using an efficient `WHERE "ID" IN (<ids>)` SQL query
+     * Note method this should not to be confused with prepopulateVersionNumberCache()
+     */
+    public static function prepopulateVersionsCache(string $baseDataClass, array $ids): void
+    {
+        $cachedIDs = [];
+        foreach ($ids as $id) {
+            $key = self::class::generateVersionsCacheKey($id);
+            if ($key && array_key_exists($key, self::class::$versions_cache)) {
+                $cachedIDs[] = $id;
+            }
+        }
+        // Only fetch uncached IDs
+        $ids = array_diff($ids, $cachedIDs);
+        if (empty($ids)) {
+            return;
+        }
+        $versions = self::class::innerReadVersions($baseDataClass, $ids);
+        foreach ($ids as $id) {
+            $key = self::class::generateVersionsCacheKey($id);
+            if ($key) {
+                $vers = array_filter($versions, fn($ver) => (int) $ver->RecordID === $id);
+                self::class::$versions_cache[$key] = ArrayList::create($vers);
+            }
+        }
+    }
+
+    /**
+     * Get an ArrayList of Versions for an ID
+     *
+     * @return ArrayList<Versioned_Version>
+     */
+    private static function readVersions(
+        DataObject $owner,
+        mixed $filter,
+        mixed $sort,
+        mixed $limit,
+        mixed $join,
+    ): ArrayList {
+        // When an object is not yet in the Database, we can't get its versions
+        if (!$owner->isInDB()) {
+            return ArrayList::create();
+        }
+        $id = $owner->ID;
+        $key = self::class::generateVersionsCacheKey($id, $filter, $sort, $limit, $join);
+        if ($key && array_key_exists($key, self::class::$versions_cache)) {
+            return self::class::$versions_cache[$key];
+        }
+        $baseDataClass = DataObject::getSchema()->baseDataClass($owner);
+        $versions = self::class::innerReadVersions($baseDataClass, [$id], $filter, $sort, $limit);
+        $list = ArrayList::create($versions);
+        return $list;
+    }
+
+    /**
+     * Internal shared logic used by prepopulateVersionsCache() and readVersionsID()
+     */
+    private static function innerReadVersions(
+        string $baseDataClass,
+        array $ids,
+        mixed $filter = '',
+        mixed $sort = '',
+        mixed $limit = '',
+    ): array {
+        $oldMode = Versioned::get_reading_mode();
+        Versioned::set_stage(Versioned::DRAFT);
+        $list = DataObject::get($baseDataClass, $filter, $sort, $limit);
+        $query = $list->dataQuery()->query();
+        $baseTable = null;
+        foreach ($query->getFrom() as $table => $tableJoin) {
+            if (is_string($tableJoin) && $tableJoin[0] == '"') {
+                $baseTable = str_replace('"', '', $tableJoin ?? '');
+            } elseif (is_string($tableJoin) && substr($tableJoin ?? '', 0, 5) != 'INNER') {
+                $query->setFrom([
+                    $table => "LEFT JOIN \"$table\" ON \"$table\".\"RecordID\"=\"{$baseTable}_Versions\".\"RecordID\""
+                        . " AND \"$table\".\"Version\" = \"{$baseTable}_Versions\".\"Version\""
+                ]);
+            }
+            $query->renameTable($table, $table . '_Versions');
+        }
+        // Add all <basetable>_Versions columns
+        foreach (array_keys(Config::inst()->get(Versioned::class, 'db_for_versions_table')) as $name) {
+            $query->selectField(sprintf('"%s_Versions"."%s"', $baseTable, $name), $name);
+        }
+        $usePlaceholders = DataList::config()->get('use_placeholders_for_integer_ids');
+        // Check for non int IDs to protected against SQL injection
+        $notInts = array_filter($ids, fn($id) => !ctype_digit((string) $id) || $id != (int) $id);
+        if ($usePlaceholders || count($notInts)) {
+            $in = DB::placeholders($ids);
+            $query->addWhere(["\"{$baseTable}_Versions\".\"RecordID\" IN ($in)" => $ids]);
+        } else {
+            $in = implode(', ', $ids);
+            $query->addWhere("\"{$baseTable}_Versions\".\"RecordID\" IN ($in)");
+        }
+        $query->setOrderBy(($sort)
+            ? $sort
+            : "\"{$baseTable}_Versions\".\"LastEdited\" DESC, \"{$baseTable}_Versions\".\"Version\" DESC");
+        $records = $query->execute();
+        $versions = [];
+        foreach ($records as $record) {
+            $versions[] = Versioned_Version::create($record);
+        }
+        Versioned::set_reading_mode($oldMode);
+        return $versions;
+    }
+
+    /**
+     * Generate a key for $versions_cache
+     * For the mixed params in this method, ensure they are seralizable before passing them to this method
+     * @return string|false The key as a string or false if a key was unable to be generated
+     */
+    private static function generateVersionsCacheKey(
+        int $id,
+        mixed $filter = '',
+        mixed $sort = '',
+        mixed $limit = '',
+        mixed $join = '',
+    ): string|false {
+        $keyInputs = [$filter, $sort, $limit, $join];
+        $isSerializable = fn(mixed $var) => @serialize($var) !== false;
+        $allSerializable = count(array_filter($keyInputs, $isSerializable)) === count($keyInputs);
+        if (!$allSerializable) {
+            return false;
+        }
+        return $id . '_' . md5(implode('', array_map('serialize', $keyInputs)));
+    }
+}
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -14,7 +14,6 @@ use SilverStripe\Core\Extension;
 use SilverStripe\Core\Injector\Injector;
 use SilverStripe\Core\Resettable;
 use SilverStripe\Forms\FieldList;
-use SilverStripe\Model\List\ArrayList;
 use SilverStripe\ORM\DataList;
 use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\DataQuery;
@@ -25,6 +24,8 @@ use SilverStripe\Security\Member;
 use SilverStripe\Security\Permission;
 use SilverStripe\Security\Security;
 use SilverStripe\View\TemplateGlobalProvider;
+use SilverStripe\Versioned\Traits\VersionsCacheTrait;
+use SilverStripe\Versioned\Traits\VersionNumberCacheTrait;
 
 /**
  * The Versioned extension allows your DataObjects to have several versions,
@@ -40,6 +41,9 @@ use SilverStripe\View\TemplateGlobalProvider;
  */
 class Versioned extends Extension implements TemplateGlobalProvider, Resettable
 {
+    use VersionsCacheTrait;
+    use VersionNumberCacheTrait;
+
     /**
      * Versioning mode for this object.
      * Note: Not related to the current versioning mode in the state / session
@@ -74,15 +78,6 @@ class Versioned extends Extension implements TemplateGlobalProvider, Resettable
      * The draft (default) stage
      */
     const DRAFT = 'Stage';
-
-    /**
-     * A cache used by get_versionnumber_by_stage().
-     * Clear through {@link flushCache()}.
-     * version (int)0 means not on this stage.
-     *
-     * @var array
-     */
-    protected static $cache_versionnumber;
 
     /**
      * Set if draft site is secured or not. Fails over to
@@ -190,15 +185,6 @@ class Versioned extends Extension implements TemplateGlobalProvider, Resettable
     private static $db = [
         'Version' => 'Int'
     ];
-
-    /**
-     * Used to enable or disable the prepopulation of the version number cache.
-     * Defaults to true.
-     *
-     * @config
-     * @var boolean
-     */
-    private static $prepopulate_versionnumber_cache = true;
 
     /**
      * Indicates whether augmentSQL operations should add subselects as WHERE conditions instead of INNER JOIN
@@ -2073,54 +2059,8 @@ SQL
      */
     public function Versions($filter = "", $sort = "", $limit = "", $join = "")
     {
-        $owner = $this->owner;
-
-        // When an object is not yet in the Database, we can't get its versions
-        if (!$owner->isInDB()) {
-            return ArrayList::create();
-        }
-
-        // Make sure the table names are not postfixed (e.g. _Live)
-        $oldMode = static::get_reading_mode();
-        static::set_stage(static::DRAFT);
-
-        $list = DataObject::get(DataObject::getSchema()->baseDataClass($owner), $filter, $sort, $limit);
-
-        $query = $list->dataQuery()->query();
-
-        $baseTable = null;
-        foreach ($query->getFrom() as $table => $tableJoin) {
-            if (is_string($tableJoin) && $tableJoin[0] == '"') {
-                $baseTable = str_replace('"', '', $tableJoin ?? '');
-            } elseif (is_string($tableJoin) && substr($tableJoin ?? '', 0, 5) != 'INNER') {
-                $query->setFrom([
-                    $table => "LEFT JOIN \"$table\" ON \"$table\".\"RecordID\"=\"{$baseTable}_Versions\".\"RecordID\""
-                        . " AND \"$table\".\"Version\" = \"{$baseTable}_Versions\".\"Version\""
-                ]);
-            }
-            $query->renameTable($table, $table . '_Versions');
-        }
-
-        // Add all <basetable>_Versions columns
-        foreach (Config::inst()->get(static::class, 'db_for_versions_table') as $name => $type) {
-            $query->selectField(sprintf('"%s_Versions"."%s"', $baseTable, $name), $name);
-        }
-
-        $query->addWhere([
-            "\"{$baseTable}_Versions\".\"RecordID\" = ?" => $owner->ID
-        ]);
-        $query->setOrderBy(($sort) ? $sort
-            : "\"{$baseTable}_Versions\".\"LastEdited\" DESC, \"{$baseTable}_Versions\".\"Version\" DESC");
-
-        $records = $query->execute();
-        $versions = new ArrayList();
-
-        foreach ($records as $record) {
-            $versions->push(new Versioned_Version($record));
-        }
-
-        Versioned::set_reading_mode($oldMode);
-        return $versions;
+        $owner = $this->getOwner();
+        return Versioned::readVersions($owner, $filter, $sort, $limit, $join);
     }
 
     /**
@@ -2507,55 +2447,17 @@ SQL
     }
 
     /**
-     * Pre-populate the cache for Versioned::get_versionnumber_by_stage() for
-     * a list of record IDs, for more efficient database querying.  If $idList
-     * is null, then every record will be pre-cached.
+     * Alias of prepopulateVersionNumberCacheForStage()
      *
      * @param string $class
      * @param string $stage
      * @param array $idList
+     *
+     * @deprecated 6.1.0 Use prepopulateVersionNumberCacheForStage() instead
      */
     public static function prepopulate_versionnumber_cache($class, $stage, $idList = null)
     {
-        ReadingMode::validateStage($stage);
-        if (!Config::inst()->get(static::class, 'prepopulate_versionnumber_cache')) {
-            return;
-        }
-
-        $singleton = DataObject::singleton($class);
-        $baseClass = $singleton->baseClass();
-        $baseTable = $singleton->baseTable();
-        $stageTable = $singleton->stageTable($baseTable, $stage);
-
-        $filter = "";
-        $parameters = [];
-        if ($idList) {
-            // Validate the ID list
-            foreach ($idList as $id) {
-                if (!is_numeric($id)) {
-                    throw new InvalidArgumentException(
-                        "Bad ID passed to Versioned::prepopulate_versionnumber_cache() in \$idList: " . $id
-                    );
-                }
-            }
-            $filter = 'WHERE "ID" IN (' . DB::placeholders($idList) . ')';
-            $parameters = $idList;
-
-        // If we are caching IDs for _all_ records then we can mark this cache as "complete" and in the case of a cache-miss
-        // no subsequent call is necessary
-        } else {
-            Versioned::$cache_versionnumber[$baseClass][$stage] = [ '_complete' => true ];
-        }
-
-        $versions = DB::prepared_query("SELECT \"ID\", \"Version\" FROM \"$stageTable\" $filter", $parameters)->map();
-
-        foreach ($versions as $id => $version) {
-            Versioned::$cache_versionnumber[$baseClass][$stage][$id] = $version;
-        }
-
-        $className = $class instanceof DataObject ? $class->ClassName : $class;
-        $object = DataObject::singleton($className);
-        $object->invokeWithExtensions('updatePrePopulateVersionNumberCache', $versions, $class, $stage, $idList);
+        Versioned::prepopulateVersionNumberCacheForStage($class, $stage, $idList);
     }
 
     /**
@@ -3036,6 +2938,7 @@ SQL
     protected function onFlushCache()
     {
         Versioned::$cache_versionnumber = [];
+        Versioned::$versions_cache = [];
         $this->versionModifiedCache = [];
     }
 
--- a/tests/php/Traits/TestObject.php
+++ b/tests/php/Traits/TestObject.php
@@ -1,0 +1,20 @@
+<?php
+
+namespace SilverStripe\Versioned\Tests\Traits;
+
+use SilverStripe\ORM\DataObject;
+use SilverStripe\Versioned\Versioned;
+use SilverStripe\Dev\TestOnly;
+
+class TestObject extends DataObject implements TestOnly
+{
+    private static $table_name = 'Versioned_Tests_Traits_TestObject';
+
+    private static array $db = [
+        'Title' => 'Varchar'
+    ];
+
+    private static array $extensions = [
+        Versioned::class,
+    ];
+}
--- a/tests/php/Traits/TestTraitHolder.php
+++ b/tests/php/Traits/TestTraitHolder.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace SilverStripe\Versioned\Tests\Traits;
+
+use SilverStripe\Versioned\Traits\VersionNumberCacheTrait;
+use SilverStripe\Versioned\Traits\VersionsCacheTrait;
+use SilverStripe\Dev\TestOnly;
+
+class TestTraitHolder implements TestOnly
+{
+    use VersionNumberCacheTrait;
+    use VersionsCacheTrait;
+}
--- a/tests/php/Traits/VersionNumberCacheTraitTest.php
+++ b/tests/php/Traits/VersionNumberCacheTraitTest.php
@@ -1,0 +1,166 @@
+<?php
+
+namespace SilverStripe\Versioned\Tests\Traits;
+
+use ReflectionProperty;
+use PHPUnit\Framework\Attributes\DataProvider;
+use SilverStripe\Dev\SapphireTest;
+use SilverStripe\Versioned\Versioned;
+
+class VersionNumberCacheTraitTest extends SapphireTest
+{
+    protected static $fixture_file = 'VersionNumberCacheTraitTest.yml';
+
+    protected static $extra_dataobjects = [
+        TestObject::class,
+    ];
+
+    protected function tearDown(): void
+    {
+        $this->resetCache();
+        parent::tearDown();
+    }
+
+    public function testCacheNullByDefault(): void
+    {
+        $this->assertSame(null, $this->getCache());
+    }
+
+    public static function providePrepopulateVersionNumberCacheForStage(): array
+    {
+        return [
+            'draft-all-no-publish' => [
+                'stage' => Versioned::DRAFT,
+                'ids' => [],
+                'publishAll' => false,
+                'expected' => [
+                    TestObject::class => [
+                        Versioned::DRAFT => [
+                            '_complete' => true,
+                            1 => 1,
+                            2 => 1,
+                        ]
+                    ]
+                ],
+            ],
+            'live-all-no-publish' => [
+                'stage' => Versioned::LIVE,
+                'ids' => [],
+                'publishAll' => false,
+                'expected' => [
+                    TestObject::class => [
+                        Versioned::LIVE => [
+                            '_complete' => true,
+                        ],
+                    ]
+                ],
+            ],
+            'draft-all-publish' => [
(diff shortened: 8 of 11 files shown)


emteknetnz:
Replaced by https://github.com/silverstripe/silverstripe-versioned/pull/502

src/Versioned.php
```diff
@@ -2504,16 +2519,39 @@ public static function prepopulate_versionnumber_cache($class, $stage, $idList =
         } else {
             Versioned::$cache_versionnumber[$baseClass][$stage] = [ '_complete' => true ];
         }
-
         $versions = DB::prepared_query("SELECT \"ID\", \"Version\" FROM \"$stageTable\" $filter", $parameters)->map();
-
-        foreach ($versions as $id => $version) {
-            Versioned::$cache_versionnumber[$baseClass][$stage][$id] = $version;
+        if ($idList) {
```
emteknetnz: This part is something of a bugfix

Missing versions e.g. unpublished live records, aren't cached as zero, therefore the caching doesn't actually do anything

0 is the existing way to denote missing records https://github.com/silverstripe/silverstripe-versioned/blob/2/src/Versioned.php#L2448

Note this "bugfix" could in theory target 2.4 instead of 3, however nothing in sink actually called prepopulate_versionnumber_cache() originally in sink 5.4 so easier to just do in this PR, otherwise a bunch of duplicate sql queries keep showing up which makes it harder to validate that things like asset admin queries are being cached properly

src/Extensions/DataListExtension.php
```diff
@@ -0,0 +1,26 @@
...
+namespace SilverStripe\Versioned\Extensions;
+
+use RuntimeException;
+use SilverStripe\Core\Extension;
+use SilverStripe\ORM\DataObject;
+use SilverStripe\Versioned\Versioned;
+
+class DataListExtension extends Extension
+{
+    public function updatePrepopulateCaches(string &$dataClass, array &$ids): void
```
emteknetnz: This implements this [version hook](https://github.com/silverstripe/silverstripe-framework/pull/11710/files#diff-9f98a9c32b63e751ae7323470ecb6c15c135df9705e0431d9f1ccf0bd15a26d6R2072)

src/Versioned.php
```diff
@@ -2460,39 +2589,54 @@ protected function onPrepopulateTreeDataCache($recordList = null, array $options
...
-     * is null, then every record will be pre-cached.
-     *
-     * @param string $class
-     * @param string $stage
-     * @param array $idList
+     * Shorthand method to call prepopulateVersionNumberCacheForStage() for each applicable stage
+     * Not to be confused with prepopulateVersionsCache()
      */
-    public static function prepopulate_versionnumber_cache($class, $stage, $idList = null)
+    public static function prepopulateVersionNumberCache(string $dataClass, ?array $idList = null): void
```
emteknetnz, line 2600: This is intended as replacement for prepopulate_versionnumber_cache() which has now been deprecated

src/Versioned.php
```diff
@@ -2460,39 +2589,54 @@ protected function onPrepopulateTreeDataCache($recordList = null, array $options
...
+            Versioned::prepopulateVersionNumberCacheForStage($dataClass, $stage, $idList);
+        }
+    }
+
+    /**
+     * Pre-populate the cache for Versioned::get_versionnumber_by_stage() for
+     * a list of record IDs, for more efficient database querying.
+     * If $idList is null, then every record will be pre-cached.
+     */
+    public static function prepopulateVersionNumberCacheForStage(
```
emteknetnz, line 2616: This is intended as replacement for prepopulate_versionnumber_cache() which has now been deprecated

src/Versioned.php
```diff
@@ -2023,26 +2010,64 @@ public function stagesDifferRecursive(): bool
...
+    }
+
+    /**
+     * Shared internal logic which allows passing either a single ID for Versions()
+     * or passing an array of IDs for VersionsCachingTrait::prepopulateVersionsCache()
+     *
+     * @param int|array $idOrIds A single record ID, or an array of record IDs
+     * @return ArrayList<Versioned_Version>|array<ArrayList<Versioned_Version>> an ArrayList when $idOrIds is an int, an array when $idOrIds is an array
+     */
+    private static function sharedVersions(
```
emteknetnz, line 2028: This is code that used to be in Versions()

It's been turned into a shared private method method as VersionsCachingTrait::prepopulateVersionsCache() also uses its log

It has an int|array $idOrIds argument. The intention is that VersionsCachingTrait::prepopulateVersionsCache() will first call it with an array of ids, and a `WHERE "ID" IN (?, ?, ?)` query is used and the cache will be prepoputlated

The later when regular code calls Versions() then a single int ID will be passed and retrieve data from the cache

src/Traits/VersionNumberCachingTrait.php
```diff
@@ -0,0 +1,123 @@
...
+            self::class::prepopulateVersionNumberCacheForStage($dataClass, $stage, $idList);
+        }
+    }
+
+    /**
+     * Pre-populate the cache for Versioned::get_versionnumber_by_stage() for
+     * a list of record IDs, for more efficient database querying.
+     * If $idList is null, then every record will be cached.
+     */
+    public static function prepopulateVersionNumberCacheForStage(
```
emteknetnz, line 53: This was originally Versioned::prepopulate_versionnumber_cache()

src/Traits/VersionNumberCachingTrait.php
```diff
@@ -0,0 +1,123 @@
...
+    /**
+     * Alias of prepopulateVersionNumberCacheForStage()
+     *
+     * @param string $class
+     * @param string $stage
+     * @param array $idList
+     *
+     * @deprecated 6.1.0 Use prepopulateVersionNumberCacheForStage() instead
+     */
+    public static function prepopulate_versionnumber_cache($class, $stage, $idList = null)
```
emteknetnz, line 119: Renaming and deprecating this because
- It uses a pretty old fashioned method naming convention
- There's a private static that's we're keeping that is also called `prepopulate_versionnumber_cache`

src/Versioned.php
```diff
@@ -2460,60 +2396,21 @@ protected function onPrepopulateTreeDataCache($recordList = null, array $options
...
-     * Pre-populate the cache for Versioned::get_versionnumber_by_stage() for
-     * a list of record IDs, for more efficient database querying.  If $idList
-     * is null, then every record will be pre-cached.
+     * Alias of prepopulateVersionNumberCacheForStage()
      *
      * @param string $class
      * @param string $stage
      * @param array $idList
+     *
+     * @deprecated 6.1.0 Use prepopulateVersionNumberCacheForStage() instead
```
emteknetnz, line 2456: Renaming and deprecating this because
- It uses a pretty old fashioned method naming convention
- I've added strong typing to the new method
- There's a private static that's we're keeping that is also called `prepopulate_versionnumber_cache`

src/Traits/VersionNumberCacheTrait.php
```diff
@@ -0,0 +1,109 @@
...
+            self::class::prepopulateVersionNumberCacheForStage($dataClass, $stage, $idList);
+        }
+    }
+
+    /**
+     * Pre-populate the cache for Versioned::get_versionnumber_by_stage() for
+     * a list of record IDs, for more efficient database querying.
+     * If $idList is null, then every record will be cached.
+     */
+    public static function prepopulateVersionNumberCacheForStage(
```
emteknetnz, line 63: This was originally Versioned::prepopulate_versionnumber_cache()

src/Traits/VersionsCacheTrait.php
```diff
@@ -0,0 +1,197 @@
...
+    }
+
+    /**
+     * Shared internal logic which allows passing either a single ID for Versioned::Versions()
+     * or passing an array of IDs for VersionsCachingTrait::prepopulateVersionsCache()
+     *
+     * @param int|array $idOrIds A single record ID, or an array of record IDs
+     * @return ArrayList<Versioned_Version>|array<ArrayList<Versioned_Version>> an ArrayList when $idOrIds is an int, an array when $idOrIds is an array
+     */
+    private static function readVersions(
```
emteknetnz, line 61: This is code that used to be in `Versioned::Versions()`

It's been turned into a shared private method method as VersionsCachingTrait::prepopulateVersionsCache() also uses its log

It has an int|array $idOrIds argument. The intention is that VersionsCachingTrait::prepopulateVersionsCache() will first call it with an array of ids, and a WHERE "ID" IN (?, ?, ?) query is used and the cache will be prepoputlated

The later when regular code calls Versions() then a single int ID will be passed and retrieve data from the cache

src/Extensions/DataListExtension.php
ChloeSartorelli:
```suggestion
    public function updatePrepopulateCaches(string $dataClass, array $ids): void
```
We don't actually want to change the upstream values of these, so we shouldn't pass by reference.

src/Versioned.php
ChloeSartorelli: There's quite a few of these snake case methods in `Versioned` - I don't personally think there's a strong case for renaming this method here in isolation. That said, I think there's a reasonable likelyhood that for CMS 7 we'll want to go for PSR12 finally which would mean deprecating and renaming a bunch of these anyway, so I'm okay with this one being done early.

_config/versionedextension.yml
```diff
@@ -20,3 +20,9 @@ Name: versioned-table
 SilverStripe\ORM\DataQuery:
   extensions:
     - SilverStripe\Versioned\VersionedTableDataQueryExtension
+---
+Name: versioned-extensions
+---
+SilverStripe\ORM\DataList:
+  extensions:
+    - SilverStripe\Versioned\Extensions\DataListExtension
```
ChloeSartorelli: Feels more like this should go in `versionedcache.yml` since it's related to Versioned cache?

My assumption is that `versionedextension.yml` was meant to be just the injector config for the `Versioned` extension.

The dataquery extension was probably added here because someone was "extension" in the file name and went "Oh I guess this is where the extensions config goes" but I don't think it should have been added here either.

src/Extensions/DataListExtension.php
```diff
@@ -0,0 +1,26 @@
...
+    {
+        if (!is_a($dataClass, DataObject::class, true)) {
+            throw new RuntimeException("Class $dataClass is not a DataObject");
+        }
```
ChloeSartorelli: Is it possible for a `DataList` to have a data class that isn't a `DataObject`? I thought that was just not possible?

src/Traits/VersionsCacheTrait.php
```diff
@@ -0,0 +1,197 @@
...
+        return self::class::$useVersionsCache;
+    }
+
+    /**
+     * Set whether to use the Versions cache
+     */
+    public static function setUseVersionsCache(bool $use): void
+    {
+        self::class::$useVersionsCache = $use;
+    }
```
ChloeSartorelli: If we have cached data, when would we want to explicitly _not_ use it?

With the way `DataListExtension` is implemented, this will be true whenever we're prepopulating version cache, won't it? When do we want to turn it off? And... why aren't we turning if off in those scenarios?

src/Traits/VersionNumberCacheTrait.php
```diff
@@ -0,0 +1,109 @@
...
+use SilverStripe\ORM\DB;
+use SilverStripe\Versioned\ReadingMode;
+use SilverStripe\Versioned\Versioned;
+
+trait VersionNumberCacheTrait
+{
+    /**
+     * A cache used by get_versionnumber_by_stage().
+     * Clear through {@link flushCache()}.
+     * version (int)0 means not on this stage.
```
ChloeSartorelli:
```suggestion
     * A value of 0 in [$baseClass][$stage][$id] means a record with that ID is not in that stage.
```

src/Traits/VersionNumberCacheTrait.php
```diff
@@ -0,0 +1,109 @@
...
+namespace SilverStripe\Versioned\Traits;
+
+use InvalidArgumentException;
+use SilverStripe\Core\Config\Config;
+use SilverStripe\ORM\DataObject;
+use SilverStripe\ORM\DB;
+use SilverStripe\Versioned\ReadingMode;
+use SilverStripe\Versioned\Versioned;
+
+trait VersionNumberCacheTrait
```
ChloeSartorelli, line 18:
```suggestion
/**
 * @internal This trait exists to help encapsulate specific functionality for the Versioned class. Its use with any other class is not supported.
 */
trait VersionNumberCacheTrait
```

src/Traits/VersionsCacheTrait.php
```diff
@@ -0,0 +1,197 @@
...
+
+use SilverStripe\Core\Config\Config;
+use SilverStripe\ORM\DataObject;
+use SilverStripe\ORM\DB;
+use SilverStripe\Model\List\ArrayList;
+use SilverStripe\Versioned\Versioned;
+use SilverStripe\Versioned\Versioned_Version;
+
+/**
+ * Functionality to cache Versioned::Versions(), which is used for the <DataObject>_Versions tables
```
ChloeSartorelli, line 14:
```suggestion
 * Functionality to cache Versioned::Versions(), which is used for the <DataObject>_Versions tables
 * @internal This trait exists to help encapsulate specific functionality for the Versioned class. Its use with any other class is not supported.
```

src/Traits/VersionNumberCacheTrait.php
```diff
@@ -0,0 +1,109 @@
...
+     *
+     * @var array
+     */
+    protected static $cache_versionnumber;
+
+    /**
+     * Used to enable or disable the prepopulation of the version number cache.
+     * Defaults to true.
+     *
+     * @config
```
ChloeSartorelli:
```suggestion
```

src/Traits/VersionNumberCacheTrait.php
```diff
@@ -0,0 +1,109 @@
...
+     * @var boolean
+     */
+    private static $prepopulate_versionnumber_cache = true;
+
+    /**
+     * Shorthand method to call prepopulateVersionNumberCacheForStage() for each applicable stage
```
ChloeSartorelli:
```suggestion
     * Shorthand method to call prepopulateVersionNumberCacheForStage() for all applicable stages
```

src/Traits/VersionNumberCacheTrait.php
```diff
@@ -0,0 +1,109 @@
...
+     * This is cache the Version numbers of a particular DataClass
```
ChloeSartorelli:
```suggestion
     * This caches the version numbers of a particular DataObject class
```
I assume that's what you meant?

src/Traits/VersionNumberCacheTrait.php
```diff
@@ -0,0 +1,109 @@
...
+     * A cache used by get_versionnumber_by_stage().
+     * Clear through {@link flushCache()}.
+     * version (int)0 means not on this stage.
+     *
+     * @var array
+     */
+    protected static $cache_versionnumber;
+
+    /**
+     * Used to enable or disable the prepopulation of the version number cache.
```
ChloeSartorelli, line 30:
```suggestion
     * Used to enable or disable the prepopulation of the version number cache.
     * Set this configuration on the extension, not on the class being extended. For example:
     * <code>
     * SilverStripe\Versioned\Versioned:
     *    prepopulate_versionnumber_cache: false
     * </code>
```

src/Traits/VersionNumberCacheTrait.php
```diff
@@ -0,0 +1,109 @@
...
+    public static function prepopulateVersionNumberCacheForStage(
+        string $dataClass,
+        string $stage,
+        ?array $idList = null,
+    ): void {
+        if (!is_a($dataClass, DataObject::class, true)) {
+            throw new InvalidArgumentException("Class $dataClass is not a DataObject");
+        }
+        ReadingMode::validateStage($stage);
+        if (!Config::inst()->get(self::class, 'prepopulate_versionnumber_cache')) {
```
ChloeSartorelli:
```suggestion
        if (!Config::inst()->get(static::class, 'prepopulate_versionnumber_cache')) {
```

src/Traits/VersionNumberCacheTrait.php
```diff
@@ -0,0 +1,109 @@
...
+     * If $idList is null, then every record will be cached.
+     */
+    public static function prepopulateVersionNumberCacheForStage(
+        string $dataClass,
+        string $stage,
+        ?array $idList = null,
+    ): void {
+        if (!is_a($dataClass, DataObject::class, true)) {
+            throw new InvalidArgumentException("Class $dataClass is not a DataObject");
+        }
```
ChloeSartorelli: I don't think we need this check - we don't do this check for anything else in Versioned that I was able to find. 

There was a weird `$class instanceof DataObject ? $class->ClassName : $class;` at the bottom of the original `prepopulate_versionnumber_cache()` - but the method would have already failed before then.

I think we can skip checks like this. Versioned doesn't work with something that isn't `DataObject`, and failures that will happen naturally should point developers to any mistakes they made passing invalid values in.
```suggestion
```

ChloeSartorelli: The alternative is to add this check in everywhere in `Versioned` that accepts a classname as an argument, which seems unnecessary to me.

src/Versioned.php
```diff
@@ -2460,60 +2396,21 @@ protected function onPrepopulateTreeDataCache($recordList = null, array $options
     {
         $idList = is_array($recordList) ? $recordList :
             ($recordList instanceof DataList ? $recordList->column('ID') : null);
-        Versioned::prepopulate_versionnumber_cache($this->owner->baseClass(), Versioned::DRAFT, $idList);
-        Versioned::prepopulate_versionnumber_cache($this->owner->baseClass(), Versioned::LIVE, $idList);
+        Versioned::prepopulateVersionNumberCache($this->owner->baseClass(), $idList);
```
ChloeSartorelli: This will break the edge case where someone has overridden `prepopulate_versionnumber_cache()` in a subclass with custom functionality.

The only ways to not break that edge case are either to leave these calls as they were, or to have the new `prepopulateVersionNumberCache()` call `prepopulate_versionnumber_cache()` (and let it alias over to `prepopulateVersionNumberCache()` when not overridden).

src/Traits/VersionNumberCacheTrait.php
```diff
@@ -0,0 +1,109 @@
...
+            // Validate the ID list
+            foreach ($idList as $id) {
+                if (!is_numeric($id)) {
+                    $selfClass = self::class;
+                    throw new InvalidArgumentException(
+                        "Bad ID passed to $selfClass::prepopulateVersionNumberCacheForStage() in \$idList: " . $id
+                    );
+                }
+            }
+            $filter = 'WHERE "ID" IN (' . DB::placeholders($idList) . ')';
```
ChloeSartorelli: Using placeholders here is bad for performance. Since we're doing performance updates it might make sense to do a similar enhancement here as https://github.com/silverstripe/silverstripe-framework/issues/10887

Not suggesting that should be part of this PR specifically, since this is copy/pasted code - but if you agree that sounds like a good idea, I'll raise a separate card for it.

src/Traits/VersionsCacheTrait.php
```diff
@@ -0,0 +1,197 @@
...
+trait VersionsCacheTrait
+{
+    /**
+     * In-memory cache of Versions i.e. the <DataObject>_Versions table
+     * @internal
+     */
+    private static array $versions_cache = [];
+
+    /**
+     * Used whether to use the Versions cache
```
ChloeSartorelli:
```suggestion
     * Determines whether to use cached data in the Versions() method
```
I assume that's the only method that uses it? Also specify this in the getter and setter

src/Traits/VersionsCacheTrait.php
```diff
@@ -0,0 +1,197 @@
...
+    ): void {
+        self::class::readVersions($baseDataClass, $ids, null, $filter, $sort, $limit, $join);
+    }
+
+    /**
+     * Shared internal logic which allows passing either a single ID for Versioned::Versions()
+     * or passing an array of IDs for VersionsCachingTrait::prepopulateVersionsCache()
+     *
+     * @param int|array $idOrIds A single record ID, or an array of record IDs
+     * @return ArrayList<Versioned_Version>|array<ArrayList<Versioned_Version>> an ArrayList when $idOrIds is an int, an array when $idOrIds is an array
```
ChloeSartorelli:
```suggestion
     * @return ArrayList<Versioned_Version>|array<int, ArrayList<Versioned_Version>> an ArrayList when $idOrIds is an int, an array when $idOrIds is an array
```
The IDs as a key in the array is important part of the return value

src/Traits/VersionsCacheTrait.php
```diff
@@ -0,0 +1,197 @@
...
+    private static function readVersions(
+        string $baseDataClass,
+        int|array $idOrIds,
+        ?DataObject $owner = null,
+        mixed $filter = '',
+        mixed $sort = '',
+        mixed $limit = '',
+        mixed $join = '',
+    ): ArrayList|array {
+        // Do not attempt to cache anything where $filter etc are not strings
```
ChloeSartorelli: > Do not attempt to cache anything where $filter etc are not strings

Why? That means e.g. `filter: ['MyField' => 'Some value']` will result in not caching anything.

src/Traits/VersionsCacheTrait.php
```diff
@@ -0,0 +1,197 @@
...
+        $useCache = self::class::getUseVersionsCache() && is_string($filter) && is_string($sort)
```
ChloeSartorelli: If `getUseVersionsCache()` returns false, `prepopulateVersionsCache()` will spend a bunch of time doing queries but not do anything with the values, which is explicitly _bad_ for performance. Surely this check should only be in `Versions()`?

I.e. when `prepopulateVersionsCache()` is called we _always_ want to populate the cache, but when `Versions()` is called we may-or-may-not want to read from that cache.

src/Traits/VersionsCacheTrait.php
```diff
@@ -0,0 +1,197 @@
...
+
+        if (is_int($idOrIds)) {
+            $query->addWhere([
+                "\"{$baseTable}_Versions\".\"RecordID\" = ?" => $idOrIds
+            ]);
+        } else {
+            $in = DB::placeholders($idOrIds);
+            $query->addWhere([
+                "\"{$baseTable}_Versions\".\"RecordID\" IN ($in)" => $idOrIds
+            ]);
```
ChloeSartorelli: As with the placeholders in the other trait, using placeholders here is bad for performance. Would you like to take care of respecting `DataList.use_placeholders_for_integer_ids` here or in the same issue that will do it for the other trait?

src/Versioned.php
```diff
@@ -2023,62 +2009,12 @@ public function stagesDifferRecursive(): bool
...
-        $versions = new ArrayList();
-
-        foreach ($records as $record) {
-            $versions->push(new Versioned_Version($record));
-        }
-
-        Versioned::set_reading_mode($oldMode);
-        return $versions;
+        $id = $this->getOwner()->ID;
+        $owner = $this->getOwner();
```
ChloeSartorelli:
```suggestion
        $owner = $this->getOwner();
        $id = $owner->ID;
```
Tiny subjective change but it's nicer to not call that method a second time if we're making a variable of it anyway.

src/Traits/VersionsCacheTrait.php
```diff
@@ -0,0 +1,197 @@
...
+                        $arrayReturn[$id] = self::class::$versions_cache[$key];
+                    }
+                }
+                // From this point only fetch uncached IDs
+                $idOrIds = array_diff($idOrIds, array_keys($arrayReturn));
+            }
+        }
+        if (is_int($idOrIds)) {
+            // When an object is not yet in the Database, we can't get its versions
+            if (!$owner || !$owner->isInDB()) {
```
ChloeSartorelli: We're passing in the ID and the class name - why do we _also_ have to pass in the record? It seems to me we should either:

1. pass in just the record, and grab the classname and ID from it, or
1. pass in just the classname and ID, and use `get_one` to get a potentially-cached record.

src/Traits/VersionsCacheTrait.php
```diff
@@ -0,0 +1,197 @@
...
+    /**
+     * Shared internal logic which allows passing either a single ID for Versioned::Versions()
+     * or passing an array of IDs for VersionsCachingTrait::prepopulateVersionsCache()
+     *
+     * @param int|array $idOrIds A single record ID, or an array of record IDs
+     * @return ArrayList<Versioned_Version>|array<ArrayList<Versioned_Version>> an ArrayList when $idOrIds is an int, an array when $idOrIds is an array
+     */
+    private static function readVersions(
+        string $baseDataClass,
+        int|array $idOrIds,
```
ChloeSartorelli: There's quite a lot of "if it's an array, do things differently than if it's an int" that makes me feel like this would be a lot tidier to just split this out into multiple methods.
Have one that takes an int (or an object), and a separate one that takes the array of ints 

In fact frankly have one that's for getting data for a single record which may-or-may-not use the cache, and a separate one that's for caching a bunch of records and doesn't need to return anything.

Then encapsulate the shared logic into several smaller private methods as needed.

I stopped part way through reviewing this method because I think you've inadvertantly made it more complicated than needed by trying to make it DRY.

src/Versioned.php
```diff
@@ -40,6 +41,9 @@
  */
 class Versioned extends Extension implements TemplateGlobalProvider, Resettable
 {
+    use VersionsCacheTrait;
+    use VersionNumberCacheTrait;
```
ChloeSartorelli, line 45: I would have suggested to use an injectable service instead for improved customisation - but the fact that one of the traits is just moving logic from `Versioned` into a trait for ease of maintenance/readability kinda kills that idea. Something for a future major release perhaps.

GuySartorelli (changes requested):

src/Extensions/DataListExtension.php
emteknetnz: There's nothing stopping someone putting this extension on a non-DataList

src/Traits/VersionNumberCacheTrait.php
emteknetnz: Updated

_config/versionedextension.yml
emteknetnz: Updated

src/Extensions/DataListExtension.php
emteknetnz: Updated

src/Traits/VersionNumberCacheTrait.php
emteknetnz: Updated

src/Traits/VersionNumberCacheTrait.php
emteknetnz: Updated

src/Traits/VersionNumberCacheTrait.php
emteknetnz: Updated

src/Traits/VersionNumberCacheTrait.php
emteknetnz: Updated

src/Traits/VersionNumberCacheTrait.php
emteknetnz: Updated

src/Traits/VersionNumberCacheTrait.php
emteknetnz: Removed

src/Traits/VersionNumberCacheTrait.php
emteknetnz: Updated

src/Traits/VersionNumberCacheTrait.php
emteknetnz: Yup safe to remove placeholders, we're doing an is_int() check above (note was is_numeric(), though have changed since we're assuming in many places in code that ID's are always int

src/Traits/VersionsCacheTrait.php
emteknetnz: Updated

src/Traits/VersionsCacheTrait.php
emteknetnz: Updated

src/Traits/VersionsCacheTrait.php
emteknetnz: The Versions cache is off by default and needs to be explicitly turned on.  It's done in this conservative way to reduce the probability of regressions.

Currently this is only done in [AssetAdmin](https://github.com/silverstripe/silverstripe-asset-admin/pull/1560) and [ArchiveAdmin](https://github.com/silverstripe/silverstripe-versioned-admin/pull/395)

Code path is:
- AssetAdminOpen::getFileData() / ArchiveAdmin::createArchiveGridField()
- DataList::prepopulateCaches()
- DataListExtension::onPrepopulateCaches()
- VersionsCacheTrait::setUseVersionsCache(true)

src/Traits/VersionsCacheTrait.php
emteknetnz: Updated

src/Versioned.php
emteknetnz: Updated

src/Versioned.php
emteknetnz: Will leave calls as they were

src/Traits/VersionsCacheTrait.php
emteknetnz: Have redone method

src/Traits/VersionsCacheTrait.php
emteknetnz: Have redone method

src/Traits/VersionsCacheTrait.php
emteknetnz: Yup fair, have redone method

src/Traits/VersionNumberCacheTrait.php
emteknetnz: Have updated to supporting both with and without placeholders

src/Traits/VersionsCacheTrait.php
emteknetnz: Have updated to support both with and without placeholders

src/Traits/VersionsCacheTrait.php
emteknetnz: Have redone method

_config/versionedcache.yml
```diff
@@ -9,3 +9,11 @@ SilverStripe\Core\Injector\Injector:
     constructor:
       args:
         container: 'SilverStripe\Versioned\Caching\VersionedCacheAdapter'
+---
+Name: versionedcacheextension
+After:
+  - '#corecache'
+---
```
ChloeSartorelli: Don't think needs its own name - but it's also not doing any harm so I'll let you decide whether to keep or remove.

```suggestion
```

src/Extensions/DataListExtension.php
```diff
@@ -0,0 +1,27 @@
+<?php
+
+namespace SilverStripe\Versioned\Extensions;
+
+use RuntimeException;
+use SilverStripe\Core\Extension;
+use SilverStripe\ORM\DataObject;
+use SilverStripe\Versioned\Versioned;
+
+class DataListExtension extends Extension
```
ChloeSartorelli, line 12:
```suggestion
/**
 * @extends Extension<DataList>
 */
class DataListExtension extends Extension
```
And add the associated `use` statement for `DataList` so IDEs know what class that is.

src/Extensions/DataListExtension.php
ChloeSartorelli: We don't add type checks to our extension classes assuming someone will slap our `VerySpecificExtension` on a `SomeRandom` class. It's name makes it clear what it's for. I've also added a suggested change to add the appropriate generic typing in the PHPDoc which further clarifies that this is specifically for `DataList` which is specifically meant to hold `DataObject` classes.

src/Traits/VersionNumberCacheTrait.php
ChloeSartorelli: Still need to change "each" to "all"

src/Traits/VersionsCacheTrait.php
```diff
@@ -0,0 +1,187 @@
...
+     */
+    private static function readVersionsByID(
+        DataObject $owner,
+        string $baseDataClass,
+        int $id,
+        mixed $filter,
+        mixed $sort,
+        mixed $limit,
+        mixed $join,
+    ): ArrayList {
```
ChloeSartorelli:
```suggestion
     * @return SSList<Versioned_Version>
     */
    private static function readVersionsByID(
        DataObject $owner,
        string $baseDataClass,
        int $id,
        mixed $filter,
        mixed $sort,
        mixed $limit,
        mixed $join,
    ): SSList {
```


Sorry, should have caught this last review. Lets just say it'll be any list, which lets us use a `DataList` if we find a good way to do so in the future.
I know we could update that at the same time since it's private, but may as well do now when we add it in the first place (and if we promote visibility later we won't have to worry about it).

src/Traits/VersionsCacheTrait.php
```diff
@@ -0,0 +1,187 @@
...
+    /**
+     * Get an ArrayList of Versions for an ID
+     * Will use a cache if setUseVersionsCache(true) was previously called
+     *
+     * @return ArrayList<Versioned_Version>
+     */
+    private static function readVersionsByID(
+        DataObject $owner,
+        string $baseDataClass,
+        int $id,
```
ChloeSartorelli: We're passing in the ID and the class name - why do we _also_ have to pass in the record? It seems to me we should either:

1. pass in just the record, and grab the classname and ID from it, or
1. pass in just the classname and ID, and use `get_one` to get a potentially-cached record.

This was pointed out in the previous incarnation in https://github.com/silverstripe/silverstripe-versioned/pull/500#discussion_r2083713381

GuySartorelli (changes requested):
Only a partial review because the priority right now are the `6.0` PRs which are time senstive.

_config/versionedcache.yml
emteknetnz: Removed

src/Extensions/DataListExtension.php
emteknetnz: Updated

src/Extensions/DataListExtension.php
emteknetnz: Removed

src/Traits/VersionsCacheTrait.php
emteknetnz: Removed $baseDataClass and $id params.  Renamed method to readVersions()

src/Traits/VersionsCacheTrait.php
emteknetnz: I really don't agree with this, it's a private method so it doesn't matter, and we'll probably never actually update to support more things.

More importantly, it's used directly by Versioned::Versions(), which has a docblocked return type of ArrayList so we can't actually change it

src/Extensions/DataListExtension.php
```diff
@@ -0,0 +1,28 @@
...
+use SilverStripe\ORM\DataObject;
+use SilverStripe\ORM\DataList;
+use SilverStripe\Versioned\Versioned;
+
+/**
+ * @extends Extension<DataList>
+ */
+class DataListExtension extends Extension
+{
+    public function onPrepopulateCaches(array $ids): void
```
ChloeSartorelli:
```suggestion
    protected function onPrepopulateCaches(array $ids): void
```
Sorry, I should have noticed this the first time around.

src/Traits/VersionNumberCacheTrait.php
ChloeSartorelli: This still needs to be done

src/Traits/VersionsCacheTrait.php
ChloeSartorelli: > The Versions cache is off by default and needs to be explicitly turned on.

But it's completely new, right? And the only way it could ever be used is if someone calls `prepopulateVersionsCache()` which
1. didn't exist until now, so it's not called anywhere else
2. we explicitly want cache _on_ in the place we're calling this.

What I'm getting at is that I can't see what this boolean toggle is doing for us, given you have to call this new method in order to activate the caching anyway.

So whether we have this boolean toggle or not, the risk or regression with this change is the same - cache will be used in the place where we have called this new method. If we removed the boolean toggle, wouldn't the behaviour be exactly the same but just without needing this extra API?

src/Traits/VersionNumberCacheTrait.php
```diff
@@ -0,0 +1,121 @@
...
+    public static function prepopulateVersionNumberCacheForStage(
+        string $dataClass,
+        string $stage,
+        ?array $ids = null,
+    ): void {
+        $ids ??= [];
+        $notInts = array_filter($ids, fn($id) => !is_int($id));
+        if (count($notInts)) {
+            throw new InvalidArgumentException('The following IDs are non-int: ' . implode(', ', $notInts));
+        }
```
ChloeSartorelli: I don't think we should throw this exception - I don't think we do this anywhere else? Instead, probably the condition below should be `if ($usePlaceholders || /*there is at least one non-int*/)`

src/Traits/VersionNumberCacheTrait.php
```diff
@@ -0,0 +1,121 @@
...
+     * a list of record IDs, for more efficient database querying.
+     * If the $ids param is null, then every record will be cached.
+     */
+    public static function prepopulateVersionNumberCacheForStage(
+        string $dataClass,
+        string $stage,
+        ?array $ids = null,
+    ): void {
+        $ids ??= [];
+        $notInts = array_filter($ids, fn($id) => !is_int($id));
```
ChloeSartorelli: Should we use `ctype_digit` like in `RelationList::prepareForeignIDsForWhereInClause()` and `ExactMatchFilter::usePlaceholders()`?

src/Traits/VersionNumberCacheTrait.php
```diff
@@ -0,0 +1,121 @@
...
+            // If we are caching IDs for _all_ records then we can mark this cache as "complete" and in the case of a cache-miss
+            // no subsequent call is necessary
+            self::class::$cache_versionnumber[$baseClass][$stage] = [ '_complete' => true ];
+        }
+        $sql = "SELECT \"ID\", \"Version\" FROM \"$stageTable\" $filter";
+        if ($usePlaceholders) {
+            $versions = DB::prepared_query($sql, $parameters)->map();
+        } else {
+            $versions = DB::query($sql)->map();
+        }
```
ChloeSartorelli: Why are we using super raw SQL instead of the `SQLSelect` class for example?

src/Traits/VersionsCacheTrait.php
```diff
@@ -0,0 +1,187 @@
...
+        $query = $list->dataQuery()->query();
+        $baseTable = null;
+        foreach ($query->getFrom() as $table => $tableJoin) {
+            if (is_string($tableJoin) && $tableJoin[0] == '"') {
+                $baseTable = str_replace('"', '', $tableJoin ?? '');
+            } elseif (is_string($tableJoin) && substr($tableJoin ?? '', 0, 5) != 'INNER') {
+                $query->setFrom([
+                    $table => "LEFT JOIN \"$table\" ON \"$table\".\"RecordID\"=\"{$baseTable}_Versions\".\"RecordID\""
+                        . " AND \"$table\".\"Version\" = \"{$baseTable}_Versions\".\"Version\""
+                ]);
```
ChloeSartorelli, line 105: I'm assuming you already looked to see if there's an abstraction on `Versioned` to join the table like this - but why not use the left join method?

src/Traits/VersionNumberCacheTrait.php
ChloeSartorelli: Same thing in `innerReadVersions()` on the other trait

src/Traits/VersionNumberCacheTrait.php
ChloeSartorelli: Same thing in `innerReadVersions()` on the other trait

src/Traits/VersionsCacheTrait.php
```diff
@@ -0,0 +1,187 @@
...
+        $id = $owner->ID;
+        $stringable = fn(mixed $var) => $var === null || is_scalar($var) || $var instanceof Stringable;
+        $allStringable = count(array_filter([$filter, $sort, $limit, $join], $stringable)) === 4;
+        $useCache = self::class::getUseVersionsCache() && $allStringable;
+        if ($useCache) {
+            $key = self::class::generateVersionsCacheKey($id, $filter, $sort, $limit, $join);
+            if (array_key_exists($key, self::class::$versions_cache)) {
+                return self::class::$versions_cache[$key];
+            }
+        }
```
ChloeSartorelli: The idea of only caching if filter/etc are stringable or scalar doesn't feel right to me. Almost always when I use filter it's an array - I suspect the same is true of most usages of filter. Should we use serialise instead/as well?

GuySartorelli (changes requested):
Please see https://github.com/silverstripe/silverstripe-versioned-admin/pull/395#pullrequestreview-2855757249 before proceeding.

src/Traits/VersionsCacheTrait.php
emteknetnz: It's code copied over from Versioned::Versions(), I don't want to change it

src/Extensions/DataListExtension.php
emteknetnz: Updated

src/Traits/VersionNumberCacheTrait.php
emteknetnz: Updated for both

src/Traits/VersionNumberCacheTrait.php
emteknetnz: It protects against SQL injection when we're not using placeholders and doing a raw query. There's also no reason to pass in non ints, and I like hard exceptions, so this should stay. I've added inline comments above both to say it is protection against sql injection

src/Traits/VersionNumberCacheTrait.php
emteknetnz: It's code copied over from Versioned::prepopulate_versionnumber_cache(), I don't want to change it

src/Traits/VersionNumberCacheTrait.php
ChloeSartorelli: > It protects against SQL injection when we're not using placeholders and doing a raw query.

That's why I suggest using `if ($usePlaceholders || /*there is at least one non-int*/)` which is consistent with [`RelationList::prepareForeignIDsForWhereInClause()`](https://github.com/silverstripe/silverstripe-framework/blob/2af85b7859beb0afb18374e1fd4831d1bb64fb7e/src/ORM/RelationList.php#L173-L187).

You may remember when we implemented that, we considered that some projects may choose to use non-integer IDs and that we shouldn't get in the way of that. See https://github.com/silverstripe/silverstripe-framework/issues/10860 and in particular https://github.com/silverstripe/silverstripe-framework/issues/10860#issuecomment-1622853554

src/Traits/VersionsCacheTrait.php
emteknetnz: Ah yup, have updated

src/Traits/VersionNumberCacheTrait.php
ChloeSartorelli: You did touch it, it's changed since then - but I take your point that this style of SQL query was already used here.

GuySartorelli (changes requested):
Still outstanding:

- https://github.com/silverstripe/silverstripe-versioned/pull/500#discussion_r2096559370
- https://github.com/silverstripe/silverstripe-versioned/pull/500#discussion_r2083687261
- https://github.com/silverstripe/silverstripe-versioned/pull/500#discussion_r2099320267

Didn't do any additional review this round as there is still stuff left untouched from last time.

src/Traits/VersionNumberCacheTrait.php
emteknetnz: Updated

src/Traits/VersionsCacheTrait.php
emteknetnz: Any time anyone calls Versioned::Versions() it'll potentially use the cache

The bool is so that the cache is only used for scenarios where we've tested it

src/Traits/VersionNumberCacheTrait.php
emteknetnz: Done

src/Traits/VersionsCacheTrait.php
ChloeSartorelli: But it will only use that cache if the cache has been warmed.
Which only happens if we call `prepopulateVersionsCache()`.
Which in the only places we do this, we also toggle that boolean to true. So the boolean isn't doing anything.

GuySartorelli:
The conversation here is ongoing: https://github.com/silverstripe/silverstripe-versioned/pull/500#discussion_r2101602485

It unfortunately won't let me comment on it as a review so just keep checking back there when this card gets assigned to you

src/Traits/VersionsCacheTrait.php
emteknetnz: Not entirely true, if $useCache is true (or if we just removed the bool) then as data is fetched it will populate the cache with the returned results

ChloeSartorelli: I don't think an external setter method should determine that, which I think might be at the heart of the mismatch of our understanding here.

When we call `prepopulateVersionsCache()` we pass in some specific IDs and a class name. It should prepopulate the cache for those records (regardless of some external boolean) - and _only_ those records. There should be no risk of any other records being cached because anything else wasn't passed into that method.

emteknetnz: Hmm, I'll need to look into this a little more

`VersionedCacheTrait::prepopulateVersionsCache()` will only populate the cache for `$ids`, it will effectively pass blank strings  for `$filter, $sort, $limit, $join` which `Versioned::Versions()` can pass it to `VersionedCacheTrait::readVersions()`

I'm not sure if caching based on `$filter, $sort, $limit, $join` is relevant for the specific DB queries I was trying to reduce

If not, then we can remove those from the cache key, presumably remove the "incremental" caching the can be done by calling `readVersions()` i.e. `self::class::$versions_cache[$key] = $list;`, and remove the bool + getter/setter

emteknetnz: Cool looks like the "incremental caching" i.e. the cache entries added in VersionsCacheTrait::readVersions() rather than VersionsCacheTrait::prepopulateVersionsCache() does not reduce the number of queries when filtering and sorting, so we should be able to simplify things as you've recommended

src/Traits/VersionsCacheTrait.php
```diff
@@ -0,0 +1,155 @@
...
+     */
+    private static function generateVersionsCacheKey(
+        int $id,
+        mixed $filter = '',
+        mixed $sort = '',
+        mixed $limit = '',
+        mixed $join = '',
+    ): string|false {
+        $keyInputs = [$filter, $sort, $limit, $join];
+        $isSerializable = fn(mixed $var) => @serialize($var) !== false;
```
ChloeSartorelli, line 148: `@` is used to suppress errors - what errors are we suppressing here, and why? What happens in the rest of the execution when there are errors that we simply suppress here?

GuySartorelli (commented):
Just one question here, plus a note in https://github.com/silverstripe/silverstripe-versioned-admin/pull/395 from the last cycle

src/Traits/VersionsCacheTrait.php
emteknetnz: Yes that's exactly what I want to happen, if $var has something in it like an anonymous function which cannot be serialized, then `@serialize($var)` will return false without triggering a php error that will break execution

We then check if all the `$keyInputs` are serializable below. If one or more are not, then we do not attempt to create a key from them, and therefore do not attempt to cache

emteknetnz:
> plus a note in https://github.com/silverstripe/silverstripe-versioned-admin/pull/395 from the last cycle

I saw the note, though looks like there's nothing to do there? That PR isn't merged and we haven't got a finalised solution so I cannot check to see the diff in queries

GuySartorelli:
> I saw the note, though looks like there's nothing to do there? That PR isn't merged and we haven't got a finalised solution so I cannot check to see the diff in queries

You can check to see the diff in queries by running those PRs. I also think that regardless of which implementation we choose, the result in terms of db queries will be the same - the call to `versions()` will be unnecessary.

Am I correct that the `VersionsCacheTrait` functionality is for archive admin specifically, and not for asset-admin?

Either way, we should resolve the other PRs which will organically reduce the number of queries before merging these in. We may find all of these PRs are still useful, or we may find that the archive admin PR is no longer needed.

emteknetnz:
Have rebased the things and rerun queries

Looks lke VersionCacheTrait no longer has any impact on asset-admin or archive-admin

I tested running flush before each test, and to test "without VersionCacheTrait" I simply put a `return` at the top of `VersionsCacheTrait::prepopulateVersionsCache()`

asset-admin with 9 files
109 with VersionCacheTrait
108 without VersionCacheTrait

archive-admin with 5 records
99 with VersionCacheTrait
98 without VersionCacheTrait

Ha

So we definitely don't need the versioned-admin PR. I guess the question is do we want to keep the VersionCacheTrait in this PR?

GuySartorelli:
If we're not gonna use the trait (which it sounds like we're not) then we probably don't want to keep the code for it sitting here.

Maybe close this PR and spin out a new one with only the trait we'll use - and then this PR can be a reference point if we find a use for caching `versions()` queries in the future?

emteknetnz:
Replaced by https://github.com/silverstripe/silverstripe-versioned/pull/512

PR #506, opened 2025-05-13
TLN Update translations

GuySartorelli:
Automated translations update generated using [silverstripe/tx-translator](https://github.com/silverstripe/silverstripe-tx-translator)

### DIFF
--- a/lang/en.yml
+++ b/lang/en.yml
@@ -8,16 +8,16 @@ en:
       one: '1 total'
       other: '{count} total'
     NAME: Name
-    PLURALNAME: Campaigns
+    PLURALNAME: Changesets
     PLURALS:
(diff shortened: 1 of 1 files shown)


lang/en.yml
```diff
@@ -8,16 +8,16 @@ en:
       one: '1 total'
       other: '{count} total'
     NAME: Name
-    PLURALNAME: Campaigns
+    PLURALNAME: Changesets
```
ChloeSartorelli, line 11: Checked this and the plural below in both CMS 5 and 6 - it seems they don't affect campaign admin so should be safe to make these changes.

GuySartorelli:
This will be regenerated

Issue #505, opened 2025-04-16
Archived Page History Hangs


jalamanderman:
### Module version(s) affected

3.4.0

### Description

When viewing the history tab of an archived page within the CMS (not in the archive admin section), the page hangs indefinitely. 

![Image](https://github.com/user-attachments/assets/bf54fb16-74a4-44df-ae04-d83c1fa93557)

## Reproduction steps

2. Add a page to the site
3. Archive the page.
5. Click on the "History" tab of the archived page.
6. The page then hangs during loading.

## Expected Behavior:

The history of the archived page should load correctly.

## Actual Behavior:

The page hangs indefinitely during loading.

## Environment:

Silverstripe version: 5.4.0
silverstripe/subsites version: 3.4.0
PHP version: 8.3.19
Browser: Brave [Version 1.77.97 Chromium: 135.0.7049.84 (Official Build) (arm64)](https://brave.com/latest/)

## Additional Information:

I have been able to reproduce this issue with a clean Silverstripe installation

### Validations

- [x] Check that there isn't already an issue that reports the same bug
- [x] Double check that your reproduction steps work in a fresh installation of [`silverstripe/installer`](https://github.com/silverstripe/silverstripe-installer) (with any code examples you've provided)

### CMS 5 PRs

[**Installer CI run**](https://github.com/creative-commoners/silverstripe-installer/actions/runs/15036314698)

- https://github.com/silverstripe/silverstripe-versioned-admin/pull/398
- https://github.com/silverstripe/silverstripe-versioned/pull/507
- https://github.com/silverstripe/silverstripe-versioned-admin/pull/400

### CMS 6 PRs
- https://github.com/silverstripe/silverstripe-versioned-admin/pull/399

GuySartorelli:
This doesn't appear to have anything to do with subsites. The same happens on an installation without the subsites module installed.

Migrating to silverstripe/silverstripe-versioned

GuySartorelli:
Worth noting that behaviour is different between CMS5 and 6.

In CMS 5 it's as described above, plus there's an error in the graphql request (args and additional trace removed for brevity):

```text
{"data":{"readOnePage":{"id":"12","versions":null,"__typename":"Page"}},"errors":[{"message":"Cannot view versions on Page","code":0,"file":"\/var\/www\/html\/vendor\/silverstripe\/versioned\/src\/GraphQL\/Resolvers\/VersionedResolver.php","line":90,"trace":"[
    {
[...]

PR #507, opened 2025-05-15
FIX Allow viewing version history for archived records

GuySartorelli:
The `canViewStage()` check always returns false for archived records, because it was trying to fetch the draft version of the page, which doesn't exist because this page is archived.

Note that because this implementation is in a GraphQL resolver, a separate fix is needed for CMS 6.

## Issue

- https://github.com/silverstripe/silverstripe-versioned/issues/505


emteknetnz (approved):
Works well

PR #511, opened 2025-05-22
FIX Include deleted version data where appropriate.

GuySartorelli:
`latest_versions` explciitly excludes versions where `WasDeleted` is true, which is the archived version.
Several methods in `Versioned` explicitly want to _include_ that information.

## Issue

- https://github.com/silverstripe/silverstripe-versioned-admin/issues/402

### DIFF
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -296,6 +296,12 @@ class Versioned extends Extension implements TemplateGlobalProvider, Resettable
     private static $use_session = false;
 
     /**
+     * Temporarily store data about a deleted version
+     * @internal
+     */
+    private static $deletedVersionData = [];
+
+    /**
      * Reset static configuration variables to their default values.
      */
     public static function reset()
@@ -783,7 +789,7 @@ SQL
 
         $query->addWhere("\"{$baseTable}_Draft\".\"ID\" IS NULL");
 
-        $this->augmentSQLVersionedLatest($query, $dataQuery);
+        $this->augmentSQLVersionedLatestIncludingDeleted($query, $dataQuery);
     }
 
     /**
@@ -815,22 +821,35 @@ SQL
      * Return latest version instances, regardless of whether they are on a particular stage.
      * This provides "show all, including deleted" functionality.
      *
-     * Note: latest_version ignores deleted versions, and will select the latest non-deleted
-     * version.
-     *
-     * @param SQLSelect $query
-     * @param DataQuery $dataQuery
+     * Note: latest_versions ignores versions which represent the archive action (aka deleted versions),
+     * and will select the latest non-deleted version.
      */
     protected function augmentSQLVersionedLatest(SQLSelect $query, DataQuery $dataQuery)
     {
+        $this->augmentSQLVersionedLatestInner($query, $dataQuery, true);
+    }
+
+    /**
+     * Return latest version instances, regardless of whether they are on a particular stage
+     * and including versions which represent the archive action (aka deleted versions).
+     */
+    protected function augmentSQLVersionedLatestIncludingDeleted(SQLSelect $query, DataQuery $dataQuery)
+    {
+        $this->augmentSQLVersionedLatestInner($query, $dataQuery, false);
+    }
+
+    private function augmentSQLVersionedLatestInner(SQLSelect $query, DataQuery $dataQuery, bool $excludeDeleted)
+    {
         // Query against _Versions table first
-        $this->augmentSQLVersioned($query);
+        $this->augmentSQLVersioned($query, $excludeDeleted);
 
         // Join and select only latest version
         $baseTable = $this->baseTable();
         $subSelect = $this->prepareMaxVersionSubSelect($query, $dataQuery);
 
-        $subSelect->addWhere("\"{$baseTable}_Versions_Latest\".\"WasDeleted\" = 0");
+        if ($excludeDeleted) {
+            $subSelect->addWhere("\"{$baseTable}_Versions_Latest\".\"WasDeleted\" = 0");
+        }
 
         if ($this->shouldApplySubSelectAsCondition($query)) {
             $subSelect->addWhere(
@@ -1173,7 +1192,11 @@ SQL
         ];
 
         // Add any extra, unchanged fields to the version record.
-        $data = DB::prepared_query("SELECT * FROM \"{$table}\" WHERE \"ID\" = ?", [$recordID])->record();
+        if ($isDelete && isset(Versioned::$deletedVersionData[$table][$recordID])) {
+            $data = Versioned::$deletedVersionData[$table][$recordID];
+        } else {
+            $data = DB::prepared_query("SELECT * FROM \"{$table}\" WHERE \"ID\" = ?", [$recordID])->record();
+        }
         if ($data) {
             $fields = $schema->databaseFields($class, false);
             if (is_array($fields)) {
@@ -1901,10 +1924,29 @@ SQL
         return true;
     }
 
+    protected function onBeforeDelete(): void
+    {
+        // Skip if suppressed by parent delete
+        if (!$this->getDeleteWritesVersion()) {
+            return;
+        }
+        // Pre-fetch data that will be used in augmentWriteVersioned().
+        // This is necessary because that data doesn't exist after the base record is deleted.
+        $owner = $this->getOwner();
+        $baseTable = $owner->baseTable();
+        $recordID = $owner->ID;
+        Versioned::$deletedVersionData[$baseTable][$recordID] = DB::prepared_query("SELECT * FROM \"{$baseTable}\" WHERE \"ID\" = ?", [$recordID])->record();
+    }
+
     protected function onAfterDelete()
     {
         // Create deleted record for current stage
         $this->createDeletedVersion(static::get_stage());
+        // Remove pre-delete data.
+        $owner = $this->getOwner();
+        $baseTable = $owner->baseTable();
+        $recordID = $owner->ID;
+        unset(Versioned::$deletedVersionData[$baseTable][$recordID]);
     }
 
     /**


src/Versioned.php
```diff
@@ -783,7 +783,7 @@ protected function augmentSQLVersionedRemovedFromDraft(SQLSelect $query, DataQue
 
         $query->addWhere("\"{$baseTable}_Draft\".\"ID\" IS NULL");
 
-        $this->augmentSQLVersionedLatest($query, $dataQuery);
+        $this->augmentSQLVersionedLatestIncludingDeleted($query, $dataQuery);
```
ChloeSartorelli, line 792: `augmentSQLVersionedRemovedFromDraft()` and by extension `augmentSQLVersionedArchiveOnly()` are both meant to _explicitly_ include data about the archived (aka deleted) version.

GuySartorelli:
CI is failing because the `Title` column for the deleted version record is `null`. We need to determine which of the following is true:

1. Deleting a draft-only record like the test in `prepareRecordsForVersionTests()` is doing is bad practice - it should be wrapped in `suppressDeletedVersion()` (like `doArchive()` does). That would then allow the tests to pass, because there is no bad-practice "deleted" version history with a null title.
2. `Versioned::createDeletedVersion()` should retain the values for those fields, for example based on the record in memory - this would probably mean changing how `Versioned::augmentWriteVersioned()` works which currently fetches values _from the database_, which is why in this case they're null (the base record got deleted first so there's no data in the database to draw from). Changing this could result in unexpected regressions (e.g. set some value on a record but don't write to db, then delete from stage - it'll grab your unsaved data and put it in the versions table)
3. The current behaviour without this PR is all correct except for sort order in the archive admin specifically - means I'll need to find a different approach to fix the bug.
4. Something I haven't considered yet?

I am going to go with 1 above for now, get CI green, and then let the reviewer make a decision based on the final changes.

My second choice would be option 2 - though there's some obvious risks inherent in that option of ending up with incorrect data in the edge case of a draft version being deleted improperly (which is the very edge case this is all about). I think incorrect data is arguably worse than missing data, since missing data raises obvious red flags while incorrect data isn't immediately obvious.

src/Versioned.php
```diff
@@ -1441,7 +1454,7 @@ public function setDeleteWritesVersion($flag)
      * @param callable $callback
      * @return mixed Result of $callback()
      */
-    protected function suppressDeletedVersion($callback)
+    public function suppressDeletedVersion($callback)
```
ChloeSartorelli: I use this in the test below - it also seems sensible to be public in general since it's wrapping around the public `getDeleteWritesVersion()` and `setDeleteWritesVersion()` methods.

tests/php/VersionedTest.php
```diff
@@ -2097,11 +2097,11 @@ private function prepareRecordsForVersionTests(): void
...
         $noDraftRecord1->write();
         $noDraftRecord1->publishSingle();
-        $noDraftRecord1->deleteFromStage(Versioned::DRAFT);
+        $noDraftRecord1->suppressDeletedVersion(fn () => $noDraftRecord1->deleteFromStage(Versioned::DRAFT));
         $noDraftRecord2 = new VersionedTest\NoFixtureModel();
         $noDraftRecord2->Title = 'deleted on draft2';
         $noDraftRecord2->write();
         $noDraftRecord2->publishSingle();
-        $noDraftRecord2->deleteFromStage(Versioned::DRAFT);
+        $noDraftRecord2->suppressDeletedVersion(fn () => $noDraftRecord2->deleteFromStage(Versioned::DRAFT));
```
ChloeSartorelli: This change is based on option 1 from https://github.com/silverstripe/silverstripe-versioned/pull/511#issuecomment-2902905039

src/Versioned.php
```diff
@@ -815,22 +815,35 @@ protected function augmentSQLVersionedLatestSingle(SQLSelect $query, DataQuery $
...
     protected function augmentSQLVersionedLatest(SQLSelect $query, DataQuery $dataQuery)
+    {
+        $this->augmentSQLVersionedLatestInner($query, $dataQuery, true);
+    }
+
+    /**
+     * Return latest version instances, regardless of whether they are on a particular stage
+     * and including versions which represent the archive action (aka deleted versions).
+     */
+    protected function augmentSQLVersionedLatestIncludingDeleted(SQLSelect $query, DataQuery $dataQuery)
```
ChloeSartorelli, line 836: I've chosen not to add a versioned mode (see `augmentSQL()`) for this method, since we currently don't need one, and adding those isn't just a matter of updating `augmentSQL()` but also other places as well.

src/Versioned.php
emteknetnz: Can't do this as this is an API break, any subclass implementing this method will complain about the access level changing

e.g. try running the below with `php -a` and changing `CA::PF()` to public

```php
class CA {
  public function F() { return $this->PF(); }
  protected function PF() { return 'PFA'; }
}
class CB extends CA {
  protected function PF() { return 'PFB'; }
}
echo (new CA)->F() . "\n";
echo (new CB)->F() . "\n";;
```

emteknetnz (changes requested):
I'm not sure if option 1) is correct:
a) There's an API break required (though we could work around this by adding a new public method that calls the existing protected method, and deprecate the protected method)
b) It says that it matches what doArchive() does ... though doesn't _something_ still create the records in the _Versions table on delete/archive? i.e. there are will be some WasDeleted=1 WasArchived=1 records created on archive in the _Versions table - is suppressDeletedVersion() simply preventing a duplicate record from being created?

That said, I'm not sure which option is best

GuySartorelli:
The API break isn't a deal breaker, there's plenty of ways to work around that. Which option is correct doesn't depend on my specific implementation of a solution - rather, the implementation depends on which option is factually correct. If we think option 1 is the factually correct option, then we should work to implement a solution based on those facts.

> It says that it matches what doArchive() does

Sorry, I'm missing the context here. What matches what `doArchive()` does?

> though doesn't something still create the records in the _Versions table on delete/archive? i.e. there are will be some WasDeleted=1 WasArchived=1 records created on archive in the _Versions table - is suppressDeletedVersion() simply preventing a duplicate record from being created?

I'm not sure what you're asking here, sorry.

emteknetnz:
> Sorry, I'm missing the context here. What matches what doArchive() does?

Calling suppressDeletedVersion() to match doArchive() was mentioned here https://github.com/silverstripe/silverstripe-versioned/pull/511#issuecomment-2902905039

`Deleting a draft-only record like the test in prepareRecordsForVersionTests() is doing is bad practice - it should be wrapped in suppressDeletedVersion() (like doArchive() does). That would then allow the tests to pass, because there is no bad-practice "deleted" version history with a null title.`

> I'm not sure what you're asking here, sorry.

I'm assuming (possibly incorrectly) that `suppressDeletedVersion()` is called in `doArchive()` to prevent a duplicate WasDeleted=1 record on the _Versions from being created, rather than to entirely prevent a WasDeleted=1 record on the _Versions table from being created

In these PRs, I'm not sure if `suppressDeletedVersion()` is being used to incorrectly prevent a WasDeleted=1 record from being created on the _Versions table

GuySartorelli:
> I'm assuming (possibly incorrectly) that `suppressDeletedVersion()` is called in `doArchive()` to prevent a duplicate WasDeleted=1 record on the _Versions from being created, rather than to entirely prevent a WasDeleted=1 record on the _Versions table from being created
>
> In these PRs, I'm not sure if `suppressDeletedVersion()` is being used to incorrectly prevent a WasDeleted=1 record from being created on the _Versions table

With the `suppressDeletedVersion()` there, calling `doArchive()` will create a single new version row:

|ID|RecordID|Version|WasPublished|WasDeleted|WasDraft|AuthorID|PublisherID|ClassName|LastEdited|Created|CanViewType|CanEditType|Title|URLSegment|
|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|
|29|11|4|1|1|1|0|0|Page|2025-05-26 12:45:19|2025-05-26 12:42:46|Inherit|Inherit|Archiveme1|archiveme1|

This row is _manually created_ when `createDeletedVersion()` is called.

Removing the call to  `suppressDeletedVersion()` will result in three new version rows being added:

|ID|RecordID|Version|WasPublished|WasDeleted|WasDraft|AuthorID|PublisherID|ClassName|LastEdited|Created|CanViewType|CanEditType|Title|URLSegment|
|---|---|---|---|---|---|---|---|---|---|---|---|---|---|---|
|30|12|4|1|1|0|0|0|Page|2025-05-26 12:45:49|2025-05-26 12:43:42|Inherit|Inherit|Archiveme2|archiveme2|
|31|12|5|1|1|1|0|0|Page|2025-05-26 12:45:49|2025-05-26 12:43:42|Inherit|Inherit|Archiveme2|archiveme2|
|32|12|6|0|1|1|0|0|Page|2025-05-26 12:45:49|2025-05-26 12:43:42|Inherit|Inherit|NULL|NULL|

The first row is created as a result of the call chain `doArchive()` -> `doUnpublish()` -> `delete()` -> `onAfterDelete()` -> `createDeletedVersion()`. This is the deletion of the `_LIVE` table row. It gets its Title, URLSegment, and any other model-specific data from the data in the draft (base) table. This same row will also be correctly created if normally calling `doUnpublish()`.

The second row is the same as the manual row created in the first example.

The third row is created as a result of the call chain `doArchive()` -> `deleteFromStage()` -> `delete()` -> `onAfterDelete()` -> `createDeletedVersion()`. This is the deletion of the draft (base) table row. Because this happens after that row is deleted, there's no model-specific data to collect, so the model-specific columns are `null`. This is the same as the deletions that this and the other PRs are adding suppression around.

Deleting draft records without suppression should be considered bad practice if we accept the first option, with the reasoning being that it's creating a `WasDeleted=1` row without any of the model-specific data which we normally want to keep because it tells us about the state of the record at the time it was "deleted". e.g. reverting to a row like this will result in removing all data as though you just created a new record.

GuySartorelli:
tl;dr the calls to `suppressDeletedVersion()` in `doArchive()` serve two separate purposes:

1. The first suppression stops a nearly-identical (though slightly different) duplicate WasDeleted=1 version row being created when removing the row from the live table.
2. The second suppression stops an _incorrect_ (if we accept option 1) WasDeleted=1 row from being created when removing the row from the draft (base) table. This row would be missing all of the model-specific data e.g. `Title` and `URLSegment`, and reverting to this version (if it was allowed to be created) would result in data loss.

emteknetnz:
Seems like the issue essentially being worked around is that `Versioned::deleteFromStage(Versioned::DRAFT)` is creating the _Versions record `onAfterDelete()`, where is should be created `onBeforeDelete()`?

Would implementing that be viable? That should prevent the need to wrap in `suppressDeletedVersion()`?

GuySartorelli:
That would be fine in most scenarios. The only situations where I can see that being problematic are:

1. if someone cancels a delete by implementing `onBeforeDelete()` in their own class and throwing an exception, and catching it upstream - which is unlikely at best and bad practice at worst.
2. if something goes wrong before the actual deletion happens, we'd end up with a `WasDeleted=1` row in the versions table, without having actually removed the row in the base table.
3. Someone is relying on not having a `WasDeleted=1` row until after delete, somehow, for some reason.

The first above I think we can safely ignore.
The second doesn't sound good to me, but I _think_ everywhere that would be looking for deleted versions also checks for the presence of the record in the draft table, so actual chance of regression I _think_ is low.
The third is unlikely but is the most breaking if someone _is_ doing that. I can't think of any actual use case for that though.

tl;dr doing that is feasible and is _unlikely_ to cause regressions. Chance of regressions is probably about the same as the current solution so I'm happy to go either way, let me know which you prefer (or if you want to explore another avenue).

emteknetnz:
Hmm yeah I guess there is some risk there, I guess it would still be better if it was done after delete, just the issue is the regular data record is no longer there. That way we can be confident there aren't phantom _Versions records being created

Is cloning the record or dataobject into a private property somewhere right before the regular data record is deleted and then using that to populate _Versions record created after delete viable?

GuySartorelli:
> Is cloning the record or dataobject into a private property somewhere right before the regular data record is deleted and then using that to populate _Versions record created after delete viable?

We could maybe use the data in `$this->original` (we'd need to add a public method to surface that) and use that in the insert manipulation query instead of inserting from the data in the database. That feels _more_ risky to me than the other options though, because:

1. I don't know if that data _exactly_ matches what's in the database (in fact it probably doesn't, since in some cases it hasn't been pushed through the relevant `DBField` instance yet)
2. the way the query is being written is using a very obscure "manipulation" API I'm not familiar with, which automatically jacks up the chance of regressions

We could instead do a `SELECT` before inserting the delete row into the versions table, but I don't want to add extra queries (and the point about the "manipulation" API is still a problem for this option too)

emteknetnz:
> We could instead do a SELECT before inserting the delete row into the versions table, but I don't want to add extra queries (and the point about the "manipulation" API is still a problem for this option too)

Are you sure it would be adding extra queries? I would have thought that doing this onBeforeDelete or simply what there right now where it's fetching null values from the database is the same number of queries?

GuySartorelli:
Yes, adding an extra `SELECT` query would be adding extra queries. And I don't want to do it anyway for the regression risk.

emteknetnz:
Had a chat offline, assumption is that it won't add to extra queries because https://github.com/silverstripe/silverstripe-versioned/blob/2/src/Versioned.php#L1136 `$data = DB::prepared_query("SELECT * FROM \"{$table}\" WHERE \"ID\" = ?", [$recordID])->record();` is what creates the extra query, and so doing it in either the before or the after hook makes no difference.  Currently it does this in the `after` hook.

Ideally we'll do this query in the `before` hook, store it in memory in private property somewhere, and then create the _Versions record in the `after` hook, therefore removing the chance of creating a phantom _Versions record

emteknetnz (approved):
Look good to me

Issue #519, opened 2025-07-21
Set versioned mode in reset() for PHPUnit and Behat


emteknetnz:
PHPUnit and Behat both set the Versioned reading_mode to empty string in Versioned::reset() which is called on flush

This caused the need for a [workaround](https://github.com/silverstripe/silverstripe-elemental/pull/1364/files#diff-fcdd45c491f06a307292e900ba682fba608977e666dbca43bcc6376602829c39R362) which shouldn't need to be there

## Acceptance criteria
- [x] Versioned reading_mode should be set to `'Stage.' . Versioned::DRAFT` when PHPUnit and behat reset themselves between tests
- [x] This is tested against kitchen-sink CI
- [x] The workaround in elemental is removed

## PRs

[**Kitchen Sink CI**](https://github.com/creative-commoners/recipe-kitchen-sink/actions/runs/16456883610)

- https://github.com/silverstripe/silverstripe-versioned/pull/520
- https://github.com/silverstripe/silverstripe-behat-extension/pull/308
- https://github.com/silverstripe/silverstripe-versioned/pull/521

GuySartorelli:
Changing this for unit tests is causing too many tests to fail in sink. Most of the failures are due to `?stage=Stage` being appended to URLs. While I do think a stage should be set, changing that now will be too disruptive.

PR #521, opened 2025-07-22
FIX Make sure tests run in draft stage by default

GuySartorelli:
Looks like I put the set stage call in the wrong place in https://github.com/silverstripe/silverstripe-versioned/pull/520

## Issue

- https://github.com/silverstripe/silverstripe-versioned/issues/519


GuySartorelli:
This is causing too many tests to fail in sink. Most of the failures are due to `?stage=Stage` being appended to URLs. While I do think a stage should be set, changing that now will be too disruptive.

PR #522, opened 2025-07-28
ENH Combine SQL queries to improve performance.

GuySartorelli:
## Issue

- https://github.com/silverstripe/.github/issues/416

### DIFF
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -20,6 +20,7 @@ use SilverStripe\ORM\DataObject;
 use SilverStripe\ORM\DataQuery;
 use SilverStripe\ORM\DB;
 use SilverStripe\ORM\FieldType\DBDatetime;
+use SilverStripe\ORM\Queries\SQLDelete;
 use SilverStripe\ORM\Queries\SQLSelect;
 use SilverStripe\Security\Member;
 use SilverStripe\Security\Permission;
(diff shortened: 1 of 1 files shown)


src/Versioned.php
```diff
@@ -1131,10 +1132,9 @@ protected function cleanupVersionedOrphans($baseTable, $childTable)
         $count = $orphanedQuery->count();
         if ($count > 0) {
             DB::alteration_message("Removing {$count} orphaned versioned records", "deleted");
+            // We can't use $orphanedQuery as a subquery of the DELETE, since selecting from the table we're updating is forbidden in MySQL.
             $ids = $orphanedQuery->execute()->column();
-            foreach ($ids as $id) {
-                DB::prepared_query("DELETE FROM \"{$childTable}\" WHERE \"ID\" = ?", [$id]);
-            }
+            SQLDelete::create("\"{$childTable}\"", ['"ID" IN (' . DB::placeholders($ids) . ')' => $ids])->execute();
```
ChloeSartorelli, line 1137: It's not clear why these were being deleted one-by-one - it's part of the "Initial commit" with no comments. My guess is there's no specific reason for it.

Definitely a single query is more efficient than multiple queries in this case. This change reduces the queries from calling this method from n+2 to just 3. I would have reduced it to 2, but as per the comment in the code MySQL doesn't like that.

ChloeSartorelli: This functionality is covered by `VersionedTest::testDeletingOrphanedVersions()`

emteknetnz (changes requested):
Looks like you could get rid of the $count variable, move `$ids = $orphanedQuery->execute()->column();` up higher instead and just go `if (count($ids) > 0)`

That would also get rid of a mysql query?

GuySartorelli:
We still need the count variable (it's used in the alteration message) but you're right that just getting the columns and counting that array should be more efficient. Updated.

emteknetnz (approved):

PR #524, opened 2025-08-21
ENH Disable sorting on queries where sort order doesn't matter

GuySartorelli:
Sorting queries can be slow, especially on large datasets and especially if indexes aren't optimised. We should avoid sorting when the sort order doesn't matter.

## Issue

- https://github.com/silverstripe/silverstripe-framework/issues/11833

### DIFF
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -2499,7 +2499,7 @@ SQL
     protected function onPrepopulateTreeDataCache($recordList = null, array $options = [])
     {
         $idList = is_array($recordList) ? $recordList :
-            ($recordList instanceof DataList ? $recordList->column('ID') : null);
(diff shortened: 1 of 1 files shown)


src/Versioned.php
```diff
@@ -2499,7 +2499,7 @@ private static function determineVersionNumberByStage($class, $stage, $id, $cach
     protected function onPrepopulateTreeDataCache($recordList = null, array $options = [])
     {
         $idList = is_array($recordList) ? $recordList :
-            ($recordList instanceof DataList ? $recordList->column('ID') : null);
+            ($recordList instanceof DataList ? $recordList->getIDList() : null);
```
ChloeSartorelli: Doesn't matter what order the values are cached in.

emteknetnz (approved):

Issue #525, opened 2025-08-29
`Versioned::get_latest_version()` can return null - this needs to be accounted for


GuySartorelli:
`Versioned::get_latest_version()` relies on `DataList::first()` which will return `null` if there's no record to return.

This is rare, but is happening in https://github.com/tractorcow-farm/silverstripe-fluent/pull/1007 (see https://github.com/tractorcow-farm/silverstripe-fluent/actions/runs/17313034454/job/49150607636?pr=1007)

> ```text
> 1) TractorCow\Fluent\Tests\Extension\FluentVersionedExtensionTest::testCorrectVersionIdentified
> Attempt to read property "Version" on null
> 
> /home/runner/work/silverstripe-fluent/silverstripe-fluent/vendor/silverstripe/versioned/src/Versioned.php:2695
> /home/runner/work/silverstripe-fluent/silverstripe-fluent/vendor/silverstripe/framework/src/Core/Extensible.php:135
> /home/runner/work/silverstripe-fluent/silverstripe-fluent/vendor/silverstripe/framework/src/Core/CustomMethods.php:64
> /home/runner/work/silverstripe-fluent/silverstripe-fluent/tests/php/Extension/FluentVersionedExtensionTest.php:244
> /home/runner/work/silverstripe-fluent/silverstripe-fluent/src/State/FluentState.php:152
> /home/runner/work/silverstripe-fluent/silverstripe-fluent/tests/php/Extension/FluentVersionedExtensionTest.php:238
> /home/runner/work/silverstripe-fluent/silverstripe-fluent/vendor/bin/phpunit:122
> ```

## PRs

- https://github.com/silverstripe/silverstripe-versioned/pull/526

PR #526, opened 2025-08-29
FIX Allow for null values returned from get_latest_version

GuySartorelli:
While rare, it is possible for get_latest_version to return null, since it returns the value from `DataList::first()` which can return null if there's no record to return.

> [!IMPORTANT]
> This is required for https://github.com/tractorcow-farm/silverstripe-fluent/pull/1007

## Issue

- https://github.com/silverstripe/silverstripe-versioned/issues/525

### DIFF
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -2663,7 +2663,7 @@ SQL
      * @template T of DataObject
      * @param class-string<T> $class
      * @param int $id
-     * @return T&static
+     * @return null|T&static
      */
     public static function get_latest_version($class, $id)
     {
@@ -2692,7 +2692,7 @@ SQL
         }
 
         $version = static::get_latest_version($this->owner->baseClass(), $owner->ID);
-        return ($version->Version == $owner->Version);
+        return ($version?->Version == $owner->Version);
     }
 
     /**


src/Versioned.php
```diff
@@ -2692,7 +2692,7 @@ public function isLatestVersion()
         }
 
         $version = static::get_latest_version($this->owner->baseClass(), $owner->ID);
-        return ($version->Version == $owner->Version);
+        return ($version?->Version == $owner->Version);
```
emteknetnz, line 2695: Might be slightly better to have this above the return statement instead

```php
if ($version === null) {
    return false
}
````

Otherwise there might be some weird edge case where `$version` is null and `$owner->Version` is also null (or 0, though we're only using `==` so is functionally the same).  This might happen is say `isInDB()` was overrideen.

Seems like if there is no version, then this method should be returning `false`

emteknetnz (changes requested):

src/Versioned.php
ChloeSartorelli: If no version has been saved yet, presumably this should return `true`? i.e. `null == 0` would indicate we're on the latest version (in so far as there are no more recent versions)?
I don't want to change the functionality of this method (the value being `null` is no different than if it was `0` which is the default value).

This issue is for the narrow-scope problem where `$version` is null which causes this to throw an exception currently.

emteknetnz (approved):

PR #530, opened 2025-10-28
FIX Don't crash if there's no model class in gridfield

GuySartorelli:
Fixes https://github.com/silverstripe/silverstripe-reports/actions/runs/18870669305/job/53848059007 and https://github.com/silverstripe/silverstripe-framework/actions/runs/18835051455/job/53734250875

The error when going to the external link report locally is:

```text
[Emergency] Uncaught LogicException: GridField doesn't have a modelClassName, so it doesn't know the columns of this grid.
Line 241 in /var/www/html/vendor/silverstripe/framework/src/Forms/GridField/GridField.php
Trace
    SilverStripe\Forms\GridField\GridField->getModelClass()
    VersionedGridFieldRemoveComponentsExtension.php:19
```

## Issue

- https://github.com/silverstripe/.github/issues/460

### DIFF
--- a/src/VersionedGridFieldRemoveComponentsExtension.php
+++ b/src/VersionedGridFieldRemoveComponentsExtension.php
@@ -2,7 +2,7 @@
 
 namespace SilverStripe\Versioned;
 
-use SilverStripe\Core\ClassInfo;
+use LogicException;
 use SilverStripe\Core\Extension;
 use SilverStripe\Forms\GridField\GridField;
 
@@ -16,7 +16,15 @@ class VersionedGridFieldRemoveComponentsExtension extends Extension
     protected function onBeforeRenderHolder()
     {
         $owner = $this->getOwner();
-        $modelClass = $owner->getModelClass();
+        try {
+            $modelClass = $owner->getModelClass();
+        } catch (LogicException) {
+            // noop - it's possible to have a gridfield with custom components that don't rely on columns
+            // from the records in the list.
+            // Or more likely - an empty gridfield.
+            return;
+        }
+
         if (!method_exists($modelClass, 'has_extension') || !$modelClass::has_extension(Versioned::class)) {
             $owner->getConfig()->removeComponentsByType([GridFieldArchiveAction::class, GridFieldRestoreAction::class]);
         }


src/VersionedGridFieldRemoveComponentsExtension.php
```diff
@@ -16,9 +16,15 @@ class VersionedGridFieldRemoveComponentsExtension extends Extension
...
-        if (!method_exists($modelClass, 'has_extension') || !$modelClass::has_extension(Versioned::class)) {
-            $owner->getConfig()->removeComponentsByType([GridFieldArchiveAction::class, GridFieldRestoreAction::class]);
+        try {
+            $modelClass = $owner->getModelClass();
+            if (!method_exists($modelClass, 'has_extension') || !$modelClass::has_extension(Versioned::class)) {
+                $owner->getConfig()->removeComponentsByType([GridFieldArchiveAction::class, GridFieldRestoreAction::class]);
+            }
+        } catch (LogicException) {
+            // noop - it's possible to have a gridfield with custom components that don't rely on columns
+            // from the records in the list.
```
ChloeSartorelli, line 23: This part of the comment comes from GridField itself: https://github.com/silverstripe/silverstripe-framework/blob/e2df9243652ca9ca9f28c963fcb9ecd03635091d/src/Forms/GridField/GridField.php#L536-L541

src/VersionedGridFieldRemoveComponentsExtension.php
```diff
@@ -2,7 +2,7 @@
 
 namespace SilverStripe\Versioned;
 
-use SilverStripe\Core\ClassInfo;
```
ChloeSartorelli, line 5: Wasn't used

emteknetnz (changes requested):
Instead of the try/catch block, wouldn't it be better to just change the conditional statement to 

`if (!$modelClass || !method_exists($modelClass, 'has_extension') || !$modelClass::has_extension(Versioned::class)) {`

?

GuySartorelli:
How do you get the `$modelClass` variable in that scenario? Calling `$owner->getModelClass()` throws an exception for an empty list, which is what we're catching.

emteknetnz:
OK do this then

```php
$modelClass = '';
try {
    $modelClass = $owner->getModelClass();
} catch (LogicException $exception) {
    // noop ...
}
if (!$modelClass || ...)
```

That way it's obvious where the LogicException is coming from

emteknetnz (approved):

Issue #298, opened 2020-09-17
More options is showing even when there are no actions


lhalaa:
In a versioned DataObject, if we restrict users from unpublishing and archiving the object, the more options menu still shows.

<img width="445" alt="ActionMenus" src="https://user-images.githubusercontent.com/340514/93410968-3139d500-f8ee-11ea-891e-af07bc1ab8cc.png">

GuySartorelli:
PR merged. This will be automatically tagged by GitHub Actions

PR #531, opened 2025-11-03
FIX: Do not show More Options when no actions

hubertusanton:
## Description
In a versioned DataObject, if we restrict users from unpublishing and archiving the object, the more options menu still shows. Fixed with simply checking if optionsMenu contains actions.

## Manual testing steps
Create a Dataobject which has canPublish and canArchive set to false and optionsMenu will be gone.

## Issues
#298 

## Pull request checklist
- [x] The target branch is correct
    - See [picking the right version](https://docs.silverstripe.org/en/contributing/code/#picking-the-right-version)
- [x] All commits are relevant to the purpose of the PR (e.g. no debug statements, unrelated refactoring, or arbitrary linting)
    - Small amounts of additional linting are usually okay, but if it makes it hard to concentrate on the relevant changes, ask for the unrelated changes to be reverted, and submitted as a separate PR.
- [x] The commit messages follow our [commit message guidelines](https://docs.silverstripe.org/en/contributing/code/#commit-messages)
- [x] The PR follows our [contribution guidelines](https://docs.silverstripe.org/en/contributing/code/)
- [x] Code changes follow our [coding conventions](https://docs.silverstripe.org/en/contributing/coding_conventions/)
- [x] This change is covered with tests (or tests aren't necessary for this change)
- [x] Any relevant User Help/Developer documentation is updated; for impactful changes, information is added to the changelog for the intended release
- [x] CI is green

### DIFF
--- a/src/VersionedGridFieldItemRequest.php
+++ b/src/VersionedGridFieldItemRequest.php
@@ -403,9 +403,16 @@ class VersionedGridFieldItemRequest extends GridFieldDetailForm_ItemRequest
 
             if ($actionMenusMoreOptions) {
                 // Check if there are any FormAction fields (recursively through nested tabs)
-                if (!$actionMenusMoreOptions->Fields()->flattenFields()->filterByCallback(function ($field) {
-                    return $field instanceof FormAction;
-                })->first()) {
+                $hasFormAction = false;
+                $actionMenusMoreOptions->Fields()->recursiveWalk(function ($field) use (&$hasFormAction) {
+                    if ($field instanceof FormAction) {
+                        $hasFormAction = true;
+                        return false; // Stop walking
+                    }
+                    return true; // Continue walking
+                });
+
+                if (!$hasFormAction) {
                     $form->Actions()->removeByName('ActionMenus');
                 }
             }
--- a/src/VersionedGridFieldItemRequest.php
+++ b/src/VersionedGridFieldItemRequest.php
@@ -395,7 +395,10 @@ class VersionedGridFieldItemRequest extends GridFieldDetailForm_ItemRequest
             $moreOptions->push($actionArchive);
         }
 
-        $actions->insertAfter('MajorActions', $rootTabSet);
+        // Only add $moreOptions menu if it contains action buttons
+        if ($moreOptions->Fields()->count() > 1) {
+            $actions->insertAfter('MajorActions', $rootTabSet);
+        }
     }
 
     /**
--- a/src/VersionedGridFieldItemRequest.php
+++ b/src/VersionedGridFieldItemRequest.php
@@ -395,10 +395,21 @@ class VersionedGridFieldItemRequest extends GridFieldDetailForm_ItemRequest
             $moreOptions->push($actionArchive);
         }
 
-        // Only add $moreOptions menu if it contains action buttons
-        if ($moreOptions->Fields()->count() > 1) {
-            $actions->insertAfter('MajorActions', $rootTabSet);
-        }
+        $actions->insertAfter('MajorActions', $rootTabSet);
+
+        // Remove ActionMenus when ActionMenus.MoreOptions has no actions
+        $this->afterExtending('updateItemEditForm', function (Form $form) {
+            $actionMenusMoreOptions = $form->Actions()->findTab('ActionMenus.MoreOptions');
+
+            if ($actionMenusMoreOptions) {
+                // Check if there are any FormAction fields (recursively through nested tabs)
+                if (!$actionMenusMoreOptions->Fields()->flattenFields()->filterByCallback(function ($field) {
+                    return $field instanceof FormAction;
+                })->first()) {
+                    $form->Actions()->removeByName('ActionMenus');
+                }
+            }
+        });
     }
 
     /**
--- a/src/VersionedGridFieldItemRequest.php
+++ b/src/VersionedGridFieldItemRequest.php
@@ -397,22 +397,20 @@ class VersionedGridFieldItemRequest extends GridFieldDetailForm_ItemRequest
 
         $actions->insertAfter('MajorActions', $rootTabSet);
 
-        // Remove ActionMenus when ActionMenus.MoreOptions has no actions
+        // Remove ActionMenus when ActionMenus.MoreOptions has no content
         $this->afterExtending('updateItemEditForm', function (Form $form) {
             $actionMenusMoreOptions = $form->Actions()->findTab('ActionMenus.MoreOptions');
 
             if ($actionMenusMoreOptions) {
-                // Check if there are any FormAction fields (recursively through nested tabs)
-                $hasFormAction = false;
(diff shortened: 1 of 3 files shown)


GuySartorelli:
If you have an extension which implements `updateFormActions()` and adds more action buttons in the "more actions" tab, this PR might introduce a new bug where the `ActionMenus` (which you want to add your custom actions to) is missing.

I think to resolve this bug we will need to remove the field _after_ all relevant extensions have finished, e.g:

```php
$this->afterExtending('updateItemEditForm', function (Form $form) {
    if (/* check if the tabset is effectively empty here */) {
        $form->Actions()->removeByName('ActionMenus');
    }
});
```

Note that in those extensions someone might add additional tabs to the `ActionMenus` tabset, so just checking the fields inside the `MoreActions` tab will be insufficient.

hubertusanton:
I need to implement a recursively check if there are any FormActions in the fields, what is the best place to implement this? I guess not in VersionedGridFieldItemRequest itself?

lozcalver:
@hubertusanton `Form::defaultAction()` does something similar:

https://github.com/silverstripe/silverstripe-framework/blob/c61c5f16e06e3c00ffd466e4b0844cf36b13d2c9/src/Forms/Form.php#L1682-L1684

GuySartorelli:
`flattenFields()` has [a bug](https://github.com/silverstripe/silverstripe-framework/issues/10054) that makes it less desirable to use in such a central place. `recursiveWalk()` is probably the better option.

src/VersionedGridFieldItemRequest.php
```diff
@@ -396,6 +396,20 @@ protected function addVersionedButtons(DataObject $record, FieldList $actions)
...
 
         $actions->insertAfter('MajorActions', $rootTabSet);
+
+        // Remove ActionMenus when ActionMenus.MoreOptions has no actions
+        $this->afterExtending('updateItemEditForm', function (Form $form) {
+            $actionMenusMoreOptions = $form->Actions()->findTab('ActionMenus.MoreOptions');
+
+            if ($actionMenusMoreOptions) {
+                // Check if there are any FormAction fields (recursively through nested tabs)
+                if (!$actionMenusMoreOptions->Fields()->flattenFields()->filterByCallback(function ($field) {
```
ChloeSartorelli: Please use `recursiveWalk` instead as per https://github.com/silverstripe/silverstripe-versioned/pull/531#issuecomment-3523694166

GuySartorelli (changes requested):
Looking much better. One small change. Can you please also have a go at writing a quick unit test for this?

tests/php/VersionedGridFieldItemRequestTest.php
```diff
@@ -151,6 +152,65 @@ public function testUnversionedObjectsPublishChildren()
...
+            $this->assertInstanceOf(
+                FormAction::class,
+                $testAction,
+                'Test action added by extension should be present'
+            );
+        } finally {
+            // Clean up
+            TestExtension::$add_test_action = false;
+            VersionedGridFieldItemRequest::remove_extension(TestExtension::class);
+        }
```
ChloeSartorelli, line 189: For future reference you can implement the `Resettable` interface in the test fixture to reset this, and then we don't need a `try/finally`.
Since this is only done in one place it's fine as is - just a note for the future.

src/VersionedGridFieldItemRequest.php
```diff
@@ -396,6 +396,27 @@ protected function addVersionedButtons(DataObject $record, FieldList $actions)
...
+
+            if ($actionMenusMoreOptions) {
+                // Check if there are any FormAction fields (recursively through nested tabs)
+                $hasFormAction = false;
+                $actionMenusMoreOptions->Fields()->recursiveWalk(function ($field) use (&$hasFormAction) {
+                    if ($field instanceof FormAction) {
+                        $hasFormAction = true;
+                        return false; // Stop walking
+                    }
+                    return true; // Continue walking
```
ChloeSartorelli: These return statements don't seem to be used by anything - did I miss something? Can you please show me how these return statements stop or continue the walk?

tests/php/VersionedGridFieldItemRequestTest/TestExtension.php
```diff
@@ -0,0 +1,27 @@
...
+
+/**
+ * Test extension that adds actions via updateFormActions
+ */
+class TestExtension extends Extension implements TestOnly
+{
+    /**
+     * @var bool Flag to control whether to add actions
+     */
+    public static $add_test_action = false;
```
ChloeSartorelli:
```suggestion
     * Flag to control whether to add actions
     */
    public static bool $add_test_action = false;
```
We prefer strict types rather than PHPDoc type hints for new API. Doesn't really matter as much for test fixtures but since there's likely other changes to make this may as well get changed at the same time.

tests/php/VersionedGridFieldItemRequestTest/TestExtension.php
```diff
@@ -0,0 +1,27 @@
...
+
+    public function updateFormActions(FieldList $actions)
+    {
+        if (self::$add_test_action) {
```
ChloeSartorelli:
```suggestion
        if (static::$add_test_action) {
```
We don't use `self` unless we absolutely have to - and in this case `static` is more appropriate since this isn't referencing `private` API or a `final` class.

I'm surprised the linter didn't pick up on this.

tests/php/VersionedGridFieldItemRequestTest/TestExtension.php
```diff
@@ -0,0 +1,27 @@
...
+            $moreOptions = $actions->findOrMakeTab('ActionMenus.MoreOptions');
+            $moreOptions->push(FormAction::create('doTestAction', 'Test Action'));
+        }
+    }
+}
```
ChloeSartorelli: Please add a new line after this (as per linting failure).

The `.editorconfig` rules for this repo can tell your IDE to do things like this automatically. For example with vscode the [EditorConfig for VS Code](https://marketplace.visualstudio.com/items?itemName=EditorConfig.EditorConfig) plugin can deal with that for you.

GuySartorelli (changes requested):
Without this PR, the `...` menu has some information about my versioned record:
image

With this PR that information isn't surfaced anywhere.

In the screenshot that information isn't present, and in that scenario it makes sense to not show the `...` button at all, but if the information is there we should probably keep the button.

Looks like this information is added through `VersionedGridFieldItemRequest::addVersionedButtons()` in the `$infoTemplate` variable which is added into the `MoreOptions` tab as a literalfield named `Information`.

Probably the problem is that you're explicitly looking for `FormAction` - probably if there's _anything other than a `TabSet` or `Tab`_ we should keep the button.

src/VersionedGridFieldItemRequest.php
hubertusanton: They dont :)

GuySartorelli (approved):
LGTM, works as expected.

PR #533, opened 2025-11-06
ENH Add searchable fields for ChangeSet.

GuySartorelli:
Campaign admin filters use `SearchContext` to avoid re-inventing the wheel. That requires a valid searchable_fields config for the campaign model which happens to be this weird thing.

It would be good to decouple campaigns and changeset's publishrecursive logic at some stage, but this is not that stage.

## Issue

- https://github.com/silverstripe/silverstripe-campaign-admin/issues/365

### DIFF
--- a/src/ChangeSet.php
+++ b/src/ChangeSet.php
@@ -17,6 +17,8 @@ use SilverStripe\ORM\FieldType\DBDatetime;
 use SilverStripe\ORM\HasManyList;
 use SilverStripe\ORM\UnexpectedDataException;
 use SilverStripe\Core\Validation\ValidationException;
+use SilverStripe\Forms\DropdownField;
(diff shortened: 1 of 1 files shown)


src/ChangeSet.php
```diff
@@ -96,6 +98,21 @@ class ChangeSet extends DataObject
         'PublishedLabel' => 'Published',
     ];
 
+    private static $searchable_fields = [
```
emteknetnz:
```suggestion
    private static array $searchable_fields = [
```

emteknetnz (changes requested):

src/ChangeSet.php
ChloeSartorelli: Done

emteknetnz (approved):

PR #536, opened 2025-12-01
FIX Only query live stage if object has stages

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-versioned-admin/issues/437

### DIFF
--- a/src/Versioned.php
+++ b/src/Versioned.php
@@ -2709,8 +2709,8 @@ SQL
         if (!$id || !$this->isPublished()) {
             return false;
         }
-
-        $liveVersionNumber = static::get_versionnumber_by_stage($this->owner, Versioned::LIVE, $id);
+        $stage = $this->hasStages() ? Versioned::LIVE : Versioned::DRAFT;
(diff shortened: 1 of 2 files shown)


GuySartorelli (commented):
Works as expected locally, but is there a reason this isn't targetting CMS 5?

The specific use case the issue highlighted (`HistoryViewerField`) is harder to set up in CMS 5, but:
1. it's not impossible, just requires a bunch of react and graphql faffing about iirc
2. `isLiveVersion()` isn't just for `HistoryViewerField`

emteknetnz:
> it's not impossible, just requires a bunch of react and graphql faffing about iirc

That was the primary reason. I though it was actually impossible from memory, at a practical level I wouldn't expect anyone to actually do this.

I'll make a CMS 5.4 version with and run it through kitchen sink CI to see if it works.  If there are no issues that I don't think it's worth spending any time fixing for 5.4 though since it's not a reported issue

emteknetnz:
Closed in favor of CMS 5.4 version https://github.com/silverstripe/silverstripe-versioned/pull/537

PR #537, opened 2025-12-02
FIX Only query live stage if object has stages

emteknetnz:
Issue https://github.com/silverstripe/silverstripe-versioned-admin/issues/437


GuySartorelli (approved):
LGTM. Works as expected locally

PR #538, opened 2026-02-24
FIX Allow a null object when getting changeset edit link

GuySartorelli:
`ChangeSetItem::getObjectInStage()` returns the value of `DataList::byID()` which is explicitly nullable. It needs to handle null values.

Otherwise we get this error after https://github.com/silverstripe/silverstripe-assets/pull/707 gets merged:

> ERROR [Emergency]: Uncaught Error: Call to a member function CMSEditLink() on null
> IN GET /admin/campaigns/sets
> Line 537 in /var/www/html/vendor/silverstripe/versioned/src/ChangeSetItem.php

## Issue

- https://github.com/silverstripe/silverstripe-campaign-admin/issues/393

### DIFF
--- a/src/ChangeSetItem.php
+++ b/src/ChangeSetItem.php
@@ -170,7 +170,7 @@ class ChangeSetItem extends DataObject implements Thumbnail
      * If the object isn't versioned it will return the normal record.
      *
      * @param string $stage
-     * @return DataObject|Versioned|RecursivePublishable Object in this stage (may not be Versioned)
+     * @return DataObject|Versioned|RecursivePublishable|null Object in this stage (may not be Versioned)
      * @throws UnexpectedDataException
      */
     protected function getObjectInStage($stage)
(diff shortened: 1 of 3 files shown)


src/ChangeSetItem.php
```diff
@@ -170,7 +170,7 @@ public function getChangeType()
      * If the object isn't versioned it will return the normal record.
      *
      * @param string $stage
-     * @return DataObject|Versioned|RecursivePublishable Object in this stage (may not be Versioned)
+     * @return DataObject|Versioned|RecursivePublishable|null Object in this stage (may not be Versioned)
```
ChloeSartorelli, line 173: This change just reflects reality - this method returns the value from `DataList::byID()` which is explicitly nullable.

emteknetnz (changes requested):
Add a test that `CMSEditLink()` returns `null` when the underlying object can’t be loaded

GuySartorelli:
Test added

emteknetnz (approved):

PR #541, opened 2026-06-24
[CVE-2026-55779] XSS in RestoreAction message

emteknetnz:


blueo (approved):

xini:
@emteknetnz @blueo  I'm using https://github.com/Roave/SecurityAdvisories to check for known vulnerabilities. Because this fix has not been ported to Silverstripe 5, I can't run an update on Silverstripe 5 sites because silverstripe-versioned v2 gets blocked. 
Any idea how I could handle this?

blueo:
> @emteknetnz @blueo I'm using https://github.com/Roave/SecurityAdvisories to check for known vulnerabilities. Because this fix has not been ported to Silverstripe 5, I can't run an update on Silverstripe 5 sites because silverstripe-versioned v2 gets blocked. Any idea how I could handle this?

Hi @xini, you're not the first person to bring this up so we've added a section to the security release blog around this: https://www.silverstripe.org/blog/silverstripe-cms-security-patches-june-2026. I hope that answers your question

xini:
@blueo I’m sorry, but I really think you’re taking the wrong approach here. The current policy effectively means that Silverstripe 5.4, despite not being end‑of‑life, is knowingly left with unresolved security vulnerabilities. That’s not acceptable for a maintained release, and it puts users in a position where they cannot rely on the software being safe.

I strongly urge you to reconsider the policy so that all CVEs are backported to every supported, non‑EOL version.

The argument that “we’re only a small team” doesn’t really hold up when community contributions are systematically slowed down. When even core committers hesitate to merge improvements without approval from a single bottleneck, the project becomes less productive and less resilient. Why do you think that contributions have declined over the past few years? Our contributions are being blocked or ignored. That has the consequence that our patience and interest fade. If you want more hands on deck, the contribution process needs to be opened up rather than tightened down.

